PR #51, opened 2023-12-15
fix: Check CompletionItem.getKind() for null before using it.

### LINKED ISSUE #50: Completion items do not appear if the server response is not recognised
An exception can occur which causes the completion items not to appear in the IDE. 

If the `textDocument/completion` message contains items which do not have the `completionItemKind` then there is an exception caused by a null pointer de-reference. This field is optional in the specification so we should check for null.

turkeylurkey:
Fixes #50

### DIFF
--- a/src/main/java/com/redhat/devtools/lsp4ij/operations/completion/LSPCompletionContributor.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/operations/completion/LSPCompletionContributor.java
@@ -132,7 +132,8 @@ public class LSPCompletionContributor extends CompletionContributor {
             // Create lookup item
             var lookupItem = createLookupItem(file, editor, completionPrefix.getCompletionOffset(), item, itemDefaults, languageServer);
             // Group it by using completion item kind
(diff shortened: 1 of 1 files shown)


angelozerr:
@turkeylurkey thanks so much for your contribution.

We follow some convention for the commit. 

Available types:
 - feat: A new feature
 - fix: A bug fix
 - docs: Documentation only changes
 - style: Changes that do not affect the meaning of the code (white-space, formatting, missing semi-colons, etc)
 - refactor: A code change that neither fixes a bug nor adds a feature
 - perf: A code change that improves performance
 - test: Adding missing tests or correcting existing tests
 - build: Changes that affect the build system or external dependencies (example scopes: gulp, broccoli, npm)
 - ci: Changes to our CI configuration files and scripts (example scopes: Travis, Circle, BrowserStack, SauceLabs)
 - chore: Other changes that don't modify src or test files
 - revert: Reverts a previous commit

It means that you need to rename your commit like this:

```
fix: Check CompletionItem.getKind() for null before using it. 

Fixes #51

```

turkeylurkey:
Thanks, I hadn't looked at "Checks" until now.

angelozerr:
Thanks @turkeylurkey !

PR #53, opened 2023-12-18
fix: prevent from NPE when CompletionItem.textEdit field is not specified

### LINKED ISSUE #52: Completion items do not appear if the textEdit field is missing
In a completion item the label property can be the text that is inserted when the item is selected. In this case the completion item may not have the optional textEdit property. We must check if this property is supplied before using it.

turkeylurkey:
Fixes #52

### DIFF
--- a/src/main/java/com/redhat/devtools/lsp4ij/operations/completion/LSPCompletionContributor.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/operations/completion/LSPCompletionContributor.java
@@ -132,7 +132,8 @@ public class LSPCompletionContributor extends CompletionContributor {
             // Create lookup item
             var lookupItem = createLookupItem(file, editor, completionPrefix.getCompletionOffset(), item, itemDefaults, languageServer);
             // Group it by using completion item kind
-            var groupedLookupItem = PrioritizedLookupElement.withGrouping(lookupItem, item.getKind().getValue());
+            int group = item.getKind() == null ? 0 : item.getKind().getValue();
+            var groupedLookupItem = PrioritizedLookupElement.withGrouping(lookupItem, group);
             // Compute the prefix
             String prefix = completionPrefix.getPrefixFor(lookupItem.getTextEditRange(), item);
             if (prefix != null) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/redhat/devtools/lsp4ij/operations/completion/LSPCompletionContributor.java
```diff
@@ -132,9 +132,13 @@ private void addCompletionItems(PsiFile file, Editor editor, CompletionPrefix co
             // Create lookup item
             var lookupItem = createLookupItem(file, editor, completionPrefix.getCompletionOffset(), item, itemDefaults, languageServer);
             // Group it by using completion item kind
-            var groupedLookupItem = PrioritizedLookupElement.withGrouping(lookupItem, item.getKind().getValue());
+            int group = item.getKind() == null ? 0 : item.getKind().getValue();
+            var groupedLookupItem = PrioritizedLookupElement.withGrouping(lookupItem, group);
             // Compute the prefix
-            String prefix = completionPrefix.getPrefixFor(lookupItem.getTextEditRange(), item);
+            String prefix = null;
+            if (lookupItem.getTextEdit() != null) {
```
angelozerr, line 139: Exposing a new method like getTextEdit just tocheck the null value is shame. IMHO we should continue to use getTextEditRange:

```java
var textEditRange = lookupItem.getTextEditRange();
String prefix = textEditRange != null ? completionPrefix.getPrefixFor(textEditRange, item) : null;
```

getTextEditRange must be updated to return null if item.getTextEdit() is null. If you do that please add Javadoc to this getTextEditRange and anotate `with @Nullable`

```java
public @Nullable Range getTextEditRange() {
    Either<TextEdit, InsertReplaceEdit> textEdit = item.getTextEdit();
    if (textEdit == null) {
        return null;
    }
    if (textEdit.isLeft()) {
        return textEdit.getLeft().getRange();
    } else {
        // here providing insert range, currently do not know if insert or replace is requested
        return textEdit.getRight().getInsert();
    }
}
```

The main idea when we do some PR is to clean the code from LSP4IJ step by step by adding `@NotNull` and `@Nullable` annotations and add Javadoc.

turkeylurkey: Sounds good to me, I was also thinking of something like this.

angelozerr:
Thanks @turkeylurkey !

PR #55, opened 2023-12-18
fix: Check if CompletionOptions.getResolveProvider() is null before using it.

### LINKED ISSUE #54: Exception during completion if CompletionOptions.getResolveProvider returns null
In CompletionOptions the resolveProvider property is optional so in lsp4j the field is a Boolean. In the case the language server does not supply this property the field is null. We must check for null before using it.

turkeylurkey:
Let me know if you think one very long line is better. 

Fixes #54


angelozerr:
Thanks @turkeylurkey !

PR #66, opened 2024-01-04
fix: Change default highlight severity from Info to Error.

turkeylurkey:
When the language server does not set the severity for a diagnostic we choose a default of Information. On IntelliJ this means no on-screen highlighting. It would be better to be like VS Code and use a red underline (Error) by default.


angelozerr:
Thanks @turkeylurkey !

PR #79, opened 2024-01-10
feat: emulate vscode's "editor.action.triggerSuggest" command

fbricon:
Allows reusing VS Code-provided LS using the "editor.action.triggerSuggest" command, like the CSS LS:

Jan-10-2024 18-08-30


src/main/resources/META-INF/plugin.xml
```diff
@@ -173,6 +173,9 @@
                 id="lsp.console.explorer.copy.command"
                 class="com.redhat.devtools.lsp4ij.console.explorer.actions.CopyStartServerCommandAction"
                 icon="AllIcons.Actions.Copy"/>
+        <action id="editor.action.triggerSuggest"
```
angelozerr, line 176: Could you please add this action in a new block actions and add a comment like waiting for standardization of LSP commands and set a link to https://github.com/microsoft/language-server-protocol/issues/788

angelozerr:
Wow, it looks amazing @fbricon !

src/main/resources/META-INF/plugin.xml
fbricon: Done

PR #83, opened 2024-01-12
fix: make loadTemplateStream return a BufferedInputStream

fbricon:


angelozerr:
I'm just curious, what is the beneit to use BufferedInputStream?

angelozerr:
Thanks @fbricon !

PR #146, opened 2024-02-20
fix: prevent NPE in LSPUsageSearcher.processElementUsages

### LINKED ISSUE #145: NPE in LSPUsageSearcher.processElementUsages
Not sure how I got this error (can't reproduce it):

    java.lang.NullPointerException: Cannot invoke “java.util.List.iterator()” because “usages” is null
	at com.redhat.devtools.lsp4ij.usages.LSPUsageSearcher.processElementUsages(LSPUsageSearcher.java:77)
	at com.intellij.find.findUsages.FindUsagesManager.lambda$createUsageSearcher$10(FindUsagesManager.java:401)

Looks like the future completed and returned `null`, not the default value from `getNow`
https://github.com/redhat-developer/lsp4ij/blob/e0c82b83697b29afaf8dd7c771fefaf288263e28/src/main/java/com/redhat/devtools/lsp4ij/usages/LSPUsageSearcher.java#L76

fbricon:
Fixes #145


angelozerr:
Thanks @fbricon !

PR #148, opened 2024-02-20
fix: compute range from 1st non-whitespace word, if start==end

### LINKED ISSUE #125: Improve diagnostic range conversion when range.start=range.end
Given a go.mod file with an invalid statement, using gopls, error rendering looks bad:

Screenshot 2024-01-30 at 09 30 03

compared to vscode, where the entire words are underlined:
Screenshot 2024-01-30 at 09 31 09

fbricon:
Something fishy with intermediate diagnostics being computed on stale lines
Feb-20-2024 12-16-06

Fixes #125

### DIFF
--- a/src/main/java/com/redhat/devtools/lsp4ij/LSPIJUtils.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/LSPIJUtils.java
@@ -325,17 +325,21 @@ public class LSPIJUtils {
                 // Language server reports invalid diagnostic, ignore it.
                 return null;
             }
-            // Adjust start / end offset if needed
-            if (start == end) {
-                if (!adjust) {
-                    // No adjustment, the TextRange with start/end offset is invalid
-                    return null;
-                }
-                // Adjust the end offset if the offset is not at the end of the line.
-                int offset = start;
-                if (!isEndOfLine(offset, range.getEnd().getLine(), document)) {
-                    end++;
-                }
+            if (start != end) {
+                return new TextRange(start, end);
+            }
+            if (!adjust) {
+                // No adjustment, the TextRange with start/end offset is invalid
+                return null;
+            }
+            // Select token at current offset, if possible
+            TextRange tokenRange = getTokenRange(document, start);
+            if (tokenRange != null) {
+                return tokenRange;
+            }
+            // Adjust the end offset if the offset is not at the end of the line.
+            if (!isEndOfLine(document, range.getEnd().getLine(), start)) {
+                end++;
             }
             return new TextRange(start, end);
         } catch (IndexOutOfBoundsException e) {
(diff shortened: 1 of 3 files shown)


src/main/java/com/redhat/devtools/lsp4ij/LSPIJUtils.java
```diff
@@ -372,9 +373,31 @@ public static Range toRange(TextRange range, Document document) {
...
+    }
+
+    private static int getNextNonWhitespaceOffset(@NotNull Document document, int line, int offset) {
+        // since this method is private, at this point we already guarantee that:
+        // - line is valid
+        // - offset < document.getTextLength()
+        int textLength = document.getTextLength();
+
+        for (; offset <= document.getLineEndOffset(line) && offset < textLength; offset++) {
+            if (Character.isWhitespace(document.getText().charAt(offset))) {
```
angelozerr, line 392: Don't use `document.getText()` which create a new string but use **document.getCharsSequence().charAt(i);**

fbricon: nice catch!

angelozerr:
I have written a similar code at https://github.com/redhat-developer/lsp4ij/blob/265a2d2b438642a1de2911e52c8e68073e81de6c/src/main/java/com/redhat/devtools/lsp4ij/usages/LSPUsageTargetProvider.java#L49 

So I wonder if we could have a new static method utils which retrieve a TextRange of a token just by giving an offset.

For instance `LSP4IJUtils#tokenRange(int offset, Document document)`

angelozerr:
> Something fishy with intermediate diagnostics being computed on stale lines

This problem comes from with Go language server which is very slow to publish diagnostic. 

I don't know how to fix this problem in IJ.

fbricon:
> So I wonder if we could have a new static method utils which retrieve a TextRange of a token just by giving an offset.
> 
> For instance `LSP4IJUtils#tokenRange(int offset, Document document)`

Done

Hopefully I didn't break anything

Screenshot 2024-03-14 at 17 21 34

src/main/java/com/redhat/devtools/lsp4ij/LSPIJUtils.java
```diff
@@ -345,9 +349,43 @@ public static Range toRange(TextRange range, Document document) {
...
 
-    private static boolean isEndOfLine(int offset, int line, @NotNull Document document) {
-        int lineEndOffset = document.getLineEndOffset(line);
-        return offset == lineEndOffset;
+    private static boolean isEndOfLine(@NotNull Document document, int line, int offset) {
+        return offset == document.getLineEndOffset(line);
+    }
+    
+    @Nullable
+    public static TextRange getTokenRange(Document document , int offset) {
```
angelozerr, line 357: Add javadoc

angelozerr:
Do you think you could add some test?

angelozerr:
It works like a charm,thanks @fbricon !

Issue #210, opened 2024-04-14
Provide validation in the preferences of LSP console


angelozerr:
The preferences in LSP console doesn't provide no validation:

![image](https://github.com/redhat-developer/lsp4ij/assets/1932211/fcc8b9c4-8f05-43a8-a206-6ef57935b103)

although the dialog provide it:

![image](https://github.com/redhat-developer/lsp4ij/assets/1932211/640814da-2a7b-40f0-a1b0-f36fb4c71218)

Please note that those 2 preferences panel use the same class.

It would be nice to support validation in LSP console preferences.

The validation should be too improved with:

 * validate that there is on mapping which is defined.
 * display a border red for JTextArea (ex: command,configuration), I don't know why it is working for JTextField but not with JTextArea. There is the same problem when you click on JTextArea, there is no blue border like other widgets (JComboBox, JTextField).

@MituuZ are you interested to work on this issue?

MituuZ:
Yeah, I can take a look at this too

angelozerr:
Thanks!

MituuZ:
I looked around a bit and am thinking about creating the validation function (and also possibly the error stylin) to the widgets. Just have to see how the DialogWrapper works with custom validation functionality.

Another option could be to extend the DialogWrapper in the tool window too, but I'm not sure how clean that would be.

angelozerr:
To be honnest with you I have never done that, so we need some experimentation.

MituuZ:
Alright, I'll poke around a bit and see how the custom validation turns out. The basic implementation is quite simple, so I need to figure out how  to enable/disable the creation/saving if the widget validation fails.

Issue #215, opened 2024-04-14
Adding a new LS should automatically make it start if matching documents are already opened


fbricon:
- open a typescript file
- add the typescript LS, set debug mode to verbose
- the typescript LS doesn't automatically start, unless you put the focus back to the .ts file

There should be a mechanism to loop through all opened editors et check if the newly added server matches their document

fbricon:
See https://github.com/redhat-developer/lsp4ij/blob/0247aac899c85cd69e773b8439b0d743f16aaf96/src/main/java/com/redhat/devtools/lsp4ij/server/definition/LanguageServerDefinitionListener.java#L25

angelozerr:
@MituuZ are you interested to work on this issue?

If yes don't hesitate to ping me if you need more informations. As @fbricon said to implement this issue, we need to use the existing listener  LanguageServerDefinitionListener loop for each opened editors and starts the language server if one editor matches the language server (it is just my first feeling perhaps it s more complicated)

MituuZ:
Getting the editors shouldn't be hard, so it might be relatively simple. You can assign this to me, I'll get it to it later

angelozerr:
Thanks @MituuZ !

angelozerr:
Fixed with https://github.com/redhat-developer/lsp4ij/pull/262

PR #222, opened 2024-04-17
feat: Validate preferences in lsp console

MituuZ:
Set up common validation style for LSP console and new LS dialog for text fields and text areas.

Use similar validation for LSP console and new LS dialog. Prevent saving/creating if validations fail.


MituuZ:
Fixes #210

Basic working implementation, but still needs some cleanup and docs

src/main/java/com/redhat/devtools/lsp4ij/console/actions/ApplyLanguageServerSettingsAction.java
```diff
@@ -80,7 +80,7 @@ public void update(@NotNull AnActionEvent e) {
             }
         }
 
-        e.getPresentation().setEnabled(modified);
+        e.getPresentation().setEnabled(modified && languageServerView.isValid());
```
angelozerr, line 83: Perhaps user want to save an invalid language server defintion?

src/main/java/com/redhat/devtools/lsp4ij/settings/ui/ServerNameWidget.java
```diff
@@ -0,0 +1,36 @@
+package com.redhat.devtools.lsp4ij.settings.ui;
```
angelozerr, line 1: Add header by using EPL license and adding your name / email

src/main/java/com/redhat/devtools/lsp4ij/settings/ui/ServerNameWidget.java
```diff
@@ -0,0 +1,36 @@
...
+
+import com.intellij.openapi.ui.ValidationInfo;
+import com.intellij.ui.components.JBTextField;
+import com.redhat.devtools.lsp4ij.LanguageServerBundle;
+
+import javax.swing.border.Border;
+
+public class ServerNameWidget extends JBTextField implements ValidatableConsoleWidget {
```
angelozerr, line 9: Add javadoc

src/main/java/com/redhat/devtools/lsp4ij/settings/ui/ValidatableConsoleWidget.java
```diff
@@ -0,0 +1,62 @@
+package com.redhat.devtools.lsp4ij.settings.ui;
```
angelozerr, line 1: Add header by using EPL license and adding your name / email

src/main/java/com/redhat/devtools/lsp4ij/console/actions/ApplyLanguageServerSettingsAction.java
MituuZ: I went with the same validation that creating a new LS has, should this be removed and allow saving an invalid configuration?

MituuZ: Alright, that does make sense too. I'll change the validation on that

angelozerr: Thanks!

angelozerr:
@MituuZ it should be nice to see all errors when you open dialog or change a field value.

As New dialog, preferences and LSP console detail uses the same componet LanguageServerPanel, I think you should delegate validation in languageServerPanel.doValidateAll()

I have tried to start quickly the idea https://github.com/redhat-developer/lsp4ij/compare/main...angelozerr:lsp4ij:poc_error?expand=1 so I think we should work just with ValidationInfo and not have an isValid method.

MituuZ:
I'll take a look and re-implement the validation according to that

angelozerr:
I did that quickly so dont hesitate to my code.

It should be very nice to see all errors or when you fix an error (ex server name) it should be nice to see the error in command.

As we use tabs, I wonder if we could se et the tab in red which contains error.

src/main/java/com/redhat/devtools/lsp4ij/settings/ui/ValidatableConsoleWidget.java
```diff
@@ -0,0 +1,86 @@
...
+     * Overridable method that implements the field validation
+     * Handles updating the style of the component for error/valid
+     */
+    void validateInput();
+
+    /**
+     * Overridable method that returns a ValidationInfo, is the component is not valid
+     * @return a ValidationInfo element if component validation fails, null if component is valid
+     */
+    ValidationInfo getValidationInfo();
```
angelozerr, line 80: I wonder if we could simplify more the API and just have one method void `validate(List<ValidationInfo> validations)`.

It will give the capability to add several validation for the same widget (not used today yet)

This validate method could manage border too.

For the listener, we should not need since addValidator in https://github.com/redhat-developer/lsp4ij/pull/222/files#diff-c904db49f6e85960bbd7640a0e932dc0505da3b799010fad27ef9e4af6753ef0R268 add a listner and refresh validation for all fields.

MituuZ: Yeah, that sounds good

MituuZ: Still needs a cleanup and javadocs around, but I'm thinking that now it's starting to look nice.

I'll continue tomorrow on those

src/main/java/com/redhat/devtools/lsp4ij/settings/ui/CommandLineWidget.java
```diff
@@ -10,23 +10,45 @@
...
+
+    @Override
+    public void validate(List<ValidationInfo> validations) {
+        boolean valid = true;
+        if (getDocument() != null && getText().isBlank()) {
+            setErrorBorder(this);
+            valid = false;
+            validations.add(new ValidationInfo(errorMessage, this));
+        }
+        if (valid) {
```
MituuZ, line 50: I'll move the error border setting to here, so we have a single place that handles it. Same with the server name widget

src/main/java/com/redhat/devtools/lsp4ij/settings/ui/LanguageServerPanel.java
```diff
@@ -232,4 +243,26 @@ public ComboBox<ServerTrace> getServerTraceComboBox() {
     public ComboBox<ErrorReportingKind> getErrorReportingKindCombo() {
         return errorReportingKindCombo;
     }
+
+    public @NotNull List<ValidationInfo> doValidateAll() {
+        List<ValidationInfo> validations = new ArrayList<>();
+        var serverNameWidget = getServerName();
```
MituuZ, line 249: @angelozerr what do you think about adding `@NotNull` to the getServerName and getCommandLine?

Would be cleaner to just call them like this:
```java
getServerName().validate(validations);
getCommandline().validate(validations)
```

angelozerr: Server name and command can be null when language server definition comes from an extension point declared in an ij plugin

MituuZ:
@angelozerr Should I continue with the tab validations for the LSP console on this or should that be a new issue?

angelozerr:
Let is do that in an annoter pr. I wanted just to share with you an idea that I had had

src/main/java/com/redhat/devtools/lsp4ij/settings/ui/ValidatableDialog.java
```diff
@@ -0,0 +1,33 @@
...
+ *******************************************************************************/
+package com.redhat.devtools.lsp4ij.settings.ui;
+
+import com.intellij.openapi.project.Project;
+import com.intellij.openapi.ui.DialogWrapper;
+
+/**
+ * Shareable component for shared validations using DialogWrapper
+ */
+public abstract class ValidatableDialog extends DialogWrapper {
```
angelozerr, line 20: @fbricon what do you think about the  ValidatableDialog name?

src/main/java/com/redhat/devtools/lsp4ij/settings/ui/ValidatableConsoleWidget.java
```diff
@@ -0,0 +1,43 @@
...
+
+import javax.swing.*;
+import java.util.List;
+import java.awt.*;
+
+/**
+ * A shared interface meant to simplify creating validatable components used in
+ * NewLanguageServerDialog and LanguageServerView (LSP console)
+ */
+public interface ValidatableConsoleWidget {
```
angelozerr, line 26: @fbricon what do you think about the   ValidatableConsoleWidget name?

src/main/java/com/redhat/devtools/lsp4ij/settings/ui/ValidatableConsoleWidget.java
```diff
@@ -0,0 +1,43 @@
...
+    /**
+     * Set a common error border for the widget
+     * @param jComponent interface implementor (e.g. setErrorBorder(this);)
+     */
+    default void setErrorBorder(JComponent jComponent) {
```
angelozerr, line 31: setErrorBorder is more an utility class than an API.

Here some suggestion:

 * remove `setErrorBorder `method
 * add a new void `updateUi(boolean hasError)` method which is called after the call of validate (hasError must be comuted by coparing the size of validations before/after the cal of validate.
 * add a new method `initializeValidator(ValidateDialog)` which add the proper listener (which is done today in the addValidator)

src/main/java/com/redhat/devtools/lsp4ij/settings/ui/ServerNameWidget.java
```diff
@@ -0,0 +1,53 @@
...
+
+    @Override
+    public void validate(List<ValidationInfo> validations) {
+        List<ValidationInfo> widgetValidations = new ArrayList<>();
+
+        if (getDocument() != null && getText().isBlank()) {
+            widgetValidations.add(new ValidationInfo(errorMessage, this));
+        }
+
+        if (widgetValidations.isEmpty()) {
```
angelozerr, line 41: Don't do that here, move this code in the new `updateUi(boolean hasErrors)` method

src/main/java/com/redhat/devtools/lsp4ij/settings/ui/LanguageServerPanel.java
```diff
@@ -241,4 +256,25 @@ public ComboBox<ErrorReportingKind> getErrorReportingKindCombo() {
...
+        }
+        var commandLineWidget = getCommandLine();
+        if (commandLineWidget != null) {
+            commandLineWidget.validate(validations);
+        }
+        return validations;
+    }
+
+    private void addValidator(JTextComponent textComponent) {
+        textComponent.getDocument().addDocumentListener(new DocumentAdapter() {
```
angelozerr, line 273: Don't do that here but in the ValidableConsoleWidget impl in the `initializeValidator(ValidatableDialog dialog)` method

src/main/java/com/redhat/devtools/lsp4ij/settings/ui/ValidatableConsoleWidget.java
fbricon: fine by me

src/main/java/com/redhat/devtools/lsp4ij/settings/ui/ValidatableDialog.java
fbricon: It's fine

src/main/java/com/redhat/devtools/lsp4ij/settings/ui/LanguageServerPanel.java
angelozerr: I think addValidator should just add in a list CommandLineWidget, ServerNameWidget (list of ValidatableConsoleWidget) and in doValidateAll should just iterate from this list.

```java
private void addValidatableWidget(@Nullable ValidatableConsoleWidget validatableWidget) {
   if (validatableWidget == null) {
      return;
   }
   validatableWidget.initializeValidator(this);
   this.validatableWidgets.add(validatableWidget)
}
```

src/main/java/com/redhat/devtools/lsp4ij/settings/ui/LanguageServerPanel.java
```diff
@@ -241,4 +256,25 @@ public ComboBox<ErrorReportingKind> getErrorReportingKindCombo() {
         return errorReportingKindCombo;
     }
 
+    public @NotNull List<ValidationInfo> doValidateAll() {
```
angelozerr, line 259: This method should look like this:

```java
List<ValidationInfo> validations = new ArrayList<>();
for(var widget : validatableWidgets) {
    int previousSize = validations.size();
    widget.validate(validations);
    boolean hasErrors = previousSize ! validations.size();
   widget.updateUi(hasErrors);
}
return validations;
```

src/main/java/com/redhat/devtools/lsp4ij/settings/ui/ValidatableConsoleWidget.java
angelozerr: Thanks @fbricon . I think it should be renamed with `ValidatableWidget`

angelozerr:
@MituuZ I did some suggestion to improve API. If you have no time to work on it, let's me know and I will do that.

MituuZ:
@angelozerr 

I can keep working on this if you'd like, most likely tomorrow morning. Though I do think you have a very clear vision on how this should be finished.

It might easier for you to finish this, and I can continue on #215 , which I started earlier today.

angelozerr:
> It might easier for you to finish this

Ok I will take care of that.

angelozerr:
@MituuZ I have tried to work on this PR. At first using DialogWrapper is an excellent idea, thanks for that.

The thing which is bad in this PR is that we manage border (because JBTextArea doesn't manage border).. By managing border we loose some features like set a light red color when the JTectField is on error and loose the focus.

In otherwise, we must not manage border. I have tried to manage border only for JBTextArea, it starts working but when teh textarea is on error, the red border hide the cursor when cursor is on left of the textarea.

I need to investigate more this PR.

MituuZ:
Yeah, I feel you. 

Didn't want to handle the border manually either, but it seemed like the only way to get both fields working the same way and having the same styling for both. Didn't find what the default error borders are for the JBTextField so we could have copied them either.

angelozerr:
Here the awfull code that I'm using:

```java
public CommandLineWidget() {
        super(5, 0);
        super.setLineWrap(true);
        super.setWrapStyleWord(true);
        super.setFont(new JBTextField().getFont());
        super.setBorder(new JBTextField().getBorder());
        super.setMargin(new JBTextField().getMargin());
    }
```

When I will have something which will work, I will push my work.

MituuZ:
Checking this it suggests using red and light-red for highlight colors on validation errors. I believe I did try light red, but if recall correctly it was a different color than what the JBTextField had:

https://jetbrains.design/intellij/principles/validation_errors/#how-it-works-1

angelozerr:
> https://jetbrains.design/intellij/principles/validation_errors/#how-it-works-1

Great links, thanks.

I need to study more the topic.

PR #235, opened 2024-04-21
fix: update UD LS default command templates for Mac/Linux

fbricon:


angelozerr:
Thanks @fbricon !

PR #259, opened 2024-04-24
fix: Apply button is not enabled when mappings change

### LINKED ISSUE #219: Apply buton is not enabled when mappings change
In LSP console when you add / remove / change mappings, the Apply button is not availble.

I think there is a bug in https://github.com/redhat-developer/lsp4ij/blob/4829a156059d8395f5c8cc27a1b161ddd202b6cf/src/main/java/com/redhat/devtools/lsp4ij/settings/LanguageServerView.java#L101 (perhaps we should use Objects.deepEquals?)

fbricon:
fixes #219


angelozerr:
It works like a charm, thanks @fbricon !

PR #265, opened 2024-04-26
fix: remove last internal API usage in LSPIJUtils

fbricon:
Also fixes some warnings in that file


fbricon:
clearly not ready yet, since it broke a bunch of tests

PR #268, opened 2024-04-29
feat: Add edit action

### LINKED ISSUE #211: Provide an Edit action to open the Language Server preferences
It would be nice to provide an Edit action like we have for Delete:

image

to open the preferences and select the proper language server:

image

@MituuZ are you interested to work on this issue?

MituuZ:
Fixes #211 

Had some issues opening the correct ls and decided to go with an override on the reloadTree method by using the settings state.

Tried to get the configurable instance first, but didn't find any nice way to access it.


angelozerr:
I wonder if it should be better to use https://github.com/JetBrains/intellij-community/blob/dcd7375add6f58980d50f616324550b95790a6b1/platform/platform-api/src/com/intellij/openapi/options/ShowSettingsUtil.java#L33 like 

https://github.com/JetBrains/intellij-community/blob/dcd7375add6f58980d50f616324550b95790a6b1/python/src/com/jetbrains/python/debugger/variablesview/usertyperenderers/ConfigureTypeRenderersAction.kt#L105

does.

MituuZ:
I'll try with that, the setting does feel a bit hacky way of handling the selection

angelozerr:
> I'll try with that, 

Thanks.

> the setting does feel a bit hacky way of handling the selection

Indeed we should avoid doing that.

MituuZ:
Changed the implementation to not use settings as a workaround, but kept the same logic.

Using either of these throw this error after triggering the edit action and closing the editor, not sure why as the KotlinCompilerConfigurableTab is not explicitly used or modified in any way. Could it be just me?

```java
ShowSettingsUtil.getInstance().showSettingsDialog(project, LanguageServerListConfigurable.class);

ShowSettingsUtil.getInstance().showSettingsDialog(project, LanguageServerListConfigurable.class, languageServerListConfigurable ->
    languageServerListConfigurable.setDisplayNodeName(languageServerDefinition.getDisplayName());
});
```

```
java.lang.RuntimeException: Memory leak detected: 'org.jetbrains.kotlin.idea.base.compilerPreferences.configuration.KotlinCompilerConfigurableTab@7c039bb4' (class org.jetbrains.kotlin.idea.base.compilerPreferences.configuration.KotlinCompilerConfigurableTab) was registered in Disposer as a child of 'ROOT_DISPOSABLE' (class com.intellij.openapi.util.Disposer$2) but wasn't disposed.
```

src/main/java/com/redhat/devtools/lsp4ij/console/explorer/actions/EditServerAction.java
```diff
@@ -0,0 +1,50 @@
...
+import com.intellij.openapi.actionSystem.AnAction;
+import com.intellij.openapi.actionSystem.AnActionEvent;
+import com.intellij.openapi.options.ShowSettingsUtil;
+import com.intellij.openapi.project.Project;
+import com.redhat.devtools.lsp4ij.LanguageServerBundle;
+import com.redhat.devtools.lsp4ij.server.definition.LanguageServerDefinition;
+import com.redhat.devtools.lsp4ij.settings.LanguageServerListConfigurable;
+import org.jetbrains.annotations.NotNull;
+
+public class EditServerAction extends AnAction {
```
angelozerr, line 25: Add javadoc

angelozerr:
@MituuZ it works like a charm in my case.

Could you just add the missing javadoc in the new class please.

After that I'm OK to merge it.

> Could it be just me? I'll try with a second computer a bit later

I have not this problem.

angelozerr:
Thanks so much @MituuZ !

PR #295, opened 2024-05-21
fix: Copy-and-paste error in intention description

InSyncWithFoo:
Namely, `Refactor/description.html`:

https://github.com/redhat-developer/lsp4ij/blob/2a9cf025118279982a8e3a6e55e9d04696a78b01/src/main/resources/intentionDescriptions/Refactor/description.html#L3-L4


angelozerr:
Thanks @InSyncWithFoo !

PR #372, opened 2024-06-20
fix: actions not getting enabled on 2024.2

### LINKED ISSUE #371: Go To/ LSP * actions missing in CLion 2024.2 EAP
I'm working on migrating the LSP integration in my plugin onto lsp4ij, and I noticed that, inside a `runIDE` instance with CLion version set to `242-EAP-SNAPSHOT`, the `Go To -> LSP *` actions are missing, but in a `2024.1.4` instance, they're present.

Tested with lsp4ij snapshot `0.0.2-20240618-163640` from the marketplace nightly maven

FalsePattern:
`AnActionEvent.getData(CommonDataKeys.VIRTUAL_FILE)` is a lazily computed value, which are forbidden to be retrieved on the EDT since 2024.2 (implicitly returning a null VirtualFile, and logging a stacktrace).

This caused the `Go To -> LSP *` actions to stay perpetually disabled.

Fixes #371


angelozerr:
Thanks @FalsePattern !

PR #375, opened 2024-06-20
docs: Mention `order=""` in `signatureHelp` section

### LINKED ISSUE #373: Override default behaviour for `signatureHelp`
<kbd>Ctrl</kbd> + <kbd>P</kbd> doesn't trigger the language server's `textDocument/signatureHelp`, but PyCharm's default:

I already added the `<codeInsight.parameterInfo/>` element as guided by [the docs](https://github.com/redhat-developer/lsp4ij/blob/main/docs/LSPSupport.md#signature-help).

Am I missing something?

---

To reproduce, clone [this PR](https://github.com/InSyncWithFoo/pyright-for-pycharm/pull/42), run `runIde`, choose a Python interpreter (Python 3.12+), [install Pyright](https://insyncwithfoo.github.io/pyright-for-pycharm/how-to/#how-to-install-the-pyright-executables) and set up the path in <i>Settings</i>, then open a file and paste the following:

```python
from collections.abc import Callable

class Decorator[**P, T]:
	def __init__(self, f: Callable[P, T]) -> None:
		...
	def __call__(self, *args: P.args, **kwargs: P.kwargs) -> T:
		...

@Decorator
def add(x: int, y: int) -> int:
	return x + y
```

```python
add()  # < Place the cursor between the parentheses, then press Ctrl + P.
       # Notice that there are no `signatureHelp` requests sent.
```

Compare with VSCode:

InSyncWithFoo:
Fixes #373.

### DIFF
--- a/docs/LSPSupport.md
+++ b/docs/LSPSupport.md
@@ -329,7 +329,8 @@ the `codeInsight.parameterInfo` extension point. By default, LSP4IJ registers th
 ```
 
 If you use another language, you will have to declare `codeInsight.parameterInfo` with your language.
-To override the IDE's default behaviour, use `order="first"`:
+If `codeInsight.parameterInfo` for the language is already defined by another plugin or the IDE,
+you can use `order="first"` to gain priority:
 
 ```xml
 <codeInsight.parameterInfo
--- a/docs/LSPSupport.md
+++ b/docs/LSPSupport.md
@@ -323,12 +323,21 @@ the `codeInsight.parameterInfo` extension point. By default, LSP4IJ registers th
         implementationClass="com.redhat.devtools.lsp4ij.features.signatureHelp.LSPParameterInfoHandler"/>
 
 <codeInsight.parameterInfo
-id="LSPParameterInfoHandlerForTextMate"
-language="textmate"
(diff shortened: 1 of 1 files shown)


docs/LSPSupport.md
```diff
@@ -323,12 +323,21 @@ the `codeInsight.parameterInfo` extension point. By default, LSP4IJ registers th
...
-id="LSPParameterInfoHandlerForTextMate"
-language="textmate"
-implementationClass="com.redhat.devtools.lsp4ij.features.signatureHelp.LSPParameterInfoHandler"/>
+        id="LSPParameterInfoHandlerForTextMate"
+        language="textmate"
+        implementationClass="com.redhat.devtools.lsp4ij.features.signatureHelp.LSPParameterInfoHandler"/>
 ```
 
 If you use another language, you will have to declare `codeInsight.parameterInfo` with your language.
+To override the IDE's default behaviour, use `order="first"`:
```
angelozerr, line 332: I find using  `IDE's default behaviour` is hard to understand since the problem comes from when you install a plugin and it already defines the codeInsight.parameterInfo  for the language.

I would say a sentence like:

`if there is a plugin that already defines the `codeInsight.parameterInfo` for the language and you want to override it, you must use `order="first"``

angelozerr:
Thanks @InSyncWithFoo !

PR #405, opened 2024-07-04
fix: Semantic token color providers not being applied to language server definition

FalsePattern:
Fixed a typo in LanguageServersRegistry that ignored SemanticTokensColorsProviders during language server registration from extension points.


angelozerr:
Thanks @FalsePattern !

PR #412, opened 2024-07-06
fix: make Semantic Tokens editor read-only

fbricon:
Users should not be able to write in the semantic tokens editor (new content is deleted after being added). This PR makes the ST editor non-editable.


angelozerr:
Thanks @fbricon !

PR #414, opened 2024-07-07
fix: find templates directory when path contain spaces

### LINKED ISSUE #393: On MacOS, Server templates are missing unless LSP4IJ is run in dev mode
See:

Screenshot 2024-06-30 at 19 24 06

Templates are only showing up when LSP4IJ is started from source with `./gradlew runIde`

fbricon:
Fixes #393


angelozerr:
I have tested with my Windows OS and I can see template again.

Thanks @fbricon !

PR #434, opened 2024-07-30
feat: add the 'New Language Server' menu on explorer right-click

fbricon:
When no server is selected:
Screenshot 2024-07-30 at 19 01 58

When a server is selected:
Screenshot 2024-07-30 at 19 10 20


angelozerr:
Thanks @fbricon !

Issue #454, opened 2024-08-13
Unable to show diagnostics when renaming a currently-open file


bzoracler:
I'm trying to get my Language Server to analyse and re-publish diagnostics for a file that has been renamed using PyCharm's `Rename` dialogue (https://www.youtube.com/watch?v=IqPm9RYTNBM&t=32s). However, the diagnostics are not appearing, even though I've confirmed that they are being generated.

Looking at the `Traces`, I see that, upon getting PyCharm to rename the file, the following events are happening:

- ```
  Sending notification 'textDocument/didClose'
  Params: {
    "textDocument": {
      "uri": <URI before renaming>
    }
  }
  ```
- ```
  Sending notification 'workspace/didChangeWatchedFiles'
  Params: {
    "changes": [
      {
        "uri": <URI before renaming>,
        "type": 3
      },
      {
        "uri": <URI after renaming>,
        "type": 1
      }
    ]
  }
- ```
  Received notification 'textDocument/publishDiagnostics'
  Params: {
    "uri": <URI after renaming>,
    "diagnostics": [
      {
        "range": {
          "start": {
            "line": 22,
            "character": 8
          },
          "end": {
            "line": 22,
            "character": 9
          }
        },
        "severity": 1,
        "code": <...>,
        "source": <...>,
        "message": <...>
      },
      <...>
    ]
  }
  ```

However, the published diagnostics are not shown in the PyCharm editor.

I suspect the problem is that a `textDocument/didOpen` event on the `<URI after renaming>` is not being received by the server (or even being generated at all). None of the LSP methods seem to be registering changes on the opened file `<URI after renaming>` until after I close this file and re-open it.

Is there something I forgot to do on the server side to make this work? Could it be an OS-specific bug, e.g. with Windows Paths and capitalisation? (I didn't test this on any other OS's except Windows, as I don't have access to any others at the moment).

Environment:
- OS: Windows 11 Home 22631.3958
- PyCharm 2024.1.6, 2024.2
- LSP4IJ version: 0.3.0

---

EDIT: Forgot to mention - I'm using a User-defined language server, not a plugin.

angelozerr:
At first I suggest that you integrate your ls in vscode and study the LSP traces. 

Your LS should support https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_didRename to rename a file.

My fear is  that you have not LSP rename but an another rename dialog, and I'm not sure if it will work corretly.

[...]

PR #456, opened 2024-08-14
fix: document synchronisation after renaming actions

bzoracler:
Fixes https://github.com/redhat-developer/lsp4ij/issues/454

Both File Rename/Refactor events (changes the name of a file to a new name in the same directory) and File Move events (changes the directory of a file without changing its name) now trigger a textDocument/didOpen client-to-server notification, as specified in the [LSP specification's Renaming a document](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_didRename).

This allows the following without closing and re-opening the renamed file:
* Diagnostics to be re-reported in a renamed file which was not closed by the IDE after renaming (e.g., for Python files: `my_file.py` > `my_new_file.py`);
* Diagnostics to be newly reported if the file changes from a non-associated to an associated file (`e.g. my_file.txt` > `my_file.py`).


src/main/java/com/redhat/devtools/lsp4ij/LSPFileListener.java
```diff
@@ -153,6 +153,33 @@ private FileEvent fe(URI uri, FileChangeType type) {
...
+     * Implements the <a href=https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_didRename">
+     * LSP specification for document renaming</a>. Used for both file renames within the same directory and file shifts
+     * to a different directory.
+     *
+     * @param virtualParentFile directory of the file after renaming
+     * @param oldFileName file name before renaming
+     * @param virtualNewFile virtual file after renaming
+     * @return URI of the file before renaming
+     */
+    private @NotNull URI didRename(VirtualFile virtualParentFile, String oldFileName, VirtualFile virtualNewFile) {
```
bzoracler, line 158: I replaced `didClose` (https://github.com/redhat-developer/lsp4ij/pull/456/commits/f0f008d977f2450f0154217bd810a40959ef7018) with `didRename`, instead of making a separate `didOpen` method.

This is because I didn't think it made sense to have a stand-alone `didOpen` method for the renaming action. A `didOpen` action should only be triggered if we know that a previous document was closed. Otherwise, a rename or move on a non-open document in the IDE might also trigger `didOpen`.

angelozerr:
Thanks @bzoracler !

Is your PR fixes your issue? If yes let me try your PR with another language server (clojure, JDT LS, etc) that I used to test the feature.

It would be very nice to write test (perform a rename file in a test and track expected LSP requests like didOpen, didClose)  but it is not a trivial issue.

bzoracler:
Yes, I built the plugin and tested it locally. The diagnostics are now re-triggered in the renamed file without having to close and re-open the document.

I'm not very familiar with LSP4IJ or IntelliJ IDEs - if testing is a blocker to this PR, could you point me to an existing test or tests in LSP4IJ to get started?

angelozerr:
> I'm not very familiar with LSP4IJ or IntelliJ IDEs - if testing is a blocker to this PR, 

No it is not a blocker since I have no written tests for this feature.

> could you point me to an existing test or tests in LSP4IJ to get started?

LSP features Tests are located in https://github.com/redhat-developer/lsp4ij/tree/main/src/test/java/com/redhat/devtools/lsp4ij/features

The main idea is to emuluate the LSP response of an existing language server (ex : a rename symbol with go ls -> https://github.com/redhat-developer/lsp4ij/blob/main/src/test/java/com/redhat/devtools/lsp4ij/features/rename/GoRenameTest.java)

We should have this kind of same tests with rename files, but I had no time to do that. 

Let me test your PR with other language servers and check if it is working well. 

If you are motivated to write tests it should be very fantastic and we could do that in an another PR.

src/main/java/com/redhat/devtools/lsp4ij/LSPFileListener.java
angelozerr: Good catch!

angelozerr:
I have tested with some other language server and it seems it doesn't break something.

Great PR @bzoracler !

Do you think you could work on tests?

PR #481, opened 2024-08-27
fix: fix #477 semantichilight add a one-second debounce time.

### LINKED ISSUE #477: Semantic highlighting often flickers while editing code.
semantic

CppCXY:
after apply modify:
semantic


angelozerr:
Thanks @CppCXY for your contribution.

I understand more the problem with your PR and Ithink the best solution should not to use a debounce but highlight only tokens which are changing to avoid highlight again some lines which doesn't change.

What do you think about that?.

PR #524, opened 2024-09-13
chore: don't use internal ActionToolbarImpl

fbricon:
Fixes internal usage in 2023.3.eap:

Screenshot 2024-09-13 at 14 50 49


angelozerr:
Works like a cham. Thanks @fbricon !

angelozerr:
@fbricon I recreate the nighty build and it seems with eap we cznnot use the status internal annotation

Issue #417, opened 2024-07-10
Support for textDocument/documentSymbols


angelozerr:
It would be nice to provide an Outline filled with documentSymbol.

It seems this support was done in https://github.com/robocorp/robotframework-lsp/blob/master/robotframework-intellij/src/main/java/robocorp/lsp/intellij/FeatureOutline.java

angelozerr:
See comments about user who requires this feature https://plugins.jetbrains.com/plugin/24811-zowe-cobol-language-support/reviews#103378

PR #535, opened 2024-09-23
feat: add document symbol support

CppCXY:


src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPDocumentSymbolStructureViewFactory.java
```diff
@@ -0,0 +1,34 @@
+package com.redhat.devtools.lsp4ij.features.documentSymbol;
```
angelozerr, line 1: Add header (copy paste an existing hearder and add your name)

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPDocumentSymbolStructureViewModel.java
```diff
@@ -0,0 +1,217 @@
+package com.redhat.devtools.lsp4ij.features.documentSymbol;
```
angelozerr, line 1: Add header (copy paste an existing hearder and add your name)

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPDocumentSymbolSupport.java
```diff
@@ -0,0 +1,101 @@
+package com.redhat.devtools.lsp4ij.features.documentSymbol;
```
angelozerr, line 1: Add header (copy paste an existing hearder and add your name)

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPDocumentSymbolSupport.java
```diff
@@ -0,0 +1,101 @@
...
+        return getDocumentSymbols(file.getVirtualFile(), file.getProject(), documentSymbolParams, cancellationSupport);
+    }
+
+    private static @NotNull CompletableFuture<List<DocumentSymbolData>> getDocumentSymbols(@NotNull VirtualFile file,
+                                                                                           @NotNull Project project,
+                                                                                           @NotNull DocumentSymbolParams params,
+                                                                                           @NotNull CancellationSupport cancellationSupport) {
+
+        return LanguageServiceAccessor.getInstance(project)
+                .getLanguageServers(file, LanguageServerItem::isDocumentLinkSupported)
```
angelozerr, line 50: You need to create a new method `LanguageServerItem:#isDocumentSymbolSupported` to check the proper server capability

angelozerr:
Your PR looks promising. Let me time to test it and give you some feedback again. It would be nice if you could update the doc too at https://github.com/redhat-developer/lsp4ij/blob/main/docs/LSPSupport.md#language-features by enabling the [textDocument/documentSymbol](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_documentSymbol) and add a detail implementation.

You will have to uncomment the supported capability at https://github.com/redhat-developer/lsp4ij/blob/eca7ca6c67d155a96033d7b74d09836595962c8f/src/main/java/com/redhat/devtools/lsp4ij/internal/ClientCapabilitiesFactory.java#L175 and perhaps adapt it.

src/main/java/com/redhat/devtools/lsp4ij/LanguageServerItem.java
```diff
@@ -513,6 +513,11 @@ public SemanticTokensColorsProvider getSemanticTokensColorsProvider() {
         return getServerWrapper().getServerDefinition().getSemanticTokensColorsProvider();
     }
 
+    public static boolean isDocumentSymbolSupported(@Nullable ServerCapabilities serverCapabilities) {
```
angelozerr, line 516: Add Javadoc (like other *Supported method)

angelozerr:
@CppCXY please add in your commit (in the second line 

Fixes #417

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPDocumentSymbolStructureViewModel.java
```diff
@@ -0,0 +1,227 @@
...
+    private final PsiFile psiFile;
+    // Update queue to avoid multiple updates
+    private final MergingUpdateQueue updateQueue;
+
+    public LSPDocumentSymbolStructureViewModel(@NotNull PsiFile psiFile, @Nullable Editor editor) {
+        super(psiFile, editor, new LSPFileStructureViewElement(psiFile));
+        this.psiFile = psiFile;
+        this.updateQueue = new MergingUpdateQueue("lsp.documentSymbol", 300, true, null);
+        // Register psi tree change listener
+        registerPsiTreeChangeListener();
```
angelozerr, line 55: You add a listener, but you remove it never. I think you should override the dispose method of the LSPDocumentSymbolStructureViewModel and remove the listener here.

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPDocumentSymbolStructureViewFactory.java
```diff
@@ -0,0 +1,44 @@
...
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * LSP document symbol structure view factory.
+ */
+public class LSPDocumentSymbolStructureViewFactory implements PsiStructureViewFactory {
+
+    @Override
+    public @Nullable StructureViewBuilder getStructureViewBuilder(@NotNull PsiFile psiFile) {
+        return new LSPDocumentSymbolStructureViewBuilder(psiFile);
```
angelozerr, line 29: We need really take care of this LSP instanciation since we should never override existing StructureViewBuilder (we could override it with API, but let'start with this idea).

At first you need to check if the file is associated to a language server. So please add on the first line of  the method:

```java
if (!LanguageServersRegistry.getInstance().isFileSupported(psiFile)) {
            return null;
}
```

But it is not enough, we need to find if it exists a builder for teh given psi file. The usecase is for our IJ Quarkus, we have a language server which is associated to a Java language but LSP must not override the Java builder.

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPDocumentSymbolStructureViewModel.java
CppCXY: > You add a listener, but you remove it never. I think you should override the dispose method of the LSPDocumentSymbolStructureViewModel and remove the listener here.

the signature like this:
image
I've already passed in the parent, so I think explicit dispose shouldn't be necessary.

angelozerr: +1, I wil check debug if it is working. I think you need to override dipose to:

* cancel the LSPDocumentSymbolSupport.cancel() 
* cancel updateQueue like (not sure ?)

```java
updateQueue.cancelAllUpdates()
updateQueue.deactivate()
```

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPDocumentSymbolStructureViewFactory.java
angelozerr: Replace with (hope it will work for you)

```java
@Override
    public @Nullable StructureViewBuilder getStructureViewBuilder(@NotNull PsiFile psiFile) {
        if (!LanguageServersRegistry.getInstance().isFileSupported(psiFile)) {
            // the file is not associated to a language server
            return null;
        }
        if (LanguageStructureViewBuilder.INSTANCE.allForLanguage(psiFile.getLanguage()).size() > 1) {
            // A plugin already contributes with custom PsiStructureViewFactory, don't override it.
            return null;
        }
        return new LSPDocumentSymbolStructureViewBuilder(psiFile);
    }
```

src/main/resources/META-INF/plugin.xml
```diff
@@ -223,6 +223,12 @@
                 id="LSPDocumentLinkGotoDeclarationHandler"
                 implementation="com.redhat.devtools.lsp4ij.features.documentLink.LSPDocumentLinkGotoDeclarationHandler"/>
 
+        <!-- LSP textDocument/documentSymbol request support -->
+        <lang.psiStructureViewFactory
+                id="LSPDocumentSymbolStructureViewFactory"
+                language="textmate"
```
angelozerr, line 229: It will work only with textmate. We need to create those factory with Java code.

angelozerr:
> It would be nice if you could update the doc too at https://github.com/redhat-developer/lsp4ij/blob/main/docs/LSPSupport.md#language-features by enabling the [textDocument/documentSymbol](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_documentSymbol) and add a detail implementation.

If you have no time to do that,we could merge your PR and I will write it.

angelozerr:
@KUGDev I think this PR will interest you (if your language server supports documentSymbol)

CppCXY:
> If you have no time to do that,we could merge your PR and I will write it.

You can see that I rarely say long content. My Q&A mainly relies on translation, so writing long documents might be inexact

angelozerr:
> You can see that I rarely say long content. My Q&A mainly relies on translation, so writing long documents might be inexact

No problem,I will take care of that.

angelozerr:
I think we can merge the PR @CppCXY , right?

CppCXY:
> I think we can merge the PR @CppCXY , right?

yes

angelozerr:
It is a very good start, but we need to improve it like associate with other language (not only textmate) cancel the tree as soon as possible because my fear is with large file, cancel future when builder is disposed, write doc.

I will take care about that.

@CppCXY thanks so much for your great contribution!

PR #545, opened 2024-09-26
fix: Handle null section as a request for the entire configuration object

### LINKED ISSUE #466: Exception raised if `workspace/configuration` request from server to client does not get passed a `section`
When retrieving configuration settings through `workspace/configuration`, the language server specifications [allows leaving `section` unspecified](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_configuration):

```typescript
export interface ConfigurationItem {
	/**
	 * The scope to get the configuration section for.
	 */
	scopeUri?: URI;

	/**
	 * The configuration section asked for.
	 */
	section?: string;
}
```

However, if `section` is not provided, LSP4IJ raises an exception<sup>1</sup>.

I think a default behaviour should be implemented when `section` is not specified - such as (1) returning the entire configuration object, which is my preference, or (2) returning `null`.

[`com.redhat.devtools.lsp4ij.client.LanguageClientImpl.findSettings`](https://github.com/redhat-developer/lsp4ij/blob/main/src/main/java/com/redhat/devtools/lsp4ij/client/LanguageClientImpl.java)

```diff
    protected Object findSettings(String section) {
        var config = createSettings();
+       if (section == null) {
+           return config;
+       }
        if (config instanceof JsonObject json) {
            String[] sections = section.split("[.]");
            return findSettings(sections, json);
        }
        return null;
    }
```

---

1. Stack trace:
    https://github.com/redhat-developer/lsp4ij/blob/6a561e0c274211141a36b1c0eeeb1d3a056f1670/src/main/java/com/redhat/devtools/lsp4ij/client/LanguageClientImpl.java#L227
    ```
   Internal error: java.lang.NullPointerException: Cannot invoke "String.split(String)" because "section" is null

   java.util.concurrent.CompletionException: java.lang.NullPointerException: Cannot invoke "String.split(String)" because "section" is null
   	   at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:315)
   	   at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:320)
   	   at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1770)
   	   at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.exec(CompletableFuture.java:1760)
   	   at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:507)
   	   at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1491)
   	   at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:2073)
   	   at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:2035)
   	   at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:187)
   Caused by: java.lang.NullPointerException: Cannot invoke "String.split(String)" because "section" is null
   	   at com.redhat.devtools.lsp4ij.client.LanguageClientImpl.findSettings(LanguageClientImpl.java:227)
   	   at com.redhat.devtools.lsp4ij.client.LanguageClientImpl.lambda$configuration$9(LanguageClientImpl.java:203)
   	   at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1768)
   	   ... 6 more
    ```

bzoracler:
Fixes #466

### DIFF
--- a/src/main/java/com/redhat/devtools/lsp4ij/client/LanguageClientImpl.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/client/LanguageClientImpl.java
@@ -222,7 +222,7 @@ public class LanguageClientImpl implements LanguageClient, Disposable {
      *                settings object.
      * @return the settings retrieved by a valid section and null otherwise.
      */
-    protected Object findSettings(String section) {
+    protected Object findSettings(@Nullable String section) {
         var config = createSettings();
         if (config instanceof JsonObject json) {
             if (section == null) {
--- a/src/main/java/com/redhat/devtools/lsp4ij/client/LanguageClientImpl.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/client/LanguageClientImpl.java
@@ -218,12 +218,16 @@ public class LanguageClientImpl implements LanguageClient, Disposable {
      * must return an instance of {@link JsonObject} but this method can be overridden to return another structure.
      * </p>
      *
-     * @param section the section.
-     * @return the settings retrieved by the given section and null otherwise.
+     * @param section the section. In the base implementation, a null section indicates a request for the entire
+     *                settings object.
+     * @return the settings retrieved by a valid section and null otherwise.
      */
     protected Object findSettings(String section) {
         var config = createSettings();
         if (config instanceof JsonObject json) {
+            if (section == null) {
+                return config;
+            }
             String[] sections = section.split("[.]");
             return findSettings(sections, json);
         }


src/main/java/com/redhat/devtools/lsp4ij/client/LanguageClientImpl.java
```diff
@@ -218,12 +218,16 @@ public CompletableFuture<List<Object>> configuration(ConfigurationParams params)
      * must return an instance of {@link JsonObject} but this method can be overridden to return another structure.
      * </p>
      *
-     * @param section the section.
-     * @return the settings retrieved by the given section and null otherwise.
+     * @param section the section. In the base implementation, a null section indicates a request for the entire
+     *                settings object.
+     * @return the settings retrieved by a valid section and null otherwise.
      */
     protected Object findSettings(String section) {
```
bzoracler, line 225: I did not do `@Nullable String section` in the parameters here, because that indicates subclasses should also accept `null` sections.

angelozerr: If method can receive null I think it should be better to use @nullable annotation, no?

bzoracler: `findSettings` is `protected`, so if we add `@Nullable` here, I understand this to mean all third-party plugins which have subclassed `LanguageClientImpl` and overridden `findSettings` to also be `@Nullable`.

I don't know the downstream (subclassing) behaviour of adding `@Nullable` in Java here - in other languages, this would violate liskov substitution and cause compile-time errors. If you're not concerned about this, I can add `@Nullable`.

angelozerr: I have never seen some project which extends findSettings, so I would prefer providing a clean API. Si please add @Nullable

angelozerr:
Thanks @bzoracler !

PR #590, opened 2024-10-31
update RADME: Add Vespa Schema Language Support

theodorklauritzen:
Add Vespa Schema Language Support to the list of who is using LSP4IJ in the README.


fbricon:
Thanks @theodorklauritzen

PR #619, opened 2024-11-24
feat: Add support for Next/Previous Method actions

SCWells72:
Issue 618 - This provides a simple implementation of **Next/Previous Method** actions based on the existing structure view implementation by returning text offsets for name identifiers of methods and functions from the structure view. Ideally it would be implemented via `MethodNavigationOffsetProvider` -- and perhaps it should be -- but the only down-side to this approach is that when the **Previous Method** action is triggered from the first method/function in the file, it does navigate to the beginning of the file because text offset 0 is returned for other elements.

### DIFF
--- a/src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPDocumentSymbolFeature.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPDocumentSymbolFeature.java
@@ -99,7 +99,12 @@ public class LSPDocumentSymbolFeature extends AbstractLSPDocumentFeature {
                              @NotNull PsiFile psiFile) {
         SymbolKind symbolKind = documentSymbol.getKind();
         // TODO: This works quite well without having to implement MethodNavigationOffsetProvider
-        if ((symbolKind == SymbolKind.Method) || (symbolKind == SymbolKind.Function)) {
+        if ((symbolKind == SymbolKind.Class) ||
+            (symbolKind == SymbolKind.Interface) ||
+            (symbolKind == SymbolKind.Enum) ||
+            (symbolKind == SymbolKind.Struct) ||
+            (symbolKind == SymbolKind.Method) ||
+            (symbolKind == SymbolKind.Function)) {
             Position startPosition = documentSymbol.getSelectionRange().getStart();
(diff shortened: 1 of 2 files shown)


src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPDocumentSymbolFeature.java
```diff
@@ -92,6 +95,20 @@ public StructureViewTreeElement getStructureViewTreeElement(@NotNull DocumentSym
         return documentSymbol.getDetail();
     }
 
+    public int getTextOffset(@NotNull DocumentSymbol documentSymbol,
```
angelozerr, line 98: Add javadoc

angelozerr (changes requested):

src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPDocumentSymbolFeature.java
```diff
@@ -92,6 +95,20 @@ public StructureViewTreeElement getStructureViewTreeElement(@NotNull DocumentSym
         return documentSymbol.getDetail();
     }
 
+    public int getTextOffset(@NotNull DocumentSymbol documentSymbol,
+                             @NotNull PsiFile psiFile) {
+        SymbolKind symbolKind = documentSymbol.getKind();
+        // TODO: This works quite well without having to implement MethodNavigationOffsetProvider
+        if ((symbolKind == SymbolKind.Method) || (symbolKind == SymbolKind.Function)) {
+            Position startPosition = documentSymbol.getSelectionRange().getStart();
```
angelozerr, line 103: Selection range could be null

angelozerr (changes requested):

src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPDocumentSymbolFeature.java
angelozerr: Indeed but the idea is you must not trust the language server.

angelozerr:
Thanks @SCWells72 !

PR #625, opened 2024-11-25
feat: Implement "gotoClassContributor" for LSP symbols (issue 624)

SCWells72:
Issue 624 - Refactored `LSPWorkspaceSymbolContributor` into an abstract base class and added concrete implementations for both `gotoSymbolContributor` (functionally identical to before) and `gotoClassContributor`. The latter only includes symbols that are classes, interfaces, enums, and structs.


src/main/java/com/redhat/devtools/lsp4ij/features/workspaceSymbol/AbstractLSPWorkspaceSymbolContributor.java
```diff
@@ -1,109 +1,130 @@
...
+/*******************************************************************************
+ * Copyright (c) 2024 Red Hat, Inc.
+ * Distributed under license by Red Hat, Inc. All rights reserved.
+ * This program is made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution,
+ * and is available at https://www.eclipse.org/legal/epl-v20.html
+ *
+ * Contributors:
+ * Red Hat, Inc. - initial API and implementation
+ ******************************************************************************/
```
angelozerr, line 10: Add your name here and email if you wish

angelozerr: @fbricon what do you think about this idea?

src/main/java/com/redhat/devtools/lsp4ij/features/workspaceSymbol/AbstractLSPWorkspaceSymbolContributor.java
```diff
@@ -1,109 +1,130 @@
...
+
+/**
+ * Abstract base class for LSP goto symbol contributors
+ */
+abstract class AbstractLSPWorkspaceSymbolContributor implements ChooseByNameContributorEx {
+
+    private final Logger LOGGER;
+
+    protected AbstractLSPWorkspaceSymbolContributor(@NotNull Logger logger) {
+        this.LOGGER = logger;
```
angelozerr, line 45: Don't set the LOGGER here. The logger should use the AbstractLSPWorkspaceSymbolContributor  LOGGER.

SCWells72: I can certainly do that, but this is a pretty common idiom in inheritance hierarchies that allows each concrete subclass it have its own distinct logging channel that can be configured independently including inherited behavior. If you'd prefer that logging for both subclasses be configured via a single logger in `AbstractLSPWorkspaceSymbolContributor`, I can just remove the subclass loggers and restore the base class logger. Let me know your preference either way.

angelozerr: yes please to be consistent with other code

angelozerr: Thanks!

src/main/java/com/redhat/devtools/lsp4ij/features/workspaceSymbol/AbstractLSPWorkspaceSymbolContributor.java
```diff
@@ -1,109 +1,127 @@
...
+    }
+
+    private List<WorkspaceSymbolData> getWorkspaceSymbols(@NotNull String name, boolean cancel, Project project) {
+        // Consume LSP 'workspace/symbol' request
+        LSPWorkspaceSymbolSupport workspaceSymbolSupport = LSPWorkspaceSupport.getSupport(project).getWorkspaceSymbolSupport();
+        if (cancel) {
+            workspaceSymbolSupport.cancel();
+        }
+        WorkspaceSymbolParams params = new WorkspaceSymbolParams(name);
+        CompletableFuture<List<WorkspaceSymbolData>> workspaceSymbolFuture = workspaceSymbolSupport.getWorkspaceSymbol(params);
```
angelozerr, line 86: An annoying problem that I see (with our IJ Quarkus) is that our workspace symbol doesn't manage Class, Struct, etc and  our workspace symbol can takes some times to collect workspace symbol which scans all classpath to collect REST services in Java.

In Java, the Classes search is a lot of used and it means that we will execute the workspace symbol from our IJ Quarkus although we are sure that it will not report some Classes and it will impact performances of Class search.

To avoid this problem, it would be nice to add a new method like `LSPWorkspaceSymbolFeature#canReportClasses()` (@fbricon what do you think about the name) which will return true if the workspace symbol must be executed when user will search Classes and false otherwise.

By default it should return false. For user defined language server definition, we could add a new settings (same settings that caseSensitive completion like you want to add) to know if language server can report classes by executing workspace symbol.

SCWells72: @angelozerr, I would propose that the default for such a setting should be true as the potential performance issue you've described seems to be the exception and most language servers could presumably respond quickly.

angelozerr: I think we should set to false by default because I fear about workspace symbol performance from language servers.

And for user defined language servers we should use a settings, but if you want to do that in an another PR, you could set to true for UserDefinedLanguageServers for the moment.

src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPWorkspaceSymbolFeature.java
```diff
@@ -40,4 +40,19 @@ public boolean isSupported() {
...
+     * Determines whether or not the language server's workspace/symbol feature implementation is efficient enough to
+     * support the JetBrains' IDE's gotoClassContributor EP. In general, the action associated with that EP is invoked
+     * much more frequently than the action associated with the IDE's gotoSymbolContributor, so this method should only
+     * return true if the language server's implementation of workspace/symbol can realistically accommodate those
+     * requests.
+     *
+     * @return true if the language server's workspace/symbol feature can support the IDE's gotoClassContributor EP
+     * efficiently; otherwise false
+     */
+    public boolean supportsGotoClass() {
```
SCWells72, line 54: This is the method that determines whether or not the workspace symbol feature (efficiently) supports the IDE's Go To Class action. I'm open to a name change, but ultimately that's the question that's being asked, so I named it accordingly. The default response is `false` for backward-compatibility.

src/main/java/com/redhat/devtools/lsp4ij/features/workspaceSymbol/AbstractLSPWorkspaceSymbolContributor.java
```diff
@@ -1,109 +1,133 @@
...
+/**
+ * Abstract base class for LSP goto symbol contributors
+ */
+abstract class AbstractLSPWorkspaceSymbolContributor implements ChooseByNameContributorEx {
+
+    private final static Logger LOGGER = LoggerFactory.getLogger(AbstractLSPWorkspaceSymbolContributor.class);
+
+    private final LSPWorkspaceRequestedSymbolTypes requestedSymbolTypes;
+
+    protected AbstractLSPWorkspaceSymbolContributor(@NotNull LSPWorkspaceRequestedSymbolTypes requestedSymbolTypes) {
```
SCWells72, line 47: The concrete implementation of this must specify what types of symbols it is requesting, currently "All" vs. "Types".

src/main/java/com/redhat/devtools/lsp4ij/features/workspaceSymbol/AbstractLSPWorkspaceSymbolContributor.java
```diff
@@ -1,109 +1,133 @@
...
+        }
+    }
+
+    private List<WorkspaceSymbolData> getWorkspaceSymbols(@NotNull String name, boolean cancel, Project project) {
+        // Consume LSP 'workspace/symbol' request
+        LSPWorkspaceSymbolSupport workspaceSymbolSupport = LSPWorkspaceSupport.getSupport(project).getWorkspaceSymbolSupport();
+        if (cancel) {
+            workspaceSymbolSupport.cancel();
+        }
+        WorkspaceSymbolParams params = new LSPWorkspaceSymbolParams(name, requestedSymbolTypes);
```
SCWells72, line 91: This is how we convey the types of symbols being requested from here to `LSPWorkspaceSymbolSupport`. We construct a specialized subclass of the standard `WorkspaceSymbolParams` that contains both the query and the types of symbols being requested. The latter is used to filter language servers based on whether or not they can efficiently support Go To Class-style invocation.

src/main/java/com/redhat/devtools/lsp4ij/features/workspaceSymbol/LSPWorkspaceGotoClassContributor.java
```diff
@@ -0,0 +1,39 @@
...
+
+    private final static Set<SymbolKind> TYPE_SYMBOL_KINDS = Set.of(
+            SymbolKind.Class,
+            SymbolKind.Interface,
+            SymbolKind.Enum,
+            SymbolKind.Struct
+    );
+
+    public LSPWorkspaceGotoClassContributor() {
+        super(LSPWorkspaceRequestedSymbolTypes.TYPE_SYMBOLS);
```
SCWells72, line 31: Types-only.

src/main/java/com/redhat/devtools/lsp4ij/features/workspaceSymbol/LSPWorkspaceGotoSymbolContributor.java
```diff
@@ -0,0 +1,29 @@
...
+
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * LSP workspace goto symbol contributor.
+ */
+public class LSPWorkspaceGotoSymbolContributor extends AbstractLSPWorkspaceSymbolContributor {
+
+    public LSPWorkspaceGotoSymbolContributor() {
+        super(LSPWorkspaceRequestedSymbolTypes.ALL_SYMBOLS);
```
SCWells72, line 21: All symbols.

src/main/java/com/redhat/devtools/lsp4ij/features/workspaceSymbol/LSPWorkspaceRequestedSymbolTypes.java
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * Contributors:
+ * Red Hat, Inc. - initial API and implementation
+ ******************************************************************************/
+package com.redhat.devtools.lsp4ij.features.workspaceSymbol;
+
+/**
+ * The types of workspace symbols being requested in {@link LSPWorkspaceSymbolParams}.
+ */
+enum LSPWorkspaceRequestedSymbolTypes {
```
SCWells72, line 16: This is the symbolic discriminator that's used to convey the originating action to the actual LSP engagement.

src/main/java/com/redhat/devtools/lsp4ij/features/workspaceSymbol/LSPWorkspaceSymbolParams.java
```diff
@@ -0,0 +1,36 @@
...
+ ******************************************************************************/
+package com.redhat.devtools.lsp4ij.features.workspaceSymbol;
+
+import org.eclipse.lsp4j.WorkspaceSymbolParams;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Extension of {@link WorkspaceSymbolParams} that includes additional parameters specific to LSP4IJ.
+ */
+class LSPWorkspaceSymbolParams extends WorkspaceSymbolParams {
```
SCWells72, line 19: And this is the subclass of `WorkspaceSymbolParams` that can carry along that discriminator.

src/main/java/com/redhat/devtools/lsp4ij/features/workspaceSymbol/LSPWorkspaceSymbolSupport.java
```diff
@@ -54,16 +54,26 @@ protected CompletableFuture<List<WorkspaceSymbolData>> doLoad(WorkspaceSymbolPar
                                                                                             @NotNull WorkspaceSymbolParams params,
                                                                                             @NotNull CancellationSupport cancellationSupport) {
         return getLanguageServers(project,
-                        f -> f.getWorkspaceSymbolFeature().isEnabled(),
-                        f -> f.getWorkspaceSymbolFeature().isSupported())
+                f -> f.getWorkspaceSymbolFeature().isEnabled(),
+                f -> f.getWorkspaceSymbolFeature().isSupported())
                 .thenComposeAsync(languageServers -> {
+                    // Remove language servers that aren't supported efficiently by the provided params
```
SCWells72, line 60: And this is where we use the discriminator to filter out any language servers that don't support types-only queries efficiently when invoked from the Go To Class action. I'm assuming the provided collection should be considered immutable so I'm making a mutable copy before removing filtered language servers.

src/main/java/com/redhat/devtools/lsp4ij/server/definition/launching/ClientConfigurationSettings.java
```diff
@@ -25,12 +25,27 @@ public class ClientConfigurationSettings {
...
          * Whether or not completions should be offered as case-sensitive. Defaults to false.
-         **/
+         */
         public boolean caseSensitive = false;
     }
 
+    /**
+     * Client-side code workspace symbol settings.
+     */
+    public static class ClientConfigurationWorkspaceSymbolSettings {
```
SCWells72, line 35: Client-side config additions. Again, I'm not married to these names, so feel free to suggest alternatives.

src/main/java/com/redhat/devtools/lsp4ij/server/definition/launching/UserDefinedWorkspaceSymbolFeature.java
```diff
@@ -0,0 +1,29 @@
...
+
+import com.redhat.devtools.lsp4ij.client.features.LSPWorkspaceSymbolFeature;
+
+/**
+ * Adds client-side workspace symbol configuration features.
+ */
+public class UserDefinedWorkspaceSymbolFeature extends LSPWorkspaceSymbolFeature {
+
+    @Override
+    public boolean supportsGotoClass() {
```
SCWells72, line 24: Following the same pattern as before where the implementation that consults client config is in the `UserDefined*` subclass, and it's added in the same manner.

src/main/java/com/redhat/devtools/lsp4ij/server/definition/launching/UserDefinedClientFeatures.java
```diff
@@ -22,7 +22,9 @@ public class UserDefinedClientFeatures extends LSPClientFeatures {
 
     public UserDefinedClientFeatures() {
         super();
-        // Add our completion feature
+
+        // Use the extended feature implementations
         setCompletionFeature(new UserDefinedCompletionFeature());
+        setWorkspaceSymbolFeature(new UserDefinedWorkspaceSymbolFeature());
```
SCWells72, line 28: Using the extended workspace symbol feature just as we did the extended completion feature.

src/main/resources/META-INF/plugin.xml
```diff
@@ -376,8 +376,10 @@
                 implementation="com.redhat.devtools.lsp4ij.usages.LSPUsageTypeProvider"/>
 
         <!-- LSP workspace/symbol request support -->
-        <gotoSymbolContributor id="LSPWorkspaceSymbolContributor"
-                               implementation="com.redhat.devtools.lsp4ij.features.workspaceSymbol.LSPWorkspaceSymbolContributor"/>
+        <gotoClassContributor id="LSPWorkspaceGotoClassContributor"
```
SCWells72, line 379: `gotoClassContributor` EP registration.

src/main/resources/jsonSchema/clientSettings.schema.json
```diff
@@ -17,6 +17,19 @@
           "default": false
         }
       }
+    },
+    "workspaceSymbols": {
```
SCWells72, line 21: Client config JSON schema updates.

src/main/resources/templates/clojure-lsp/clientSettings.json
```diff
@@ -1,5 +1,8 @@
 {
   "completions": {
     "caseSensitive": true
+  },
+  "workspaceSymbol": {
```
SCWells72, line 5: Added default for all bundled templates with the feature enabled for languages that have the notion of a type and disabled for those that don't.

src/main/java/com/redhat/devtools/lsp4ij/features/workspaceSymbol/LSPWorkspaceSymbolParams.java
```diff
@@ -0,0 +1,36 @@
...
+    private final LSPWorkspaceRequestedSymbolTypes requestedSymbolTypes;
```
angelozerr, line 20: Use transient to avoid having this non lsp field in the JSON lsp request.

angelozerr (changes requested):

src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPWorkspaceSymbolFeature.java
angelozerr, line 54: Please add this new methods in the docs/LspApi.md

angelozerr (changes requested):

src/main/java/com/redhat/devtools/lsp4ij/settings/ui/JsonTextField.java
```diff
@@ -86,8 +87,8 @@ public void setJsonFilename(@NotNull String jsonFilename) {
 
     // Proxy some simple accessors to the editor text field
 
-    public void setText(@NotNull String text) {
-        editorTextField.setText(text);
+    public void setText(@Nullable String text) {
```
SCWells72, line 90: Made this more flexible with regard to `null` values which result in a valid default JSON document for the text editor.

src/main/java/com/redhat/devtools/lsp4ij/server/definition/launching/UserDefinedLanguageServerDefinition.java
```diff
@@ -69,6 +69,26 @@ public UserDefinedLanguageServerDefinition(@NotNull String id,
         this.clientConfigurationContent = clientConfigurationContent;
     }
 
+    // Backward-compatible signature for clients calling without client configuration content
+    public UserDefinedLanguageServerDefinition(@NotNull String id,
```
SCWells72, line 73: The comment says it all, but because this is part of the public API -- and is in fact used by my own plugin to assist with language server config -- the original constructor signature must be restored/retained, passing a null value for `clientConfiguraitonContent`.

angelozerr:
Your PR doesn't work (the fix is to rename workspaceSymbols field to workspaceSymbol of the settings).

And I think we can simplify your code, by moving filter in the LSPWorkspaceSymbolParms. We can also avoid creating WorkspaceSymbolData if it the symbol kind doesn't manatch Types  (forgo to class).

@SCWells72 please see my PR https://github.com/redhat-developer/lsp4ij/pull/667/commits/86db3f5b7fbe719171deb472ef1a8717f8b618cd which follows the idea and update your PR with my suggestion if you like it (I did that very quickly,if you want rename some method, don't hesitate and please add javadoc).

angelozerr:
Thanks @SCWells72 !

PR #635, opened 2024-11-27
fix: Make go to definition/declaration/implementation target the name identifier (issue 628)

SCWells72:
Issue 628 - When available, the target selection range should be used from the location link instead of the target range. That helps place the caret at the beginning of the target name identifier instead of the beginning of the entire target declaration. This logic has been extracted into a utility method, and all existing implementations of the same logic now use that method.


angelozerr:
Great job @SCWells72 !

PR #637, opened 2024-11-27
feat: Add support for the IDE's native Go To Implementation(s) feature (issue 636)

SCWells72:
@angelozerr this provides first-class support for Go To Implementation(s), e.g., for a type:

image

and for a method:

image

The real trick is in `LSPTargetElementEvaluator` where I find the highlighted identifier "element" at the current offset.

I don't know if the new classes here are where you'd prefer, or if you have an alternative way to infer the current element. My guess is that there will be some iteration, but hopefully this not only provides the foundation for this feature, but also for other features that depend on the inferred target element at the caret.

Let me know your thoughts.


src/main/java/com/redhat/devtools/lsp4ij/features/highlight/LSPHighlightUsagesHandlerFactory.java
```diff
@@ -50,11 +50,11 @@ public class LSPHighlightUsagesHandlerFactory implements HighlightUsagesHandlerF
     private static final Logger LOGGER = LoggerFactory.getLogger(LSPHighlightUsagesHandlerFactory.class);
 
     @Override
-    public @Nullable HighlightUsagesHandlerBase createHighlightUsagesHandler(@NotNull Editor editor, @NotNull PsiFile file) {
+    public @Nullable HighlightUsagesHandlerBase<LSPHighlightPsiElement> createHighlightUsagesHandler(@NotNull Editor editor, @NotNull PsiFile file) {
```
SCWells72, line 53: Added type parameters in a few places to get rid of Java warnings.

src/main/java/com/redhat/devtools/lsp4ij/features/highlight/LSPHighlightUsagesHandlerFactory.java
```diff
@@ -50,11 +50,11 @@ public class LSPHighlightUsagesHandlerFactory implements HighlightUsagesHandlerF
...
         if (!LanguageServersRegistry.getInstance().isFileSupported(file)) {
             return null;
         }
-        if (ProjectIndexingManager.getInstance(file.getProject()).isIndexingAll()) {
+        if (ProjectIndexingManager.isIndexingAll()) {
```
SCWells72, line 57: Simplified call to a class method via an instance.

src/main/java/com/redhat/devtools/lsp4ij/features/highlight/LSPTargetElementEvaluator.java
```diff
@@ -0,0 +1,33 @@
...
+    public @Nullable PsiElement adjustReferenceOrReferencedElement(@NotNull PsiFile file, @NotNull Editor editor, int offset, int flags, @Nullable PsiElement refElement) {
+        if (!LanguageServersRegistry.getInstance().isFileSupported(file)) {
+            return null;
+        }
+
+        if (ProjectIndexingManager.isIndexingAll()) {
+            return null;
+        }
+
+        // Try to find the highlighted identifier at the caret and return that fake PSI element
```
SCWells72, line 28: Let me know if you think there's a better way to discern the identifier at the current offset.

SCWells72:
I could also look at adding support for the corresponding gutter icons as part of this work if you'd like.

Also, is there already a way to know the base declarations, i.e., the extended/implemented base classes/interfaces or the overridden base method? If so, it would be simple to provide native IDE support for Go To Super so there's bidirectional inheritance hierarchy navigation from within the editor.

angelozerr:
Your PR seems amazing but I am not a master of IJ so I will need to take some time to debug it and play with several language servers.

My main goal now is to do the release, so I will review your PR after the release.

SCWells72:
@angelozerr No hurry at all. When I get something in my head, I have to work through it! Feel free to send your feedback when you're ready. Focus on the release for now. That's the priority.

angelozerr:
> @angelozerr No hurry at all. When I get something in my head, I have to work through it! Feel free to send your feedback when you're ready. Focus on the release for now. That's the priority.

Thanks for your understand. But it seems yiu have so many ideas (I love that) and please keep your motivation. 

Another point is the documentation. Each time I implement a feature I write doc with demo that yiu can see in the lsp support md or user guide.

It would be nice to have a doc with a little demo which shows in action your feature.

SCWells72:
> And a point which is very important is to never override standard IJ features. For instance when I have implement Find Usages, I had overrided the standard find usages and broke the find usages with Java. It was so bad.

Oh, don't worry about that. I've had a commercial plugin available for just shy of 10 years now that implements very complete support for a set of custom languages. I've also learned the hard way the dos and don'ts of JetBrains plugin dev! My goal is always to make things work as closely as possible to how they do in JetBrains' own first-class language support and not to interfere with what's already there.

angelozerr:
> > And a point which is very important is to never override standard IJ features. For instance when I have implement Find Usages, I had overrided the standard find usages and broke the find usages with Java. It was so bad.
> 
> Oh, don't worry about that. I've had a commercial plugin available for just shy of 10 years now that implements very complete support for a set of custom languages. I've also learned the hard way the dos and don'ts of JetBrains plugin dev! My goal is always to make things work as closely as possible to how they do in JetBrains' own first-class language support and not to interfere with what's already there.

Wow that is fantastic! I am very Lucky. In my case I work on IJ since just one year, I discover the IJ platform step by step when I implement features.

Hope you will like how I have implemented lsp features. If you think we need to break something we are totally opened to do some refactoring.

src/main/java/com/redhat/devtools/lsp4ij/features/highlight/LSPTargetElementEvaluator.java
```diff
@@ -0,0 +1,33 @@
+package com.redhat.devtools.lsp4ij.features.highlight;
```
angelozerr, line 1: Add header

src/main/java/com/redhat/devtools/lsp4ij/features/highlight/LSPTargetElementEvaluator.java
```diff
@@ -0,0 +1,33 @@
...
+import com.intellij.util.containers.ContainerUtil;
+import com.redhat.devtools.lsp4ij.LanguageServersRegistry;
+import com.redhat.devtools.lsp4ij.client.ProjectIndexingManager;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.Collections;
+import java.util.List;
+
+public class LSPTargetElementEvaluator extends TargetElementEvaluatorEx2 {
```
angelozerr, line 17: Add javadoc

angelozerr (changes requested):

src/main/java/com/redhat/devtools/lsp4ij/features/highlight/LSPTargetElementEvaluator.java
angelozerr: The main problem with your strategy is that if your language server doesn't support `textDocument/documentHighlight` it will not work.

If I understand correctly your code, you need to create a fake PsiElement according a given offset with a text range. Right?

If it that I do that to support find usages at https://github.com/redhat-developer/lsp4ij/blob/main/src/main/java/com/redhat/devtools/lsp4ij/usages/LSPUsageTargetProvider.java#L69-L82

If you think it is the same case, I suggest that you create a static method in LSP4IJ that you can use in both cases (in find usage target and in your case).

angelozerr:
> I could also look at adding support for the corresponding gutter icons as part of this work if you'd like.

It would be fantatsic!

> 
> Also, is there already a way to know the base declarations, i.e., the extended/implemented base classes/interfaces or the overridden base method? If so, it would be simple to provide native IDE support for Go To Super so there's bidirectional inheritance hierarchy navigation from within the editor.

You mean calling `textDocument/implementation`?

src/main/java/com/redhat/devtools/lsp4ij/LSPIJUtils.java
```diff
@@ -1221,4 +1221,39 @@ public static String getProjectUri(Project project) {
...
+
+    /**
+     * Finds the text range of the token at the specified offset in the provided file and editor.
+     *
+     * @param file   the file
+     * @param editor the editor
+     * @param offset the offset within the file
+     * @return the token at the specified offset in the file and editor, or null if no token is found
+     */
+    public static @Nullable TextRange getTokenTextRange(@NotNull PsiFile file, @NotNull Editor editor, int offset) {
```
angelozerr, line 1233: My first impression is that this method is similar to getWordRangeAt. I wonder if we could just using getWordRangeAt directly.

src/main/java/com/redhat/devtools/lsp4ij/features/workspaceSymbol/LSPWorkspaceImplementationsSearch.java
```diff
@@ -0,0 +1,102 @@
+package com.redhat.devtools.lsp4ij.features.workspaceSymbol;
```
angelozerr, line 1: Add header

angelozerr:
I have played with your PR quickly and the result looks very promising!

I wonder if it is possible to show `No implementations found` when there are no result?

src/main/java/com/redhat/devtools/lsp4ij/features/workspaceSymbol/LSPWorkspaceImplementationsSearch.java
```diff
@@ -0,0 +1,142 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Red Hat, Inc.
```
angelozerr, line 2: 2024

src/main/java/com/redhat/devtools/lsp4ij/features/workspaceSymbol/LSPWorkspaceImplementationsSearch.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2019 Red Hat, Inc.
+ * Copyright (c) 2024 Red Hat, Inc.
```
fbricon, line 2: we only need to change the date when the file changes, else you'd have to change 100% of all files every year

angelozerr:
@SCWells72 please rebase your PR from the mainto avoid compilation problem https://github.com/redhat-developer/lsp4ij/actions/runs/12087432432/job/33708863034?pr=637#step:5:59

angelozerr:
@SCWells72 your PR seems broken onetest that we have with completion https://github.com/redhat-developer/lsp4ij/actions/runs/12058597289/job/33727177314?pr=639#step:5:154 I suspect that its because of change of getRangeAt (not sure with that).

SCWells72:
> @SCWells72 your PR seems broken onetest that we have with completion https://github.com/redhat-developer/lsp4ij/actions/runs/12058597289/job/33727177314?pr=639#step:5:154 I suspect that its because of change of getRangeAt (not sure with that).

@angelozerr, I'm not seeing that test fail locally, but I am seeing three tests from `TypeScriptHoverTest` fail from `LSPTargetElementEvaluator` and am investigating them now. Any idea why there would be a difference in my test results locally and those in the CI build?

SCWells72:
@angelozerr the issue with the three failing tets in `TypeScriptHoverTest` is with the following from `LSPHoverFixtureTestCase#assertHover()`, lines 72-76:

```java
        List<? extends @NotNull DocumentationTarget> targets = IdeDocumentationTargetProvider.getInstance(myFixture.getProject()).documentationTargets(editor, file, offset);
        if (expectedHoverContent == null) {
            assertTrue("", targets.isEmpty());
            return;
        }
```

With the addition of `LSPTargetElementEvaluator`, there will always be a _prospective_ target element at the caret when a word is found there, so `#documentationTargets()` is returning a list with a single (fake) documentation target element.

I can fix these three tests by changing `editorContentText` from:

```typescript
const s = "";
s.c<caret>harAt(0);
```

to:

```typescript
const s = "";<caret>
s.charAt(0);
```

so that the caret is not inside of a word. Does that seem like a reasonable change to these tests based on the new assumption that there will always be a target element when the caret is inside of a word?

SCWells72:
Ah, that other failing test is in the context of #639. I'll work on it in the context of that branch/PR.

angelozerr:
> @angelozerr the issue with the three failing tets in `TypeScriptHoverTest` is with the following from `LSPHoverFixtureTestCase#assertHover()`, lines 72-76:
> 
> ```java
>         List<? extends @NotNull DocumentationTarget> targets = IdeDocumentationTargetProvider.getInstance(myFixture.getProject()).documentationTargets(editor, file, offset);
>         if (expectedHoverContent == null) {
>             assertTrue("", targets.isEmpty());
>             return;
>         }
> ```
> 
> With the addition of `LSPTargetElementEvaluator`, there will always be a _prospective_ target element at the caret when a word is found there, so `#documentationTargets()` is returning a list with a single (fake) documentation target element.
> 
> I can fix these three tests by changing `editorContentText` from:
> 
> ```typescript
> const s = "";
> s.c<caret>harAt(0);
> ```
> 
> to:
> 
> ```typescript
> const s = "";<caret>
> s.charAt(0);
> ```
> 
> so that the caret is not inside of a word. Does that seem like a reasonable change to these tests based on the new assumption that there will always be a target element when the caret is inside of a word?

Tests uses a mock languzge server which uses the real lsp response. This test is when yiu hover charAt method, the hover show the charAt documentation.

If yiu change caret position it doesnt reflect the real usecases.

I dont understand why your PR impacts the IJ documentation target.

It would be very nice to keep the test line today.

angelozerr:
> > @angelozerr the issue with the three failing tets in `TypeScriptHoverTest` is with the following from `LSPHoverFixtureTestCase#assertHover()`, lines 72-76:
> > 
> > ```java
> >         List<? extends @NotNull DocumentationTarget> targets = IdeDocumentationTargetProvider.getInstance(myFixture.getProject()).documentationTargets(editor, file, offset);
> >         if (expectedHoverContent == null) {
> >             assertTrue("", targets.isEmpty());
> >             return;
> >         }
> > ```
> > 
> > With the addition of `LSPTargetElementEvaluator`, there will always be a _prospective_ target element at the caret when a word is found there, so `#documentationTargets()` is returning a list with a single (fake) documentation target element.
> > 
> > I can fix these three tests by changing `editorContentText` from:
> > 
> > ```typescript
> > const s = "";
> > s.c<caret>harAt(0);
> > ```
> > 
> > to:
> > 
> > ```typescript
> > const s = "";<caret>
> > s.charAt(0);
> > ```
> > 
> > so that the caret is not inside of a word. Does that seem like a reasonable change to these tests based on the new assumption that there will always be a target element when the caret is inside of a word?
> 
> Tests uses a mock languzge server which uses the real lsp response. This test is when yiu hover charAt method, the hover show the charAt documentation.
> 
> If yiu change caret position it doesnt reflect the real usecases.
> 
> I dont understand why your PR impacts the IJ documentation target.
> 
> It would be very nice to keep the test line today.

Oh U understand it is the `LSPTargetElementEvaluator` which causes the problem, but in real usecase we qhould have the same problem, no?

SCWells72:
> Oh U understand it is the `LSPTargetElementEvaluator` which causes the problem, but in real usecase we qhould have the same problem, no?

I'm not sure I follow. Those tests were failing because they were placing the caret on what looks like an identifier and then not expecting a documentation target element, but `LSPTargetElementEvaluator` will now return a fake target element anytime the caret is within what appears to be a word. So 1) those three tests are now invalid; or 2) they shouldn't be placing the caret on what might be an identifier (what my change does); or 3) they shouldn't be expecting a null identifier but rather confirming the behavior of the non-null identifier to be correct; or 4) the result of `LSPTargetElementEvaluator` shouldn't be propagated out through `IdeDocumentationTargetProvider#documentationTargets()` based on some kind of filter criteria.

My guess is that it's likely the third or fourth option. Let me know if that doesn't make sense.

angelozerr:
If you create a ts file like test and you hover charAt, do you see documentation?

angelozerr:
I dont understand why in test case  it fails although in real case it is working.

The fix that you did by moving caret outside charAt doesnt emulate the real usecase.

SCWells72:
> I dont understand why in test case it fails although in real case it is working.
> 
> The fix that you did by moving caret outside charAt doesnt emulate the real usecase.

That's correct. The issue with, for example, `testNoHoverWithEmptyContents` is that it's asserting that when you simulate a hover over:

```typescript
const s = "";
s.c<caret>harAt(0);
```

the language server could return no target element and no hover contents. However, the former assumption is flawed, so the following assertion fails because `expectedHoverContent` is passed as `null` but `targets` has one entry:

```java
        if (expectedHoverContent == null) {
            assertTrue("", targets.isEmpty());
            return;
        }
```

So perhaps those last three tests in `TypeScriptHoverTest` are no longer valid now that an actual (fake) PSI element is returned when the caret is on a word? It looks like the preceding test methods in that same test class already cover the scenarios where content is returned for the same code/caret placement.

src/main/java/com/redhat/devtools/lsp4ij/features/workspaceSymbol/LSPWorkspaceTypeDeclarationProvider.java
```diff
@@ -0,0 +1,118 @@
...
+        } catch (CancellationException ex) {
+            // cancel the LSP requests textDocument/typeDefinition
+            typeDefinitionSupport.cancel();
+        } catch (ExecutionException e) {
+            LOGGER.error("Error while consuming LSP 'textDocument/typeDefinition' request", e);
+        }
+
+        if (isDoneNormally(typeDefinitionsFuture)) {
+            List<Location> typeDefinitions = typeDefinitionsFuture.getNow(null);
+            if (!ContainerUtil.isEmpty(typeDefinitions)) {
```
angelozerr, line 106: I think we should show an hint when there are no type definition (like you did for implementation)

src/main/java/com/redhat/devtools/lsp4ij/features/workspaceSymbol/LSPWorkspaceImplementationsSearch.java
```diff
@@ -0,0 +1,142 @@
...
+            implementationSupport.cancel();
+        } catch (ExecutionException e) {
+            LOGGER.error("Error while consuming LSP 'textDocument/implementation' request", e);
+        }
+
+        if (isDoneNormally(implementationsFuture)) {
+            List<Location> implementations = implementationsFuture.getNow(null);
+            if (ContainerUtil.isEmpty(implementations)) {
+                // No implementations found
+                Editor editor = PsiEditorUtil.findEditor(file);
```
angelozerr, line 117: To be consistent with LSP4IJ base code, write:

`Editor editor = LSPIJUtils.editorForElement(file);`

src/main/java/com/redhat/devtools/lsp4ij/features/workspaceSymbol/LSPWorkspaceImplementationsSearch.java
```diff
@@ -0,0 +1,142 @@
...
+        } catch (CancellationException ex) {
+            // cancel the LSP requests textDocument/implementation
+            implementationSupport.cancel();
+        } catch (ExecutionException e) {
+            LOGGER.error("Error while consuming LSP 'textDocument/implementation' request", e);
+        }
+
+        if (isDoneNormally(implementationsFuture)) {
+            List<Location> implementations = implementationsFuture.getNow(null);
+            if (ContainerUtil.isEmpty(implementations)) {
```
angelozerr, line 115: This code block should be perhaps move to an class utility to use it with typeDefinition ?

src/main/java/com/redhat/devtools/lsp4ij/features/workspaceSymbol/LSPWorkspaceImplementationsSearch.java
```diff
@@ -0,0 +1,142 @@
...
+            return true;
+        }
+
+        // Check if the file can support the feature
+        if (!LanguageServiceAccessor.getInstance(project)
+                .hasAny(file.getVirtualFile(), ls -> ls.getClientFeatures().getImplementationFeature().isImplementationSupported(file))) {
+            return true;
+        }
+
+        Document document = PsiDocumentManager.getInstance(project).getDocument(file);
```
angelozerr, line 88: Move this test in L76

angelozerr, line 88: Replace with 

`Document document= LSPIJUtils.getDocument(file);`

src/main/java/com/redhat/devtools/lsp4ij/features/workspaceSymbol/LSPWorkspaceTypeDeclarationProvider.java
```diff
@@ -0,0 +1,118 @@
...
+        if (ProjectIndexingManager.canExecuteLSPFeature(file) != ExecuteLSPFeatureStatus.NOW) {
+            return PsiElement.EMPTY_ARRAY;
+        }
+
+        if (!LanguageServiceAccessor.getInstance(project)
+                .hasAny(file.getVirtualFile(), ls -> ls.getClientFeatures().getTypeDefinitionFeature().isTypeDefinitionSupported(file))) {
+            return PsiElement.EMPTY_ARRAY;
+        }
+
+        Document document = PsiDocumentManager.getInstance(project).getDocument(file);
```
angelozerr, line 84: `Document document= LSPIJUtils.getDocument(file);`

angelozerr, line 84: Move this test to L74

src/main/java/com/redhat/devtools/lsp4ij/LSPIJUtils.java
```diff
@@ -413,22 +413,40 @@ private static Language doGetFileLanguage(@NotNull VirtualFile file, @NotNull Pr
...
         return ReadAction.compute(() -> FileDocumentManager.getInstance().getDocument(file));
     }
 
+    /**
+     * Returns the document corresponding to the PSI element
+     *
+     * @param element the PSI element
+     * @return the document corresponding to the PSI element, or {@code null} if no document could be found
+     */
+    public static @Nullable Document getDocument(@NotNull PsiElement element) {
```
SCWells72, line 445: Added this composite signature for convenience of translation directly from a PSI element to the corresponding document.

src/main/java/com/redhat/devtools/lsp4ij/ui/LSP4IJUiUtils.java
```diff
@@ -0,0 +1,116 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.swing.*;
+import java.util.function.Function;
+
+/**
+ * Common UI utilities.
+ */
+public final class LSP4IJUiUtils {
```
SCWells72, line 33: Added a new pure utility class for UI-related utility methods. Right now it's just the `show*Hint()` family of methods, of which only one, `showErrorHint()`, is used directly. I can remove the unused ones if you'd like, but they're very simple and could well be useful for showing other types of editor hints. Let me know your thoughts.

src/main/java/com/redhat/devtools/lsp4ij/features/workspaceSymbol/LSPWorkspaceImplementationsSearch.java
```diff
@@ -0,0 +1,122 @@
...
+        if (!element.isValid()) {
+            return true;
+        }
+
+        PsiFile file = element.getContainingFile();
+        if ((file == null) || !file.isValid()) {
+            return true;
+        }
+
+        Document document = LSPIJUtils.getDocument(file);
```
SCWells72, line 69: In both implementations, I've moved the check for a non-null document above any other potentially more expensive checks.

src/main/resources/messages/LanguageServerBundle.properties
```diff
@@ -130,6 +130,8 @@ no.usage.type.typeDefinitions=No Type Definition found.
 no.usage.type.references=No Reference found.
 no.usage.type.implementations=No Implementation found.
 
+# Type declaration provider
+goto.typeDeclaration.notFound=No type declarations found
```
SCWells72, line 134: Had to add this here because there was not already a message for this in the IDE's standard message bundle as it doesn't show any message if the **Navigate | Type Declaration** action yields no results. I agree that showing such a message provides a better experience, but that's why there's a custom message here.

angelozerr:
It work slike a charm and code is veryclean. Good job @SCWells72 !

PR #639, opened 2024-11-27
feat: Short-circuit completions when context-specific completions have been added from LSP (issue 638)

SCWells72:
Issue 638 - The LSP completion contributor now short-circuits further completions if context-specific completions were added from LSP and completion was triggered by auto-completion or a single press of the explicit completion trigger.


src/main/java/com/redhat/devtools/lsp4ij/features/completion/LSPCompletionContributor.java
```diff
@@ -139,6 +139,12 @@ private void addCompletionItems(@NotNull CompletionParameters parameters,
                completionFeature.addLookupItem(completionPrefix, result, lookupItem, size- i, item);
             }
         }
+
+        // If completions were added from LSP and this is auto-completion or the explicit completion trigger was typed
+        // only once, don't show completions from other contributors, specifically the word completion contributor
+        if ((size > 0) && (parameters.getInvocationCount() < 2)) {
```
angelozerr, line 145: Takes a sample. You are using pycharm and you have a language server which manages python.

In this case completion will not show completion managed with pycharm?

SCWells72: As long as the language server returns at least one completion for the current location, only the language server-provided completion(s) will be offered. If there are no completions returned by the language server, any other registered completion contributors are consulted including the word completion contributor.

Let me know if that doesn't make sense, but it's how pretty much all other grammar-aware completion contributors work to provide a proper signal-to-noise ratio.

angelozerr: > I misunderstood your example so I'll try again. 

Sorry with my bad explanations. Takes a real sample that we have in IJ Quarkus. Our Quarkus language server contribute to Java language to provide completion snippet (template). When comletion is opened it should show:

 * standard Java completion from IJ
 * the snippet comming from Quarkus language server

If I understand your PR, if snippet from  Quarkus language server is shown, the standard Java completion from IJ will not appear?

> Yes, you're correct that the `LSPCompletionContributor` would show completions from the Python language server and then stop collecting completions before the IDE's native Python plugin. Similarly, here's IntelliJ IDEA Ultimate Edition -- which of course has native TypeScript support -- with the TypeScript language server configured for LSP4IJ:
> 
> image
> 
> All of the displayed completions are from LSP4J.
> 
> Here's the same completion request with LSP4IJ's TypeScript language server disabled showing completions from the native TypeScript plugin:
> 
> image
> 
> I would consider having a language server active for a natively-supported language to be a misconfiguration, though. Is there a compelling use case for that?

SCWells72: Ah, yes. That's a compelling use case. It sounds like perhaps disabling other completions should be a per-language client-side setting alongside grammar case-sensitivity. Sounds like the best use of my time immediately is adding those client-side configuration options to unlock some of these other capabilities in a proper manner.

angelozerr:
@InSyncWithFoo could you please try this PR (see https://github.com/redhat-developer/lsp4ij?tab=readme-ov-file#testing-the-ci-builds) to check that startdard IDE action (implementation and declartion) is working for you and doesn't override the implementation of PyCharm for instance.

SCWells72:
@angelozerr, the last commit should address both the unit test failure and your concerns about how this change would work with Quarkus. No client-side config option should be needed for the proper behavior.

src/main/java/com/redhat/devtools/lsp4ij/features/completion/LSPCompletionContributor.java
SCWells72: I believe this concern should be resolved by the newest commits. Let me know if not.

angelozerr:
Oopps sory @InSyncWithFoo with my mistake. I wanted to say please try https://github.com/redhat-developer/lsp4ij/pull/637

src/main/java/com/redhat/devtools/lsp4ij/features/completion/LSPCompletionContributor.java
angelozerr: Is your new commit disable native language completion. For instance ifbyiu have ultimate which provides typescript completion and lsp4ij which provides completion with ts languagevserver. What happens ?

If native typescript completion it can be annoying.

I think to avoid doing this choice display native only, display lsp4ij only, display both we could provide a new completion feature method which returns a enum like override_native, ignore_lsp, both.

By default it should show both.

With user definzd ls we could have a setting like casSensitive completion.

What do you think about this idea?

SCWells72: No, the most recent change is all about disabling completions from `WordCompletionContributor` if possible by setting `BaseCompletionService.FORBID_WORD_COMPLETION=true` temporarily. The issue here is that `TextMateCompletionContributor` calls the following as its last step:

```java
WordCompletionContributor.addWordCompletionVariants(resultSetWithPrefix, parameters, Collections.emptySet(), true);
```

which muddies the offered completions when LSP4IJ is augmenting a TextMate-managed file type.

Here's an example of TypeScript completions in Ultimate Edition with the TypeScript language server enabled for LSP4IJ:

image

As you can see, completions for `Console.log()` have been added from both sources, but you won't see `log` as a word completion here because it's not a TextMate file or even when it is a TextMate file unless you press the completion trigger (**Ctrl+Space**) more than one time.

In the situation where the included change is unable to set `BaseCompletionService.FORBID_WORD_COMPLETION=true` temporarily, it will instead run the remaining completion contributors and filter out redundant completions based on the lookup string, but in actuality, that should never happen because the `CompletionProcess` is always a `UserDataHolder`. Again, that's just me being defensive.

Let me know if that doesn't address your concerns.

angelozerr: If your PR disable only word completion when LSP returns somethinf that is excellent!

src/main/java/com/redhat/devtools/lsp4ij/features/completion/LSPCompletionContributor.java
```diff
@@ -136,7 +135,50 @@ private void addCompletionItems(@NotNull CompletionParameters parameters,
...
+            }
+        }
+
+        // If completions were added from LSP and this is auto-completion or the explicit completion trigger was typed
+        // only once, add completions from other all other contributors except for the word completion contributor.
+        // Note that the word completion contributor only really kicks in when LSP completions are added in the context
+        // of a TextMate file (see TextMateCompletionContributor), but that's going to be very common for LSP usage
+        // since it's often adding (pseudo-)language support when not already present in the host JetBrains IDE.
+        if ((size > 0) && (parameters.getInvocationCount() < 2)) {
+            // Stop standard processing
```
angelozerr, line 149: What means "Stop standard processing" ?

SCWells72:
@angelozerr, rebasing this PR against the latest on main resulted in conflicts that had to be merged. After doing so, both rebase and merge resulted in duplicate commits. I _think_ you're just going to do a squash commit when you merge into main which would take care of that, but if you'd like me to create a new PR with these changes from a fresh version of main and cancel this one, just let me know. That's certainly easy enough to do. Apologies for the nuisance.

PR #647, opened 2024-12-01
fix: Show progress messages in detail

seachicken:
Hello. When I use `$/progress`, the detailed message is displayed on the front.
スクリーンショット 2024-12-01 9 23 13

The detailed message is displayed as `text2` in the [IntelliJ ProgressIndicator API](https://github.com/JetBrains/intellij-community/blob/56c38a555ee36467a9ac171f074d4d15d89a83a4/platform/core-api/src/com/intellij/openapi/progress/ProgressIndicator.java#L149).
The LSP title and message seem to correspond to the following:
| LSP | IntelliJ API |
|--------|--------|
| title | ProgressIndicator.text() |
| message | ProgressIndicator.text2() | 

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workDoneProgress

### DIFF
--- a/src/main/java/com/redhat/devtools/lsp4ij/features/progress/LSPProgressManager.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/features/progress/LSPProgressManager.java
@@ -139,7 +139,7 @@ public class LSPProgressManager implements Disposable {
                                          @Nullable Integer percentage,
                                          @NotNull ProgressIndicator progressIndicator) {
         if (message != null && !message.isBlank()) {
-            progressIndicator.setText(message);
+            progressIndicator.setText2(message);
         }
         if (percentage != null) {
             progressIndicator.setFraction(percentage.doubleValue() / 100);


angelozerr:
I have never understood the difference betweeb setText and setText2.

@SCWells72 what do you think about this change?

SCWells72:
@angelozerr, the behavior differs based on whether the progress indicator is displayed in the status bar (as in the attached screenshot) or in a modal dialog. When displayed in a modal dialog, `setText()` is for the text displayed _above_ the progress bar, and `setText2()` is for the text displayed _below_ the progress bar. When displayed in the status bar, my experience has generally been that `setText2()` doesn't really do much, so I've always used `setText()` for progress displayed that way.

Typically I would think that `title` would be what you set as the progress indicator title when the process is started, and `message` would correspond to `setText()`, but I can imagine there might be a bit of a catch-22 here with needing to start the progress indicator before calling the language server, so I think mapping `title` to `setText()` and `message` to `setText2()` seems reasonable.

angelozerr:
Thanks @SCWells72 for your feedback.

Most of code that I have seen in IJ community github uses setText. 

I think LSP4IJ should continue to use setText

@seachicken if you need really this call of setText2 we could provide an LSP Api?

seachicken:
IntelliJ's "Scanning files to index" also seems to display detailed information in `text2`, and looking at the SDK documentation, it seems like an appropriate solution.

image

https://plugins.jetbrains.com/docs/intellij/progress-bar.html#process-details

However, I will close the PR if many LSP users expect `text`. Thanks.

angelozerr:
@seachicken I have not strong opinion about that. If you think it is better I can merge it.

seachicken:
@angelozerr In my opinion, the index target file name being visible to IDE users seems noisy and inconsistent with the standard indicators.
Customizing the LSP API is also a bit difficult, so changing it to text2 is not a bad idea.

angelozerr:
Thanks @seachicken !

PR #649, opened 2024-12-01
fix: Allow requesting scalar values in workspace/configuration

### LINKED ISSUE #648: workspace/configuration does not work when scalar value is requested
When using the language server for the [slang shading language](https://github.com/shader-slang/slang) It will send a workspace/configuration request. The sections requested there point to "scalar" values. They are not json objects but strings and bools directly.

In VSCode this will work correctly. But with this plugin the return value is always null (tested with a7d451d). 

Messages in VSCode:
Image

Messages with this plugin:
Image

Configuration of the lsp in this plugin:
```{
    "slang": {
        "predefinedMacros": [],
        "additionalSearchPaths": [],
        "searchInAllWorkspaceDirectories": true,
        "enableCommitCharactersInAutoCompletion": "membersOnly",
        "format": {
            "clangFormatLocation": "",
            "clangFormatStyle": "file",
            "clangFormatFallbackStyle": "{BasedOnStyle: Microsoft, BreakBeforeBraces: Allman, ColumnLimit: 0}",
            "allowLineBreakChangesInOnTypeFormatting": false,
            "allowLineBreakChangesInRangeFormatting": false
        },
        "inlayHints": {
            "deducedTypes": true,
            "parameterNames": true
        }
    },
    "slangLanguageServer": {
        "trace": {
            "server": "verbose"
        }
    }
}

### PR DESCRIPTION
**Pull request by JanThomas118**
This is my attempt to fix #648 

My knowledge about LSP and its different language servers is very limited. Atleast for the faulty request shown in the issue and the configuration listed there this will now return the correct values.

### DIFF
--- a/src/main/java/com/redhat/devtools/lsp4ij/client/SettingsHelper.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/client/SettingsHelper.java
@@ -1,18 +1,42 @@
+/*******************************************************************************
+ * Copyright (c) 2023 Red Hat Inc. and others.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+ * which is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ *
+ * Contributors:
+ *     Red Hat Inc. - initial API and implementation
+ *******************************************************************************/
 package com.redhat.devtools.lsp4ij.client;
 
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
+/**
+ * Helpers for extracting nested settings from json
+ */
 public class SettingsHelper {
 
     private SettingsHelper() {
     }
 
-    public static JsonElement findSettings(String[] sections, JsonObject jsonObject) {
-        JsonElement current = jsonObject;
+    /**
+     * Extract nested settings from json.
+     * @param sections path to the json element to retrieve
+     * @param parent Json to look for settings in
+     * @return the settings retrieved in the specified section and null if the section was not found.
+     */
+    public static @Nullable JsonElement findSettings(@NotNull String[] sections, @Nullable JsonObject parent) {
+        JsonElement current = parent;
         for (String section : sections) {
-            if (current instanceof JsonObject object) {
-                current = object.get(section);
+            if (current instanceof JsonObject currentObject) {
+                current = currentObject.get(section);
                 if (current == null) {
                     return null;
                 }
--- a/src/main/java/com/redhat/devtools/lsp4ij/client/SettingsHelper.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/client/SettingsHelper.java
@@ -1,0 +1,23 @@
+package com.redhat.devtools.lsp4ij.client;
+
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+
+public class SettingsHelper {
+
+    private SettingsHelper() {
+    }
+
+    public static JsonElement findSettings(String[] sections, JsonObject jsonObject) {
+        JsonElement current = jsonObject;
+        for (String section : sections) {
+            if (current instanceof JsonObject object) {
+                current = object.get(section);
+                if (current == null) {
+                    return null;
+                }
+            }
+        }
+        return current;
+    }
+}
--- a/src/test/java/com/redhat/devtools/lsp4ij/client/SettingsHelperTest.java
+++ b/src/test/java/com/redhat/devtools/lsp4ij/client/SettingsHelperTest.java
@@ -1,0 +1,73 @@
+package com.redhat.devtools.lsp4ij.client;
+
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
+import com.intellij.testFramework.fixtures.BasePlatformTestCase;
+
+import java.io.StringReader;
+
+public class SettingsHelperTest extends BasePlatformTestCase {
+    // language=json
+    private final String testJson = """
+            {
+                "mylsp": {
+                    "myscalarsetting": "value",
+                    "myobjectsettings": {
+                        "subsettingA": 1,
+                        "subsettingB": 2
+                    }
+                }
+            }
+            """;
+
+    public void testGetSettingsPrimitiveValue() {
+        JsonObject jsonObject = JsonParser.parseReader(new StringReader(testJson)).getAsJsonObject();
+        String[] requestedPath = new String[]{"mylsp", "myscalarsetting"};
+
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by fbricon** [comment]
We'll likely need a new method, as I believe this one is API, changing signature would break 3rd party adopters. @angelozerr ?

**Comment by angelozerr** [comment]
@fbricon I think nobody overrides this method and use the default code.

But I think we should have tests now with this findSettings method.

@JanThomas118 do you think you could write them? If yes I suggest that you create a SettingsHelper class which have a static findSentings method.

You could after that use it in the current code and in the test.

**Comment by JanThomas118** [comment]
Good point. My change is changing public API i actually did not think about that. But the parameter could just continue being JsonObject like before. Then the signature would be identical to before.

Regarding the tests: I can try to do that. This is the first time i am writing java code in years so. I will try my best

File: src/main/java/com/redhat/devtools/lsp4ij/client/SettingsHelper.java [hunk 1]
```diff
@@ -0,0 +1,23 @@
+package com.redhat.devtools.lsp4ij.client;
+
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+
+public class SettingsHelper {
```
angelozerr, line 6: Add javdoc

src/main/java/com/redhat/devtools/lsp4ij/client/SettingsHelper.java
```diff
@@ -0,0 +1,23 @@
+package com.redhat.devtools.lsp4ij.client;
```
angelozerr, line 1: Add header

src/main/java/com/redhat/devtools/lsp4ij/client/SettingsHelper.java
```diff
@@ -0,0 +1,23 @@
...
+
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+
+public class SettingsHelper {
+
+    private SettingsHelper() {
+    }
+
+    public static JsonElement findSettings(String[] sections, JsonObject jsonObject) {
```
angelozerr, line 11: Add javdoc

angelozerr, line 11: Add @NotNull for sections and @Nullable I think for jsonObject

angelozerr, line 11: Rename jsonObject with parent

src/main/java/com/redhat/devtools/lsp4ij/client/SettingsHelper.java
```diff
@@ -0,0 +1,23 @@
...
+        JsonElement current = jsonObject;
+        for (String section : sections) {
+            if (current instanceof JsonObject object) {
```
angelozerr, line 14: Rename object with child

src/test/java/com/redhat/devtools/lsp4ij/client/SettingsHelperTest.java
```diff
@@ -0,0 +1,73 @@
+package com.redhat.devtools.lsp4ij.client;
+
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
+import com.intellij.testFramework.fixtures.BasePlatformTestCase;
+
+import java.io.StringReader;
+
+public class SettingsHelperTest extends BasePlatformTestCase {
```
angelozerr, line 10: Add javadoc

src/test/java/com/redhat/devtools/lsp4ij/client/SettingsHelperTest.java
```diff
@@ -0,0 +1,73 @@
+package com.redhat.devtools.lsp4ij.client;
```
angelozerr, line 1: Add header

src/test/java/com/redhat/devtools/lsp4ij/client/SettingsHelperTest.java
```diff
@@ -0,0 +1,73 @@
...
+                    "myobjectsettings": {
+                        "subsettingA": 1,
+                        "subsettingB": 2
+                    }
+                }
+            }
+            """;
+
+    public void testGetSettingsPrimitiveValue() {
+        JsonObject jsonObject = JsonParser.parseReader(new StringReader(testJson)).getAsJsonObject();
```
angelozerr, line 25: To avoid duplicating code I suggest that you move this load and the assert check in one static methad assertFindSettings(String json, String[] sections, String expected)

src/test/java/com/redhat/devtools/lsp4ij/client/SettingsHelperTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+    public void testGetSettingsIdentityOnEmptySections() {
+        JsonObject jsonObject = JsonParser.parseReader(new StringReader(testJson)).getAsJsonObject();
+        String[] requestedPath = new String[0];
+
+        JsonElement result = SettingsHelper.findSettings(requestedPath, jsonObject);
+
+        assertEquals(jsonObject, result);
+    }
+}
```
angelozerr, line 73: Please add findSettings tests with one section and three section (today your test is with zero or two sections)

fbricon:
> @fbricon I think nobody overrides this method and use the default code.

We have no way to be sure. We should just be conservative about it

angelozerr:
Ok lets keep the original signature.

src/main/java/com/redhat/devtools/lsp4ij/client/SettingsHelper.java
JanThomas118: I agree that "object" was a terrible name. Child would not be a correct name though because the actual child is retrieved with get() one line below. This variable is just "current" casted to JsonObject. I therefore called it currentObject

src/test/java/com/redhat/devtools/lsp4ij/client/SettingsHelperTest.java
angelozerr: @JanThomas118 is there any chance that you add those tests?

angelozerr: It would be nice also to add test with zero and for sections to cover all usecases.

@JanThomas118 if you have no time to do that, please tell me. We could merge your PR and I will add tests.

src/test/java/com/redhat/devtools/lsp4ij/client/SettingsHelperTest.java
JanThomas118: As far as i understood your suggestion this has now been done. Correct me of i am wrong

src/test/java/com/redhat/devtools/lsp4ij/client/SettingsHelperTest.java
JanThomas118: Zero sections is handled by testGetSettingsIdentityOnEmptySections. Currently there are tests for selections with length 0 1 2 and 3. If you want one with four i can add that but what benefit do you hope this brings when 2 and 3 is already covered?

angelozerr: I suggested to add  four to cover the case where sections.length > flattered JSON properties lengh.

But it is OK.

PR #655, opened 2024-12-01
fix: Fixed an issue with the JSON text editors caused by them being in scroll panes

SCWells72:
Those components have their own scrolling, and the nested scrolling mechanisms were causing issues with keyboard navigation, e.g., page up/down, etc., and other features such a Find (Ctrl+F). Without the surrounding scroll panes, it was also important that they be added so that they fill their available space vertically.


angelozerr:
Please rename title of the issue with feat:

PR #659, opened 2024-12-02
feat: Add client-side language server configuration with per-language case sensitivity as the first available option (issue 626)

SCWells72:
This should be a pretty complete implementation of client-side configuration via JSON based on prior discussions with @angelozerr. As expected, the first client-side configuration option is a language-level option designating the case-sensitivity of the associated grammar. That option defaults to `false` for backward-compatibility, and I've added default client-side settings to the all existing language server templates with the appopriate value for `caseSensitive`. Unsurprisingly, only one template-supported language, HTML, is case-insensitive.


src/main/java/com/redhat/devtools/lsp4ij/LanguageServersRegistry.java
```diff
@@ -162,7 +162,8 @@ private void loadServersAndMappingFromSettings() {
                                 launch.getUserEnvironmentVariables(),
                                 launch.isIncludeSystemEnvironmentVariables(),
                                 launch.getConfigurationContent(),
-                                launch.getInitializationOptionsContent()),
+                                launch.getInitializationOptionsContent(),
+                                launch.getClientConfigurationContent()),
```
SCWells72, line 166: There's obviously going to be a proliferation of these types of changes where I've added `clientConfiguration` as another option of `UserDefinedLanguageServerDefinition` and its peer types/methods/etc.

src/main/java/com/redhat/devtools/lsp4ij/LanguageServersRegistry.java
```diff
@@ -483,13 +486,15 @@ private void removeAssociationsFor(LanguageServerDefinition definition) {
         boolean mappingsChanged = !Objects.deepEquals(settings.getMappings(), request.mappings());
         boolean configurationContentChanged = !Objects.equals(settings.getConfigurationContent(), request.configurationContent());
         boolean initializationOptionsContentChanged = !Objects.equals(settings.getInitializationOptionsContent(), request.initializationOptionsContent());
+        // Not checking whether client config changed because that shouldn't result in a LanguageServerChangedEvent
```
SCWells72, line 489: Specifically calling out that we don't need to include client config changes when computing whether or not to send an event here because the actual language server doesn't care at all about client-side config changes.

src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPCompletionFeature.java
```diff
@@ -220,27 +220,38 @@ public boolean isItemTextBold(@NotNull CompletionItem item) {
...
         // Compute the prefix
         var textEditRange = ((LSPCompletionProposal) lookupItem).getTextEditRange();
         String prefix = textEditRange != null ? completionPrefix.getPrefixFor(textEditRange, item) : null;
         if (prefix != null) {
-            // Add the IJ completion item (lookup item) by using the computed prefix
-            result.withPrefixMatcher(prefix)
-                    .caseInsensitive()
-                    .addElement(prioritizedLookupItem);
+            // Add the IJ completion item (lookup item) by using the computed prefix respecting the language's case-sensitivity
+            if (caseSensitive) {
```
SCWells72, line 238: I could condense these two conditionals slightly be extracting `result.withPrefixMatcher()` into a variable and setting other state conditionally, but I didn't find that quite as readable.

src/main/java/com/redhat/devtools/lsp4ij/features/completion/LSPCompletionContributor.java
```diff
@@ -136,11 +143,23 @@ private void addCompletionItems(@NotNull CompletionParameters parameters,
             // Create lookup item
             LookupElement lookupItem = completionFeature.createLookupElement(item, context);
             if (lookupItem != null) {
-               completionFeature.addLookupItem(completionPrefix, result, lookupItem, size- i, item);
+                completionFeature.addLookupItem(completionPrefix, result, lookupItem, size - i, item, caseSensitive);
             }
         }
     }
 
+    private static boolean isCaseSensitive(@NotNull LanguageServerItem languageServer, @NotNull PsiFile file) {
```
SCWells72, line 151: This should be pretty straightforward, but basically try to infer language grammar case-sensitivity via the file => language ID => client-side settings.

src/main/java/com/redhat/devtools/lsp4ij/launching/UserDefinedLanguageServerSettings.java
```diff
@@ -119,6 +119,8 @@ public static class UserDefinedLanguageServerItemSettings {
 
         private String initializationOptionsContent;
 
+        private String clientConfigurationContent = "{}";
```
SCWells72, line 122: The default value of `{}` here is what allows smooth backward-compatibility with existing language servers.

src/main/java/com/redhat/devtools/lsp4ij/launching/templates/LanguageServerTemplate.java
```diff
@@ -40,6 +40,7 @@ public String getName() {
     public static final String TEMPLATE_FILE_NAME = "template.json";
     public static final String INITIALIZATION_OPTIONS_FILE_NAME = "initializationOptions.json";
     public static final String SETTINGS_FILE_NAME = "settings.json";
+    public static final String CLIENT_SETTINGS_FILE_NAME = "clientSettings.json";
```
SCWells72, line 43: Following the existing pattern of the server-side config file being named `settings.json` and naming the client-side config `clientSettings.json`.

src/main/java/com/redhat/devtools/lsp4ij/launching/templates/LanguageServerTemplateManager.java
```diff
@@ -131,6 +135,7 @@ public LanguageServerTemplate createLsTemplate(@NotNull VirtualFile templateFold
         String templateJson = null;
         String settingsJson = null;
         String initializationOptionsJson = null;
+        String clientSettingsJson = null;
```
SCWells72, line 138: The real changes in this file begin here where we add handling for `clientSettings.json`.

src/main/java/com/redhat/devtools/lsp4ij/server/definition/launching/ClientConfigurationLanguageSettings.java
```diff
@@ -0,0 +1,30 @@
...
+
+import com.google.gson.reflect.TypeToken;
+
+import java.lang.reflect.Type;
+import java.util.Map;
+
+/**
+ * Language-specific client-side settings for a user-defined language server configuration.
+ */
+public class ClientConfigurationLanguageSettings {
```
SCWells72, line 24: These are the actual per-language client-side settings. Again, there's only one right now, `caseSensitive`, but it I think we've discussed a few others in the context of other issues/PRs.

src/main/java/com/redhat/devtools/lsp4ij/server/definition/launching/UserDefinedLanguageServerDefinition.java
```diff
@@ -157,9 +171,19 @@ public Object getLanguageServerInitializationOptions() {
         return initializationOptions;
     }
 
+    public Map<String, ClientConfigurationLanguageSettings> getLanguageServerClientConfiguration() {
+        if ((clientConfiguration == null) && (clientConfigurationContent != null) && !clientConfigurationContent.isBlank()) {
+            try {
+                clientConfiguration = JSONUtils.getLsp4jGson().fromJson(clientConfigurationContent, ClientConfigurationLanguageSettings.TYPE_TOKEN);
```
SCWells72, line 177: Unlike the server-side settings and initialization options, we have a concrete schema here and can deserialize into a specific type, in this case, a map keyed by language ID with values that are per-language client-side settings.

src/main/java/com/redhat/devtools/lsp4ij/settings/jsonSchema/AbstractLSPJsonSchemaFileProvider.java
```diff
@@ -0,0 +1,83 @@
...
+import com.jetbrains.jsonSchema.impl.JsonSchemaVersion;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.net.URL;
+
+/**
+ * Abstract base class for JSON schema file providers that are based on JSON schema files bundled in the plugin distribution.
+ */
+abstract class AbstractLSPJsonSchemaFileProvider implements JsonSchemaFileProvider {
```
SCWells72, line 31: Now we can get back to JSON schema-based validation given that we have a well-defined schema for our client-side configuration. This is the abstract base class for the one I've added and any others that might be needed in the future. It already encapsulates loading the JSON schema definition from the plugin distribution.

src/main/java/com/redhat/devtools/lsp4ij/settings/jsonSchema/LSPClientConfigurationJsonSchemaFileProvider.java
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * Contributors:
+ *     Red Hat Inc. - initial API and implementation
+ *******************************************************************************/
+package com.redhat.devtools.lsp4ij.settings.jsonSchema;
+
+/**
+ * JSON schema file provider for language server client-side configuration.
+ */
+public class LSPClientConfigurationJsonSchemaFileProvider extends AbstractLSPJsonSchemaFileProvider {
```
SCWells72, line 19: And this is the first concrete JSON schema file provider for client-side config.

src/main/java/com/redhat/devtools/lsp4ij/settings/ui/JsonTextField.java
```diff
@@ -57,6 +65,25 @@ public JsonTextField(@NotNull Project project) {
         add(editorTextField, BorderLayout.CENTER);
     }
 
+    /**
+     * Sets the editor's filename so that the correct JSON schema will be used for code completion and validation.
+     *
+     * @param jsonFilename the JSON file name
+     */
+    public void setJsonFilename(@NotNull String jsonFilename) {
```
SCWells72, line 73: This allows us to make a `JsonTextField` use one of the registered JSON schema file providers based on the associated file name.

src/main/java/com/redhat/devtools/lsp4ij/settings/ui/LanguageServerPanel.java
```diff
@@ -250,15 +260,21 @@ private static JLabel createLabelForComponent(@NotNull @NlsContexts.Label String
     }
 
     private void createConfigurationField(FormBuilder builder) {
-        configurationWidget = new LanguageServerConfigurationWidget(project);
+        configurationWidget = new JsonTextField(project);
```
SCWells72, line 263: I removed the explcit subclasses of `JsonTextField` as they weren't adding any real value.

src/main/resources/META-INF/plugin.xml
```diff
@@ -158,6 +158,7 @@
     <depends>com.intellij.modules.platform</depends>
     <depends optional="true">org.jetbrains.plugins.textmate</depends>
     <depends optional="true">com.redhat.devtools.intellij.telemetry</depends>
+    <depends optional="true" config-file="plugin-json.xml">com.intellij.modules.json</depends>
```
SCWells72, line 161: Added the optional dependency for JSON schema validation using the proper mechanism of loading the associated plugin EP implementations via the separate file above. It would likely make sense to do the same with the other two optional dependencies if possible.

src/main/resources/jsonSchema/clientSettings.schema.json
```diff
@@ -0,0 +1,20 @@
+{
```
SCWells72, line 1: JSON schema for client-side configuration.

src/main/resources/messages/LanguageServerBundle.properties
```diff
@@ -33,10 +33,10 @@ language.server.mappings.fileNamePattern.no=No file name patterns mappings confi
 language.server.mappings.languageId.column=Language Id
 
 language.server.tab.configuration=Configuration
+language.server.tab.configuration.server=Server configuration
+language.server.tab.configuration.client=Client configuration
 language.server.configuration=Configuration:
-language.server.configuration.emptyText=Defines the configuration of the language server
```
SCWells72, line 37: These two messages were unused after the move to `JsonTextField` which doesn't support placeholder text.

src/main/resources/messages/LanguageServerBundle.properties
```diff
@@ -33,10 +33,10 @@ language.server.mappings.fileNamePattern.no=No file name patterns mappings confi
 language.server.mappings.languageId.column=Language Id
 
 language.server.tab.configuration=Configuration
+language.server.tab.configuration.server=Server configuration
```
SCWells72, line 36: These are the configuration sub-tab labels. Let me know if you'd prefer them to be changed.

src/main/resources/templates/typescript-language-server/clientSettings.json
```diff
@@ -0,0 +1,8 @@
+{
```
SCWells72, line 1: Client-side config defaults for the TypeScript language server template.

src/main/resources/templates/lemminx/clientSettings.json
```diff
@@ -0,0 +1,11 @@
+{
+  "xml": {
```
SCWells72, line 2: I don't know if the language IDs are correct here as they're not explicitly speicfied in this template's `template.json`. All three languages are case-sensitive, though.

src/main/resources/jsonSchema/clientSettings.schema.json
```diff
@@ -0,0 +1,22 @@
...
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "https://github.com/redhat-developer/lsp4ij/tree/main/src/main/resources/jsonSchema/clientSettings.schema.json",
+  "title": "LSP4IJ user-defined language server client-side settings JSON schema",
+  "type": "object",
+  "additionalProperties": false,
+  "patternProperties": {
+    "^[a-zA-Z0-9-_]+$": {
```
SCWells72, line 8: Let me know if you're aware of language IDs that wouldn't be matched by this.

src/main/java/com/redhat/devtools/lsp4ij/LanguageServersRegistry.java
angelozerr: For the moment I agree with you. Lets keep like this, but we could have a client settings which could impact the editor (like enable/disable lsp codelens which should refresh the editor and in this case we should send an event)

src/main/java/com/redhat/devtools/lsp4ij/features/completion/LSPCompletionContributor.java
```diff
@@ -125,6 +129,9 @@ private void addCompletionItems(@NotNull CompletionParameters parameters,
         items.sort(completionProposalComparator);
         int size = items.size();
 
+        // Determine whether or not the language is case-sensitive so that we can add completions accordingly
+        boolean caseSensitive = isCaseSensitive(languageServer, parameters.getOriginalFile());
```
angelozerr, line 133: To be consistent with IJ, I would rename with isCaseInsensitive

src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPCompletionFeature.java
```diff
@@ -220,27 +220,38 @@ public boolean isItemTextBold(@NotNull CompletionItem item) {
...
+     * @param caseSensitive
      */
     @ApiStatus.Internal
     public void addLookupItem(@NotNull CompletionPrefix completionPrefix,
                               @NotNull CompletionResultSet result,
                               @NotNull LookupElement lookupItem,
                               int priority,
-                              @NotNull CompletionItem item) {
+                              @NotNull CompletionItem item,
+                              boolean caseSensitive) {
```
angelozerr, line 231: Even if it is internal, please try to avoid updating the signature. The caseSensitive can be get inside the method.

src/main/java/com/redhat/devtools/lsp4ij/features/completion/LSPCompletionContributor.java
```diff
@@ -136,11 +143,23 @@ private void addCompletionItems(@NotNull CompletionParameters parameters,
...
-               completionFeature.addLookupItem(completionPrefix, result, lookupItem, size- i, item);
+                completionFeature.addLookupItem(completionPrefix, result, lookupItem, size - i, item, caseSensitive);
             }
         }
     }
 
+    private static boolean isCaseSensitive(@NotNull LanguageServerItem languageServer, @NotNull PsiFile file) {
+        LanguageServerDefinition serverDefinition = languageServer.getServerDefinition();
+        if (serverDefinition instanceof UserDefinedLanguageServerDefinition languageServerDefinition) {
+            Map<String, ClientConfigurationLanguageSettings> clientConfiguration = languageServerDefinition.getLanguageServerClientConfiguration();
```
angelozerr, line 154: This code must not be here but in the custom implementation for user defined of ClientFeatures which will register a custom CompletionFeatures

src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPCompletionFeature.java
SCWells72: I did look at doing exactly that but wasn't sure how to get the language server definition -- or at least the language ID -- from the existing parameters that are passed into `addLookupItem()`. I can get the PSI element from the lookup item. Is there a santioned way to get the language server definition/language ID given only a PSI element? I saw `LanguageServerRegistry#findLanguageServerDefinitionFor(Language, FileType, VirtualFile)` and `LanguageServerAccessor#getMatchedLanguageServerDefinitions()`, but those both seem internal right now as well.

src/main/resources/templates/clojure-lsp/clientSettings.json
```diff
@@ -0,0 +1,5 @@
+{
```
SCWells72, line 1: All of the languages integrated via these templates are for case-sensitive languages except for HTML (see below).

src/main/resources/templates/vscode-html-language-server/clientSettings.json
```diff
@@ -0,0 +1,5 @@
+{
+  "completions": {
+    "caseSensitive": false
```
SCWells72, line 3: This is the only one that's case-insensitive.

src/main/java/com/redhat/devtools/lsp4ij/server/definition/launching/UserDefinedCompletionFeature.java
```diff
@@ -0,0 +1,34 @@
...
+import com.redhat.devtools.lsp4ij.server.definition.LanguageServerDefinition;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Adds client-side completion configuration features.
+ */
+public class UserDefinedCompletionFeature extends LSPCompletionFeature {
+    public boolean isCaseSensitive(@NotNull PsiFile file) {
+        LanguageServerDefinition serverDefinition = getClientFeatures().getServerDefinition();
+        if (serverDefinition instanceof UserDefinedLanguageServerDefinition languageServerDefinition) {
```
angelozerr, line 27: This condition is always true

src/main/java/com/redhat/devtools/lsp4ij/server/definition/launching/UserDefinedCompletionFeature.java
```diff
@@ -0,0 +1,34 @@
...
+            ClientConfigurationSettings clientConfiguration = languageServerDefinition.getLanguageServerClientConfiguration();
+            return (clientConfiguration != null) && clientConfiguration.completions.caseSensitive;
+        }
+        // Default to case-insensitive if unspecified for backward-compatibility
+        return false;
```
angelozerr, line 32: I will return the super.isCaseSensitive

src/main/java/com/redhat/devtools/lsp4ij/server/definition/launching/UserDefinedCompletionFeature.java
SCWells72: I'm fine changing it to an immediate unchecked cast if you are, but `getServerDefinition()` in that expression returns `LanguageServerDefinition` and not `UserDefinedLanguageServerDefinition`, hence the `instanceof` check before safely casting. What's your preference?

angelozerr: Cast it directly since we are sure that it is a user defined ls

SCWells72: Done. Both of those feedback items are implemented. Let me know when you've completed your feedback and, once I've addressed all of it, I'll push another commit.

angelozerr:
I have tested the PR and it works very good, amazing job @SCWells72 !

There are few changes to do, but I merge the PR and I will create a little PR to fix few code.

Thanks @SCWells72 !

PR #662, opened 2024-12-04
fix: One more commit for issue 626 that removes an unnecessary instanceof/cast.

SCWells72:
Totally your call as to whether you want to merge this or discard it, but this was the last of the feedback I'd received.


angelozerr:
Thanks @SCWells72

PR #663, opened 2024-12-04
feat: Omit word completions (by default) when there are completions from the language server (issue 638)

SCWells72:
This is a reissue of PR #639 which ran into issues with rebase/merge against main. The same comments in that PR apply here. The only differences are additional comments added to describe behavior in relatively recent vs. older IDE version and to be more clear about what's happening when `result.stopHere()` is called.

### DIFF
--- a/src/main/java/com/redhat/devtools/lsp4ij/features/completion/LSPCompletionContributor.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/features/completion/LSPCompletionContributor.java
@@ -10,17 +10,17 @@
  ******************************************************************************/
 package com.redhat.devtools.lsp4ij.features.completion;
 
-import com.intellij.codeInsight.completion.CompletionContributor;
-import com.intellij.codeInsight.completion.CompletionParameters;
-import com.intellij.codeInsight.completion.CompletionResultSet;
+import com.intellij.codeInsight.completion.*;
 import com.intellij.codeInsight.lookup.*;
 import com.intellij.codeInsight.lookup.impl.LookupImpl;
 import com.intellij.openapi.editor.Document;
 import com.intellij.openapi.editor.Editor;
 import com.intellij.openapi.progress.ProcessCanceledException;
 import com.intellij.openapi.progress.ProgressManager;
+import com.intellij.openapi.util.UserDataHolder;
 import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.psi.PsiFile;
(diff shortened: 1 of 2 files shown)


src/main/java/com/redhat/devtools/lsp4ij/features/completion/LSPCompletionContributor.java
```diff
@@ -137,6 +136,51 @@ private void addCompletionItems(@NotNull CompletionParameters parameters,
...
+        // only once, add completions from other all other contributors except for the word completion contributor.
+        // Note that the word completion contributor only really kicks in when LSP completions are added in the context
+        // of a TextMate file (see TextMateCompletionContributor), but that's going to be very common for LSP usage
+        // since it's often adding (pseudo-)language support when not already present in the host JetBrains IDE.
+        if ((size > 0) && (parameters.getInvocationCount() < 2)) {
+            // Don't allow automatic execution of the remaining completion contributors; we'll execute them ourselves below
+            result.stopHere();
+
+            // Try to disable word completion by temporarily setting FORBID_WORD_COMPLETION=true on the completion process.
+            // Note that this will not work in some older IDE versions where WordCompletionContributor.addWordCompletionVariants()
```
SCWells72, line 153: Added a comment about why this may not be working in older supported IDE versions in case that gets logged as an issue in the future.

src/main/java/com/redhat/devtools/lsp4ij/features/completion/LSPCompletionContributor.java
```diff
@@ -137,6 +136,51 @@ private void addCompletionItems(@NotNull CompletionParameters parameters,
...
+            }
+        }
+
+        // If completions were added from LSP and this is auto-completion or the explicit completion trigger was typed
+        // only once, add completions from other all other contributors except for the word completion contributor.
+        // Note that the word completion contributor only really kicks in when LSP completions are added in the context
+        // of a TextMate file (see TextMateCompletionContributor), but that's going to be very common for LSP usage
+        // since it's often adding (pseudo-)language support when not already present in the host JetBrains IDE.
+        if ((size > 0) && (parameters.getInvocationCount() < 2)) {
+            // Don't allow automatic execution of the remaining completion contributors; we'll execute them ourselves below
```
SCWells72, line 149: Made the comment here more explicit about what's happening.

angelozerr:
Thanks again @SCWells72 for your great contribution!

PR #665, opened 2024-12-05
feat: LSP code folding improvements for braced blocks and code block provider implementation based on that same information

SCWells72:
This updates the `lang.foldingBuilder` implementation to examine the returned folding range and extend it to represent the entire braced block (where "braced" can be curly braces, square brackets, or parenthesis based on the language) just like it would for a natively-supported language, e.g., Java, Kotlin, etc. Folding ranges that are not bounded on both sides by paired braces are used as they were previously.

Additionally this logic has been made available to a `codeBlockProvider` implementation that provides brace matching and navigation up containing braced blocks using **Ctrl+[** and **Ctrl+]**.

If this isn't something you want in LSP4IJ -- or if you only want pieces of it -- that's fine. Just opening for feedback and to see where you all might want to go with it.


src/main/java/com/redhat/devtools/lsp4ij/features/foldingRange/LSPCodeBlockProvider.java
```diff
@@ -0,0 +1,86 @@
...
+    @Override
+    public @Nullable TextRange getCodeBlockRange(Editor editor, PsiFile file) {
+        List<FoldingRange> foldingRanges = LSPFoldingRangeBuilder.getFoldingRanges(file);
+        if (!ContainerUtil.isEmpty(foldingRanges)) {
+            Document document = editor.getDocument();
+            String documentText = document.getText();
+            int documentLength = documentText.length();
+
+            // Adjust the offset slightly based on before/after brace
+            int offset = editor.getCaretModel().getOffset();
```
SCWells72, line 42: These adjustments are pretty much the same as in the standard JetBrains language plugins where if the caret is immediately before an open brace or after a close brace, it is adjusted into the braced block for performing brace matching.

src/main/java/com/redhat/devtools/lsp4ij/features/foldingRange/LSPCodeBlockProvider.java
```diff
@@ -0,0 +1,86 @@
...
+            Character beforeCharacter = offset > 0 ? documentText.charAt(offset - 1) : null;
+            Character afterCharacter = offset < documentLength ? documentText.charAt(offset) : null;
+            if (isOpenBraceChar(afterCharacter)) {
+                offset++;
+            } else if (isCloseBraceChar(beforeCharacter)) {
+                offset--;
+            }
+
+            // See if we're anchored by a known brace character
+            Character openBraceChar = null;
```
SCWells72, line 52: With any required adjustments done, see if we're now searching immediately inside of a braced block and make note of the expected open/close braces.

src/main/java/com/redhat/devtools/lsp4ij/features/foldingRange/LSPCodeBlockProvider.java
```diff
@@ -0,0 +1,86 @@
...
+import static com.redhat.devtools.lsp4ij.features.foldingRange.LSPFoldingRangeBuilder.*;
+
+/**
+ * Code block provider that uses the folding range information from {@link LSPFoldingRangeBuilder}.
+ */
+public class LSPCodeBlockProvider implements CodeBlockProvider {
+
+    @Override
+    public @Nullable TextRange getCodeBlockRange(Editor editor, PsiFile file) {
+        List<FoldingRange> foldingRanges = LSPFoldingRangeBuilder.getFoldingRanges(file);
```
SCWells72, line 35: Call the folding builder logic to get the folding ranges for the file.

src/main/java/com/redhat/devtools/lsp4ij/features/foldingRange/LSPCodeBlockProvider.java
```diff
@@ -0,0 +1,86 @@
...
+                openBraceChar = documentText.charAt(offset - 1);
+                closeBraceChar = getCloseBraceChar(openBraceChar);
+            } else if ((offset < (documentLength - 1)) && isCloseBraceChar(documentText.charAt(offset + 1))) {
+                closeBraceChar = documentText.charAt(offset + 1);
+                openBraceChar = getOpenBraceChar(closeBraceChar);
+            }
+
+            List<TextRange> containingTextRanges = new ArrayList<>(foldingRanges.size());
+            for (FoldingRange foldingRange : foldingRanges) {
+                TextRange textRange = LSPFoldingRangeBuilder.getTextRange(foldingRange, document, openBraceChar, closeBraceChar);
```
SCWells72, line 64: Convert the folding range into a text range. Previously this would use the provided start/end lines and start/end characters where the latter might not be present. Now it will try to find the full bounding braced block if the range begins on a open brace by searching for the close brace starting at the end line/character. If found, that's the resulting text range exclusive of the braces themselves. If not found, it should gracefully degrade to the previous behavior.

src/main/java/com/redhat/devtools/lsp4ij/features/foldingRange/LSPCodeBlockProvider.java
```diff
@@ -0,0 +1,86 @@
...
+                if (textRange != null) {
+                    // If this is the exact range for which a matching brace was requested, return it
+                    if ((textRange.getStartOffset() == offset) || (textRange.getEndOffset() == offset)) {
+                        return textRange;
```
SCWells72, line 68: If the request started on an open or close brace and we have that exact braced block, return its text range so that brace matching works properly.

src/main/java/com/redhat/devtools/lsp4ij/features/foldingRange/LSPCodeBlockProvider.java
```diff
@@ -0,0 +1,86 @@
...
+                    }
+                    // Otherwise add it to the list of containing ranges and we'll find the smallest at the end
+                    else if (textRange.contains(offset)) {
+                        containingTextRanges.add(textRange);
```
SCWells72, line 72: If not and the original offset was within this text range, collect it so we can find the "closest" (i.e., smallest) containing text range for the original offset below.

src/main/java/com/redhat/devtools/lsp4ij/features/foldingRange/LSPCodeBlockProvider.java
```diff
@@ -0,0 +1,86 @@
...
+                    }
+                }
+            }
+
+            // If we made it here and found containing text ranges, return the smallest one
+            if (!ContainerUtil.isEmpty(containingTextRanges)) {
+                containingTextRanges.sort(Comparator.comparingInt(TextRange::getLength));
```
SCWells72, line 79: Which is exactly what's happening here...

src/main/java/com/redhat/devtools/lsp4ij/features/foldingRange/LSPFoldingRangeBuilder.java
```diff
@@ -44,6 +48,14 @@ public class LSPFoldingRangeBuilder extends CustomFoldingBuilder {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(LSPFoldingRangeBuilder.class);
 
+    // TODO: Can this be derived from the language server? If not, these are the standard for modern languages
+    private static final Map<Character, Character> BRACE_PAIR_CHARS_FWD = Map.of(
```
SCWells72, line 52: Are these available via the language server? We could also make this part of client config, but my guess is that's going to be overkill. The vast majority of languages use paired braces/brackets/parens for block scope, and those that don't generally don't use any kind of block scope delimiters, e.g., Python. Thoughts?

src/main/java/com/redhat/devtools/lsp4ij/features/foldingRange/LSPFoldingRangeBuilder.java
```diff
@@ -54,8 +66,30 @@ protected void buildLanguageFoldRegions(@NotNull List<FoldingDescriptor> descrip
         }
 
         PsiFile file = root.getContainingFile();
+        List<FoldingRange> foldingRanges = getFoldingRanges(file);
```
SCWells72, line 69: Again, the LSP client logic here has been encapsulated into this reusable method.

src/main/java/com/redhat/devtools/lsp4ij/features/foldingRange/LSPFoldingRangeBuilder.java
```diff
@@ -54,8 +66,30 @@ protected void buildLanguageFoldRegions(@NotNull List<FoldingDescriptor> descrip
         }
 
         PsiFile file = root.getContainingFile();
+        List<FoldingRange> foldingRanges = getFoldingRanges(file);
+        if (!ContainerUtil.isEmpty(foldingRanges)) {
+            for (FoldingRange foldingRange : foldingRanges) {
+                TextRange textRange = getTextRange(foldingRange, document);
```
SCWells72, line 72: And again, adapt each folding range to the corresponding text range. The only reason a null text range would be returned is of the language server returned a folding range that didn't make sense, e.g., end before start.

src/main/java/com/redhat/devtools/lsp4ij/features/foldingRange/LSPFoldingRangeBuilder.java
```diff
@@ -54,8 +66,30 @@ protected void buildLanguageFoldRegions(@NotNull List<FoldingDescriptor> descrip
...
+                    } else {
+                        descriptors.add(new FoldingDescriptor(root.getNode(), textRange));
+                    }
+                }
+            }
+        }
+    }
+
+    @NotNull
+    static List<FoldingRange> getFoldingRanges(@Nullable PsiFile file) {
```
SCWells72, line 86: Note that all of the reusable aspects are package-protected so that they don't become part of the public API of LSP4IJ but are usable by the code block provider.

src/main/java/com/redhat/devtools/lsp4ij/features/foldingRange/LSPFoldingRangeBuilder.java
```diff
@@ -64,41 +98,98 @@ protected void buildLanguageFoldRegions(@NotNull List<FoldingDescriptor> descrip
...
+            @NotNull Document document,
+            @Nullable Character openBraceChar,
+            @Nullable Character closeBraceChar
+    ) {
+        TextRange textRange = null;
+
+        String documentText = document.getText();
+        int documentLength = documentText.length();
+
+        int start = LSPFoldingRangeBuilder.getStartOffset(foldingRange, document);
```
SCWells72, line 149: `getStartOffset()` and `getEndOffset()` are already going to be adjusted as they were previously if the start/end characters aren't specified by the language server, so that behavior is retained.

src/main/java/com/redhat/devtools/lsp4ij/features/foldingRange/LSPFoldingRangeBuilder.java
```diff
@@ -64,41 +98,98 @@ protected void buildLanguageFoldRegions(@NotNull List<FoldingDescriptor> descrip
...
+        if ((startChar != null) && ((openBraceChar == null) || (startChar == openBraceChar))) {
+            // If necessary, infer the braces for this block
+            if ((openBraceChar == null) && isOpenBraceChar(startChar)) {
+                openBraceChar = startChar;
+                closeBraceChar = getCloseBraceChar(openBraceChar);
+            }
+
+            int end = LSPFoldingRangeBuilder.getEndOffset(foldingRange, document);
+            // The end offsets can fall a bit short, so look for the closing brace character
+            if (closeBraceChar != null) {
```
SCWells72, line 160: We only search for a close brace if we know definitively that we started on an open brace.

src/main/resources/META-INF/plugin.xml
```diff
@@ -310,6 +310,13 @@
                              language="textmate"
                              implementationClass="com.redhat.devtools.lsp4ij.features.foldingRange.LSPFoldingRangeBuilder"
                              order="first"/>
+        <!-- The code block providers are based on the same information from textDocument/folding -->
+        <codeBlockProvider language="TEXT"
```
SCWells72, line 314: `codeBlockProvider`s are registered for both plain text and TextMate.

src/main/java/com/redhat/devtools/lsp4ij/features/foldingRange/LSPCodeBlockProvider.java
```diff
@@ -0,0 +1,86 @@
...
+ * Code block provider that uses the folding range information from {@link LSPFoldingRangeBuilder}.
+ */
+public class LSPCodeBlockProvider implements CodeBlockProvider {
+
+    @Override
+    public @Nullable TextRange getCodeBlockRange(Editor editor, PsiFile file) {
+        List<FoldingRange> foldingRanges = LSPFoldingRangeBuilder.getFoldingRanges(file);
+        if (!ContainerUtil.isEmpty(foldingRanges)) {
+            Document document = editor.getDocument();
+            String documentText = document.getText();
```
angelozerr, line 38: You should avoid using `document#getText()`, when you want to check some character at a index (like You are doing). because calling getText create a new String:

```
@Override
  public @NotNull String getText() {
    return ReadAction.compute(this::doGetText);
  }

  private @NotNull String doGetText() {
    String s = dereference(myTextString);
    if (s == null) {
      myTextString = new SoftReference<>(s = myText.toString());
    }
    return s;
  }
```

You should use `getCharsSequence()` like Javadoc suggest:

image

SCWells72: That's funny. I was actually using `getCharsSequence()` for most of the development and changed it (what I thought would be) briefly to `getText()` to check one thing. Evidently I forgot to switch that back. Thanks for noticing!

SCWells72:
> @SCWells72 have you a concrete sample (withTypeScript) which shows in action your PR (ex : a demo)?

Yes, of course. Let's start with the code folding enhancement. Here's what code folding looks like in first-class supported languages, e.g., TypeScript in Ultimate Edition/WebStorm:

image

and here's what it looks like in LSP4IJ _without_ the changes in this PR:

image

and finally here's what it looks like in LSP4IJ _with_ the changes in this PR:

image

And here's what the code block provider does:

LSP_Code_Block_Provider

Basically it lets you navigate between brace pairs and to the parent braced block depending on where those actions are activated. I've enabled Presentation Assistant in that little animated gif so you can see what I'm doing and the resulting editor behavior.

Let me know if that doesn't make sense.

angelozerr:
> > @SCWells72 have you a concrete sample (withTypeScript) which shows in action your PR (ex : a demo)?
> 
> Yes, of course. Let's start with the code folding enhancement. Here's what code folding looks like in first-class supported languages, e.g., TypeScript in Ultimate Edition/WebStorm (one folded conditional block and one folded function):
> 
> image
> 
> and here's what it looks like in LSP4IJ _without_ the changes in this PR (exact same folded blocks):
> 
> image
> 
> and finally here's what it looks like in LSP4IJ _with_ the changes in this PR (again, exact same folded blocks):
> 
> image
> 
> And here's what the code block provider does:
> 
> LSP_Code_Block_Provider [ LSP_Code_Block_Provider ](https://private-user-images.githubusercontent.com/7671043/393277231-812b03d0-172e-4f63-8a9a-09f2a6450b35.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MzM0OTU3NjIsIm5iZiI6MTczMzQ5NTQ2MiwicGF0aCI6Ii83NjcxMDQzLzM5MzI3NzIzMS04MTJiMDNkMC0xNzJlLTRmNjMtOGE5YS0wOWYyYTY0NTBiMzUuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI0MTIwNiUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNDEyMDZUMTQzMTAyWiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9ODM5MTZhYzNjMmQ5ZjEzYmM2ZmMwZWExYWEwZmU5MTYzYjc2NDdjMjBmZTkxNTU0ZTU2ZDQwZGE4MTQzOGM4MyZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QifQ.pSPrnJs27e_12KqcFhn8AAbweMgE9pWrnre9Ik4bBXQ) [ ](https://private-user-images.githubusercontent.com/7671043/393277231-812b03d0-172e-4f63-8a9a-09f2a6450b35.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MzM0OTU3NjIsIm5iZiI6MTczMzQ5NTQ2MiwicGF0aCI6Ii83NjcxMDQzLzM5MzI3NzIzMS04MTJiMDNkMC0xNzJlLTRmNjMtOGE5YS0wOWYyYTY0NTBiMzUuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI0MTIwNiUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNDEyMDZUMTQzMTAyWiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9ODM5MTZhYzNjMmQ5ZjEzYmM2ZmMwZWExYWEwZmU5MTYzYjc2NDdjMjBmZTkxNTU0ZTU2ZDQwZGE4MTQzOGM4MyZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QifQ.pSPrnJs27e_12KqcFhn8AAbweMgE9pWrnre9Ik4bBXQ)
> 
> Basically it lets you navigate between brace pairs and to the parent braced block depending on where those actions are activated. I've enabled Presentation Assistant in that little animated gif so you can see what I'm doing and the resulting editor behavior.
> 
> Let me know if that doesn't make sense.

Thanks so much for your great explanation. It looks like amazing!

src/main/java/com/redhat/devtools/lsp4ij/features/foldingRange/LSPFoldingRangeBuilder.java
SCWells72: As discussed in Slack, JetBrains has long had a concrete assumption that structural block delimiters are basically limited to braces, brackets, and parens as evidenced in their own `AbstractFileType`:

image

In that same discussion, we decided that it should be fine to support exactly those same delimiters until/unless we learn of a language being integrated via a language server that supports something other than those for explicit block designation (i.e., not like Python or YAML where blocks are based on indentation).

angelozerr:
@SCWells72 we can see that in an another PR but it would be very nice to write tests with folding like I did for cmpletion, semantic tokens, documentation etc https://github.com/redhat-developer/lsp4ij/tree/9ca402a8726304e6da7efc470d0009c8d72b8c5c/src/test/java/com/redhat/devtools/lsp4ij/features

angelozerr:
@SCWells72 your PR is amazing, but IMHO we should provide the capability to:

 * disable this behavior
 * choose the tokens pairs according the LS.

And LSPFoldingFeatures is exactly the component which could be used to configure that. By default we could enable it and returns standard pairs like you did.

A custom language server could override those pairs (for user defined, let's keep with default pairs for the moment).

What do you think about that?

SCWells72:
@angelozerr, I'm happy to add tests, though it won't likely happen until the beginning of next week.

I'm curious as to why you think someone would want the behavior disabled, though. It's basically just implementing the standard IDE language behavior and features for LSP-integrated languages. Certainly if there were bugs they'd need to be addressed, but what would be the use case for disabling it.

Regarding making the token pairs configurable, I'll actually push back on that one until/unless there is clear evidence of a language -- whether integrated via LSP or just in general -- that uses some other token pair to establish a block that would be reported via `textDocument/foldingRange`. And maybe there are existing examples in which case I'd totally change my stance, but if there aren't, I do feel that like that adding complexity without a driving concrete, real-world use case.

I'll definitely take care of unit tests next week. We can continue to discuss the configuration aspects.

angelozerr:
Thats very cool. What about writting tests in a separate PR to merge it asap.
The like merging features in order to test him asap

src/main/resources/META-INF/plugin.xml
angelozerr: Please update the doc at https://github.com/redhat-developer/lsp4ij/blob/main/docs/LSPSupport.md#folding-range with codeBlock extension point sionce it is working by default only for TEXT and textmate

angelozerr:
It isworking like a charm, thanks @SCWells72 !

src/main/resources/META-INF/plugin.xml
SCWells72: Ah, you mean specifically the `codeBlockProvider` extension point. I get it now. I'll update the docs. I'll also include an updated demo using TypeScript. This will all be in a new, distinct PR.

SCWells72: See #676.

PR #669, opened 2024-12-06
feat: Text document selection range support (issue 651)

SCWells72:
This PR integrates the LSP `textDocument/selectionRange` feature into the IDE's `extendWordSelectionHandler` EP that provides the **Extend/Shrink Selection** actions (which are among my absolute favorites in the IDE). Note that it does not yet include unit tests, but I do plan to add them once I know that there's interest in moving forward with it.

Here's an example of it working in TypeScript:

LSP_Extend_Shrink_Selection

and here it is in CSS:

LSP_Extend_Shrink_Selection_CSS

I'll add a few notes on things that I know will likely warrant further discussion.


docs/LSPApi.md
```diff
@@ -14,6 +14,7 @@ The [LSPClientFeatures](https://github.com/redhat-developer/lsp4ij/blob/main/src
 - [LSP documentSymbol feature](#lsp-documentSymbol-feature)
```
SCWells72, line 14: I recommend that you disable whitespace differences because I ran several of the existing files through the Java formatter against the project's code style and it resulted in minor whitespace changes.

src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPClientFeatures.java
```diff
@@ -1105,6 +1140,7 @@ public void setServerCapabilities(@NotNull ServerCapabilities serverCapabilities
 
     @Nullable
     public TextDocumentServerCapabilityRegistry<? extends TextDocumentRegistrationOptions> getCapabilityRegistry(String method) {
+        // TODO: Can these be changed to else/if? Or can this be changed to switch?
```
SCWells72, line 1143: I don't know how often this methods gets called, but it looks like it can be changed to either `if ... else if ... else if ...` or ideally `switch (method)` to make it more efficient. I'm happy to do that as part of these changes if you'd like.

src/main/java/com/redhat/devtools/lsp4ij/features/selectionRange/LSPSelectionRangeSupport.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    public CompletableFuture<List<SelectionRange>> getSelectionRanges(SelectionRangeParams params) {
+        return super.getFeatureData(params);
+    }
+
+    @Override
+    protected boolean isValidLSPFuture() {
+        // Disable caching until/unless it can take into account the params, in this case, the caret offset
+        return false;
```
SCWells72, line 48: Is there another way to disable caching here? The issue is that the default caching is based on the file's modification count, but the response here is parameterized by the offset at the caret when the feature is invoked, so file-level caching causes issues. Is there a precedent for caching by both file modification count and some other parameter such as offset?

src/main/java/com/redhat/devtools/lsp4ij/features/selectionRange/LSPBasicWordSelectionFilter.java
```diff
@@ -0,0 +1,35 @@
...
+
+import com.intellij.openapi.util.Condition;
+import com.intellij.psi.PsiElement;
+import com.intellij.psi.PsiFile;
+import com.redhat.devtools.lsp4ij.LSPFileSupport;
+
+/**
+ * Prevents {@link com.intellij.codeInsight.editorActions.wordSelection.WordSelectioner} from adding selections to LSP files.
+ */
+public class LSPBasicWordSelectionFilter implements Condition<PsiElement> {
```
SCWells72, line 21: Without this, `WordSelectioner` will add the entire file as a selection range.

src/main/java/com/redhat/devtools/lsp4ij/features/selectionRange/LSPSelectionRangeSupport.java
angelozerr, line 48: See LSPCompletionSupport which have the same usecase.

angelozerr:
Wow again an impressive PR. Let me some times to give yiu detailled review.

angelozerr:
@SCWells72 it would be nice also to update the LSPSupport markdown to enable the selectiinRange and add a detailled info about implementation and a little demo with some expl1nation line shortcut to use.

I could do that if you wish.

SCWells72:
I'll update `LSPSupport.md` now. I'll let you add a demo if that's okay.

src/main/java/com/redhat/devtools/lsp4ij/features/selectionRange/LSPExtendWordSelectionHandler.java
```diff
@@ -0,0 +1,156 @@
...
+            return false;
+        }
+
+        PsiFile file = element.getContainingFile();
+        if ((file == null) || !file.isValid()) {
+            return false;
+        }
+
+        // Only if textDocument/selectionRange is supported for the file
+        return LSPFileSupport.getSupport(file).getSelectionRangeSupport() != null;
```
angelozerr, line 62: This check is every time not null. You need to check if there is a language server which support selectionrange:

```java
// Check if the file can support the feature
return LanguageServiceAccessor.getInstance(project)
                .hasAny(file.getVirtualFile(), ls -> ls.getClientFeatures().getSelectionRangeFeature().isSelectionRangeSupported(file)));
```

src/main/java/com/redhat/devtools/lsp4ij/features/selectionRange/LSPExtendWordSelectionHandler.java
```diff
@@ -0,0 +1,156 @@
...
+            LOGGER.error("Error while consuming LSP 'textDocument/selectionRanges' request", e);
+            return Collections.emptyList();
+        }
+
+        if (!isDoneNormally(selectionRangesFuture)) {
+            return Collections.emptyList();
+        }
+
+        // textDocument/selectionRanges has been collected correctly, create list of IJ SelectionDescriptor from LSP SelectionRange list
+        return selectionRangesFuture.getNow(null);
```
angelozerr, line 137: It seems method should never return null, replace with:

`return selectionRangesFuture.getNow(Colletions.emptyList());`

docs/LSPSupport.md
```diff
@@ -370,6 +370,15 @@ Here is an example with the [Go Language Server](./user-defined-ls/gopls.md) sho
 
 ![textDocument/foldingRange](./images/lsp-support/textDocument_foldingRange.gif)
 
+### Selection range
+
+[textDocument/selectionRange](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_selectionRange) is implemented with
+the `extendWordSelectionHandler` extension point and used via the IDE's **Extend Selection** and **Shrink Selection** actions.
```
angelozerr, line 376: Please add shortcut sample in () to execute those actions.

angelozerr (changes requested):

docs/LSPSupport.md
SCWells72: Sorry, but what does "in ()" mean? Is there already a list of supported keyboard shortcuts for LSP4IJ-integrated actions?

src/main/java/com/redhat/devtools/lsp4ij/features/selectionRange/LSPExtendWordSelectionHandler.java
SCWells72: Please see what I've done in the latest commit. I looked for something that does what the new method `isSupported(PsiFile, Predicate<LSPClientFeatures>)` is doing but couldn't find anything quite like it. If something does exist, please let me know and I'll migrate to it; if not, perhaps that method should be moved somewhere more common for future needs? If so, where? `LanguageServerAccessor`? `LSP4IJUtils`? ???

src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPClientFeatures.java
```diff
@@ -1105,91 +1140,74 @@ public void setServerCapabilities(@NotNull ServerCapabilities serverCapabilities
 
     @Nullable
     public TextDocumentServerCapabilityRegistry<? extends TextDocumentRegistrationOptions> getCapabilityRegistry(String method) {
-        if (LSPRequestConstants.TEXT_DOCUMENT_CODE_ACTION.equals(method)) {
-            // register 'textDocument/codeAction' capability
-            return getCodeActionFeature().getCodeActionCapabilityRegistry();
+        if (method == null) {
```
SCWells72, line 1143: This is the result of changing this to use `switch` which should use a jump table instead of having to evaluate each and every condition.

src/main/java/com/redhat/devtools/lsp4ij/features/selectionRange/LSPExtendWordSelectionHandler.java
```diff
@@ -0,0 +1,194 @@
...
+        if ((file == null) || !file.isValid()) {
+            return false;
+        }
+
+        // Only if textDocument/selectionRange is supported for the file
+        return isSupported(file, LSPClientFeatures::getSelectionRangeFeature);
+    }
+
+    // TODO: This seems so incredibly boiler-plate that it should already exist somewhere common
+    private static boolean isSupported(@NotNull PsiFile file,
```
SCWells72, line 73: Is there not already something that does this in an encapsulated manner like this? If not, should this be moved into a common location?

docs/LSPSupport.md
angelozerr: I wanted to say add (Ctrl+W) / (Ctrl+Shift+W)

src/main/java/com/redhat/devtools/lsp4ij/features/selectionRange/LSPExtendWordSelectionHandler.java
```diff
@@ -0,0 +1,194 @@
...
+            return false;
+        }
+
+        PsiFile file = element.getContainingFile();
+        if ((file == null) || !file.isValid()) {
+            return false;
+        }
+
+        // Only if textDocument/selectionRange is supported for the file
+        return isSupported(file, LSPClientFeatures::getSelectionRangeFeature);
```
angelozerr, line 69: Remove your isSupported method with

```java

        Project project = file.getProject();
        if (project.isDisposed()) {
            return false;
        }
        
        // Only if textDocument/selectionRange is supported for the file
        return LanguageServiceAccessor.getInstance(project)
                .hasAny(file.getVirtualFile(), ls -> ls.getClientFeatures().getSelectionRangeFeature().isSelectionRangeSupported(file));

```

src/main/java/com/redhat/devtools/lsp4ij/features/selectionRange/LSPExtendWordSelectionHandler.java
angelozerr: The short answer is https://github.com/redhat-developer/lsp4ij/pull/669/files#r1877498632

Here a detailled explanation: when you need to check that an IJ feature is enabled (ex : to enable/disable Implementations menu item,  selectionrange like you need for your PR) it MUST be very fast otherwise IJ will freeze. 

You must never wait for starting language servers which can takes some times (some language servers can take 15sec to start it), so IJ could freeze during 15 sec.

Using document matcher requires to open a read action. In Ij Quarkus we use that to associate our Quarkus Language server to Java files only when the project have some Java class in the classpath.

Using hasAny is very fast. It checks:

 * if there is one or more language server which are associated to the file
 * if there is one language server has thecapability but ONLY when language server is started.

It means that if you execute Ctrl+W and language servers is starting, the action will be disable but I think it is better to loose the Ctrl+W feature and avoid freezing IJ.

angelozerr, line 73: Please remove this method

src/test/java/com/redhat/devtools/lsp4ij/features/selectionRange/TypeScriptSelectionRangeTest.java
```diff
@@ -0,0 +1,225 @@
...
+ ******************************************************************************/
+
+package com.redhat.devtools.lsp4ij.features.selectionRange;
+
+import com.redhat.devtools.lsp4ij.fixtures.LSPSelectionRangeFixtureTestCase;
+
+/**
+ * Selection range tests by emulating LSP 'textDocument/selectionRange' responses from the typescript-language-server.
+ */
+public class TypeScriptSelectionRangeTest extends LSPSelectionRangeFixtureTestCase {
```
angelozerr, line 19: Thanks so much to have written tests, that's great!

src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPClientFeatures.java
angelozerr: Thanks so much to have done this horrible change!

SCWells72:
New PR for this work opened as #678.

PR #674, opened 2024-12-07
feat: Update external references on successful rename via LSP (issue 670)

SCWells72:
Once the user has chosen a different name for the identifier under the caret, external references to that identifier are collected. If any were found, they are updated via `PsiReference#handleElementRename()` on successful completion of the LSP rename action.


src/main/java/com/redhat/devtools/lsp4ij/features/rename/LSPRenameRefactoringDialog.java
```diff
@@ -174,8 +199,19 @@ static void doRename(@NotNull LSPRenameParams renameParams,
...
                 // Apply the rename from the LSP WorkspaceEdit list
-                WriteCommandAction
-                        .runWriteCommandAction(psiFile.getProject(), () -> workspaceEdits.forEach(workspaceEditData -> LSPIJUtils.applyWorkspaceEdit(workspaceEditData.edit())));
+                WriteCommandAction.runWriteCommandAction(psiFile.getProject(), () -> {
+                    workspaceEdits.forEach(workspaceEditData -> LSPIJUtils.applyWorkspaceEdit(workspaceEditData.edit()));
+
+                    // Update any found external references with the new name
+                    externalReferences.forEach(externalReference -> {
+                        // Don't let a single failed external reference keep us from updating other references
+                        try {
```
SCWells72, line 208: For relational integrity purposes, it's important that failure of a single rename propagation not short-circuit the rest.

src/main/java/com/redhat/devtools/lsp4ij/features/rename/LSPRenameRefactoringDialog.java
```diff
@@ -192,4 +228,90 @@ protected void dispose() {
...
+        Set<PsiReference> externalReferences = new LinkedHashSet<>();
+
+        Document document = LSPIJUtils.getDocument(file);
+        TextRange wordTextRange = document != null ? LSPIJUtils.getWordRangeAt(document, file, offset) : null;
+        if (wordTextRange != null) {
+            LSPPsiElement wordElement = new LSPPsiElement(file, wordTextRange);
+            String wordText = wordElement.getText();
+            if (StringUtil.isNotEmpty(wordText)) {
+                // When testing, just collect references on the current thread
+                if (ApplicationManager.getApplication().isUnitTestMode()) {
```
SCWells72, line 242: I don't love this, but it's important that the textual search occur under a modal progress indicator, and that doesn't work well with the rename refactoring unit tests. The same logic occurs either way, but in the real app it occurs under a modal progress indicator, and during tests it occurs on the current thread.

src/main/java/com/redhat/devtools/lsp4ij/features/rename/LSPRenameRefactoringDialog.java
```diff
@@ -192,4 +231,123 @@ protected void dispose() {
...
+                            Collections.emptyList();
+                    for (PsiReference reference : references) {
+                        // Deduplicate using a unique key with reference type, file, and text range
+                        String referenceKey = getReferenceKey(reference);
+                        if (referenceKey != null) {
+                            // Only add references we haven't added previously
+                            if (!externalReferencesByKey.containsKey(referenceKey)) {
+                                PsiElement targetElement = reference.resolve();
+                                PsiFile targetFile = targetElement != null ? targetElement.getContainingFile() : null;
+                                TextRange targetTextRange = targetFile != null ? targetElement.getTextRange() : null;
```
angelozerr, line 293: Move declaration of targetTextRange  and targetText  in the ifblock (L297))

src/main/java/com/redhat/devtools/lsp4ij/features/rename/LSPRenameRefactoringDialog.java
```diff
@@ -192,4 +231,123 @@ protected void dispose() {
...
+        if (!ContainerUtil.isEmpty(languageServers)) {
+            // If any supporting language server is case-sensitive, the search must be case-sensitive; it's better to
+            // miss changing things that should have been changed than to change things that should not
+            for (LanguageServerItem languageServer : languageServers) {
+                LSPClientFeatures clientFeatures = languageServer.getClientFeatures();
+                if (clientFeatures != null) {
+                    LSPRenameFeature renameFeature = clientFeatures.getRenameFeature();
+                    if (renameFeature.isRenameSupported(psiFile)) {
+                        LSPCompletionFeature completionFeature = clientFeatures.getCompletionFeature();
+                        if (completionFeature.isCaseSensitive(psiFile)) {
```
angelozerr, line 332: It is strange to use completion feature for rename. We discussed about this caseSenstive property and perhaps it would be better to move caseSensitive in the clientFeatures (we can do that since wehave not released) and the JSON client configuration content will declare just 

```
{
  "caseSensitive": true
}
```

If you do that, please update JSON Schema docs and description (this caseSensitive is used n completion and rename both).

SCWells72: Yes, this is what I meant when that flag was introduced and I said that case-sensitivity is a language/grammar-level setting and not specific to any single feature associated with that language/grammar. I agree that the setting should be up-leveled, particularly before it has ever been included in a release. I'm happy to do that as part of this work.

angelozerr: > Yes, this is what I meant when that flag was introduced and I said that case-sensitivity is a language/grammar-level setting and not specific to any single feature associated with that language/grammar.

I am sorry it was difficult for me to understand without usecases. Now I underdtand more with this new usecase.

>  I agree that the setting should be up-leveled, particularly before it has ever been included in a release. I'm happy to do that as part of this work.

Thanks!

SCWells72: Oh, not an issue at all. That wasn't meant as an "I told you so." I think we took the right approach of solving the problem we had until we saw another aspect of it. Here we are, and thankfully before the thing that has to be changed has been put into end user hands. No issues at all! I'll get it moved to a feature-independent location in the next day or two.

angelozerr: I need to discuss with @fbricon but I think we should create asap a new release to rmbed the fix https://github.com/redhat-developer/lsp4ij/pull/671

If you want we czn merge your PR and I will do the fix in order to embed the PR in the release.

docs/LSPApi.md
```diff
@@ -2,6 +2,7 @@
 
 The [LSPClientFeatures](https://github.com/redhat-developer/lsp4ij/blob/main/src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPClientFeatures.java) API allows you to customize the behavior of LSP features, including:
 
+- [Client-only features](#client-only-features)
```
SCWells72, line 5: Let me know if this break-out of client-only features isn't what you'd prefer.

src/main/java/com/redhat/devtools/lsp4ij/LanguageServiceAccessor.java
```diff
@@ -423,6 +423,26 @@ public CompletableFuture<Set<LanguageServerDefinition>> getAsyncMatched() {
...
 
+    /**
+     * Returns the language server wrappers for the specified file synchronously.
+     *
+     * @param file the file
+     * @return the language server wrappers that apply to the file
+     */
+    @ApiStatus.Internal
+    @NotNull
+    public Set<LanguageServerWrapper> getMatchedLanguageServerWrappersSync(@NotNull VirtualFile file) {
```
SCWells72, line 434: If there's a better way to do this without adding this method, please let me know. It's needed to encapsulate finding external references by just a `PsiFile`, an offset, and a processor.

src/main/java/com/redhat/devtools/lsp4ij/features/rename/LSPRenameRefactoringDialog.java
```diff
@@ -192,4 +221,32 @@ protected void dispose() {
         super.dispose();
     }
 
+    @NotNull
+    private static Set<PsiReference> getExternalReferences(@NotNull PsiFile file, int offset) {
+        Set<PsiReference> externalReferences = new LinkedHashSet<>();
+
+        // When testing, just collect references on the current thread
+        if (ApplicationManager.getApplication().isUnitTestMode()) {
```
SCWells72, line 229: This is consistent with the existing precedent of keeping things a bit simpler when tests execute.

src/main/java/com/redhat/devtools/lsp4ij/server/definition/launching/ClientConfigurationSettings.java
```diff
@@ -19,30 +19,20 @@
...
-        public boolean caseSensitive = false;
-    }
-
     /**
      * Client-side code workspace symbol settings.
      */
     public static class ClientConfigurationWorkspaceSymbolSettings {
         /**
-         * Whether or not completions should be offered as case-sensitive. Defaults to false.
+         * Whether or not the language server can support the IDE's Go To Class action efficiently. Defaults to false.
```
SCWells72, line 27: Fixed a copy-paste comment from earlier.

src/main/java/com/redhat/devtools/lsp4ij/usages/LSPExternalReferencesFinder.java
```diff
@@ -0,0 +1,172 @@
...
+import com.redhat.devtools.lsp4ij.features.LSPPsiElement;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.*;
+
+/**
+ * Utility class that helps to process/find external references to LSP4IJ-based (pseudo-)elements.
+ */
+public final class LSPExternalReferencesFinder {
```
SCWells72, line 45: This is the common utility class for finding/processing external usages of LSP4IJ's pseudo-elements.

src/main/java/com/redhat/devtools/lsp4ij/usages/LSPUsageSearcher.java
```diff
@@ -92,6 +94,15 @@ public void processElementUsages(@NotNull PsiElement element, @NotNull Processor
         } catch (Exception e) {
             LOGGER.error("Error while collection LSP Usages", e);
         }
+
+        // For completeness' sake, also collect external usages to LSP (pseudo-)elements
+        int offset = ReadAction.compute(() -> {
```
SCWells72, line 99: This is how external usages are also included.

src/main/java/com/redhat/devtools/lsp4ij/usages/LSPUsageSearcher.java
```diff
@@ -89,9 +91,14 @@ public void processElementUsages(@NotNull PsiElement element, @NotNull Processor
                     }
                 }
             }
+        } catch (ProcessCanceledException pce) {
```
SCWells72, line 94: I changed this behavior in one place based on an observed error. This same change may need to be rolled out anywhere there is `try ... catch (Exception) ...` where the caught exception could be `ProcessCanceledException`. That should only be caught/handled in VERY specific situations. Otherwise it should be propagated to the IDE.

src/main/java/com/redhat/devtools/lsp4ij/usages/LSPExternalReferencesFinder.java
```diff
@@ -0,0 +1,172 @@
...
+                caseSensitive
+        );
+    }
+
+    private static boolean isCaseSensitive(@NotNull Set<LanguageServerWrapper> languageServerWrappers, @NotNull PsiFile psiFile) {
+        // If any supporting language server is case-sensitive, the search must be case-sensitive; it's better to
+        // miss changing things that should have been changed than to change things that should not
+        for (LanguageServerWrapper languageServerWrapper : languageServerWrappers) {
+            LSPClientFeatures clientFeatures = languageServerWrapper.getClientFeatures();
+            if ((clientFeatures != null) && clientFeatures.isCaseSensitive(psiFile)) {
```
angelozerr, line 154: clientFeatures != null should never happen

src/main/java/com/redhat/devtools/lsp4ij/usages/LSPExternalReferencesFinder.java
```diff
@@ -0,0 +1,172 @@
...
+                                                  @NotNull TextRange wordTextRange,
+                                                  @Nullable ProgressIndicator progressIndicator,
+                                                  @NotNull Processor<PsiReference> processor) {
+        VirtualFile virtualFile = file.getVirtualFile();
+        if (virtualFile == null) {
+            return;
+        }
+
+        Project project = file.getProject();
+        Set<LanguageServerWrapper> languageServerWrappers = LanguageServiceAccessor.getInstance(project).getMatchedLanguageServerWrappersSync(virtualFile);
```
angelozerr, line 92: Replace with:

```java
boolean caseSensitive =
                LanguageServiceAccessor.getInstance(file.getProject())
                .hasAny(file.getVirtualFile(), ls -> ls.getClientFeatures().isCaseSensitive(file));
```

and remove your new method getMatchedLanguageServerWrappersSync.

src/main/java/com/redhat/devtools/lsp4ij/usages/LSPExternalReferencesFinder.java
SCWells72: I'll annotate that method as `@NotNull` in the commit that removes that check so that there's no confusion going forward.

src/main/java/com/redhat/devtools/lsp4ij/usages/LSPUsageSearcher.java
angelozerr: It should be done like this with the main classes.

angelozerr:
I cannot test the PR but code looks good.

Thanks @SCWells72 !

PR #679, opened 2024-12-10
test: Added unit tests for LSP-based code folding and the code block provider (issue 673)

SCWells72:
The title is pretty self-explanatory. The only thing worth pointing out explicitly is that I've added some utility methods to the common base test case class for deriving the before/after index of specific substrings in test file bodies to avoid having to hard-code them.

### DIFF
--- a/src/main/java/com/redhat/devtools/lsp4ij/features/foldingRange/LSPFoldingRangeBuilder.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/features/foldingRange/LSPFoldingRangeBuilder.java
@@ -13,6 +13,7 @@ package com.redhat.devtools.lsp4ij.features.foldingRange;
 import com.intellij.lang.ASTNode;
 import com.intellij.lang.folding.CustomFoldingBuilder;
 import com.intellij.lang.folding.FoldingDescriptor;
+import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.editor.Document;
 import com.intellij.openapi.progress.ProcessCanceledException;
 import com.intellij.openapi.util.TextRange;
@@ -62,8 +63,8 @@ public class LSPFoldingRangeBuilder extends CustomFoldingBuilder {
     protected void buildLanguageFoldRegions(@NotNull List<FoldingDescriptor> descriptors,
                                             @NotNull PsiElement root,
                                             @NotNull Document document, boolean quick) {
-        // if quick flag is set, we do nothing here
-        if (quick) {
+        // if quick flag is set and not testing, we do nothing here
+        if (quick && !ApplicationManager.getApplication().isUnitTestMode()) {
             return;
         }
 
@@ -219,6 +220,7 @@ public class LSPFoldingRangeBuilder extends CustomFoldingBuilder {
 
     @Override
     public boolean isDumbAware() {
-        return false;
+        // Allow folding in dumb mode only during unit testing
+        return ApplicationManager.getApplication().isUnitTestMode();
     }
 }
--- a/src/test/java/com/redhat/devtools/lsp4ij/features/foldingRange/TypeScriptCodeBlockProviderTest.java
+++ b/src/test/java/com/redhat/devtools/lsp4ij/features/foldingRange/TypeScriptCodeBlockProviderTest.java
@@ -23,8 +23,6 @@ public class TypeScriptCodeBlockProviderTest extends LSPCodeBlockProviderFixture
     }
 
     public void testCodeBlocks() {
-        String testFilename = "demo.ts";
-
         // language=json
         String mockFoldingRangesJson = """
                 [
@@ -40,7 +38,7 @@ public class TypeScriptCodeBlockProviderTest extends LSPCodeBlockProviderFixture
                 """;
 
         assertCodeBlock(
-                testFilename,
+                "demo.ts",
                 """
                         export class Demo {
                             demo() {<start>
@@ -52,7 +50,7 @@ public class TypeScriptCodeBlockProviderTest extends LSPCodeBlockProviderFixture
         );
 
         assertCodeBlock(
-                testFilename,
+                "demo.ts",
                 """
                         export class Demo {<start>
                             <caret>demo() {
@@ -64,7 +62,7 @@ public class TypeScriptCodeBlockProviderTest extends LSPCodeBlockProviderFixture
         );
 
         assertCodeBlock(
-                testFilename,
+                "demo.ts",
                 """
                         export class Demo {<caret><start>
                             demo() {
@@ -76,7 +74,7 @@ public class TypeScriptCodeBlockProviderTest extends LSPCodeBlockProviderFixture
         );
 
         assertCodeBlock(
-                testFilename,
+                "demo.ts",
                 """
                         export class Demo <caret>{<start>
                             demo() {
@@ -88,7 +86,7 @@ public class TypeScriptCodeBlockProviderTest extends LSPCodeBlockProviderFixture
         );
 
         assertCodeBlock(
-                testFilename,
+                "demo.ts",
                 """
                         export class Demo {<start>
                             demo() {
@@ -100,7 +98,7 @@ public class TypeScriptCodeBlockProviderTest extends LSPCodeBlockProviderFixture
         );
 
         assertCodeBlock(
-                testFilename,
+                "demo.ts",
                 """
                         export class Demo {<start>
                             demo() {
@@ -112,7 +110,7 @@ public class TypeScriptCodeBlockProviderTest extends LSPCodeBlockProviderFixture
         );
 
         assertCodeBlock(
-                testFilename,
+                "demo.ts",
                 """
                         export class <caret><start><end>Demo {
                             demo() {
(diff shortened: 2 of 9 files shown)


src/main/java/com/redhat/devtools/lsp4ij/features/foldingRange/LSPFoldingRangeBuilder.java
```diff
@@ -62,8 +63,8 @@ public class LSPFoldingRangeBuilder extends CustomFoldingBuilder {
     protected void buildLanguageFoldRegions(@NotNull List<FoldingDescriptor> descriptors,
                                             @NotNull PsiElement root,
                                             @NotNull Document document, boolean quick) {
-        // if quick flag is set, we do nothing here
-        if (quick) {
+        // if quick flag is set and not testing, we do nothing here
+        if (quick && !ApplicationManager.getApplication().isUnitTestMode()) {
```
SCWells72, line 67: Tests are run in dumb mode so if we want to test this, we need to allow that, at least for test execution.

src/main/java/com/redhat/devtools/lsp4ij/features/foldingRange/LSPFoldingRangeBuilder.java
```diff
@@ -219,6 +220,7 @@ protected boolean isRegionCollapsedByDefault(@NotNull ASTNode node) {
 
     @Override
     public boolean isDumbAware() {
-        return false;
+        // Allow folding in dumb mode only during unit testing
+        return ApplicationManager.getApplication().isUnitTestMode();
```
SCWells72, line 224: Same thing here...allow dumb mode during test execution.

src/test/java/com/redhat/devtools/lsp4ij/features/foldingRange/TypeScriptCodeBlockProviderTest.java
```diff
@@ -0,0 +1,71 @@
...
+                          },
+                          {
+                            "startLine": 1,
+                            "endLine": 2
+                          }
+                        ]
+                        """,
+                new int[][]{
+                        // From "console" to the begin and end of the "demo()" function's braced block
+                        {beforeFirst(DEMO_TS_FILE_BODY, "console", 1), afterFirst(DEMO_TS_FILE_BODY, "{", 2), beforeLast(DEMO_TS_FILE_BODY, "}", 2)},
```
SCWells72, line 55: Each of these test scenarios include the initial caret offset, the expected caret offset when move-to-block-start occurs, and the expected caret offset when move-to-block-end occurs. As you can see, I've added some utility methods to the common base test case class that allow us to derive those indices dynamically rather than hard-coding them, (hopefully) resulting in much more flexible tests if/when test file bodies change.

src/test/java/com/redhat/devtools/lsp4ij/features/foldingRange/TypeScriptFoldingRangeTest.java
```diff
@@ -0,0 +1,88 @@
...
+    private static final String DEMO_TS_FILE_BODY = """
+            export class Demo {
+                demo() {
+                    console.log('demo');
+                }
+            }
+            """;
+
+    // Demo class braced block exclusive of braces
+    private static final TextRange DEMO_CLASS_BODY_TEXT_RANGE = TextRange.create(afterFirst(DEMO_TS_FILE_BODY, "{", 1), beforeLast(DEMO_TS_FILE_BODY, "}", 1));
```
SCWells72, line 33: Using the same utility methods here to derive the expected text ranges dynamically.

src/test/java/com/redhat/devtools/lsp4ij/features/selectionRange/TypeScriptSelectionRangeTest.java
```diff
@@ -18,15 +18,18 @@
  */
 public class TypeScriptSelectionRangeTest extends LSPSelectionRangeFixtureTestCase {
 
+    private static final String DEMO_TS_FILE_NAME = "demo.ts";
```
SCWells72, line 21: No new tests were added here, but I extracted common constants for repeated strings for consistency with what I did in the new tests added in this PR.

src/test/java/com/redhat/devtools/lsp4ij/fixtures/LSPCodeBlockProviderFixtureTestCase.java
```diff
@@ -0,0 +1,78 @@
...
+        // Initialize the language server
+        try {
+            LanguageServiceAccessor.getInstance(project)
+                    .getLanguageServers(file.getVirtualFile(), null, null)
+                    .get(5000, TimeUnit.MILLISECONDS);
+        } catch (Exception e) {
+            fail(e.getMessage());
+        }
+
+        EditorTestUtil.buildInitialFoldingsInBackground(editor);
```
SCWells72, line 60: This is the important step required to trigger folding behavior in the file. You'll see it in both of the new base test case classes.

src/test/java/com/redhat/devtools/lsp4ij/fixtures/LSPCodeInsightFixtureTestCase.java
```diff
@@ -72,4 +73,85 @@ private void registerServer() {
     private void unregisterServer() {
         LanguageServersRegistry.getInstance().removeServerDefinition(myFixture.getProject(), serverDefinition);
     }
+
+    // Utility methods for getting the before/after index of the n'th occurrence of a substring of the test file body
```
SCWells72, line 77: Here are the utility methods referenced earlier.

src/test/java/com/redhat/devtools/lsp4ij/fixtures/LSPFoldingRangeFixtureTestCase.java
```diff
@@ -0,0 +1,79 @@
...
+        // Initialize the language server
+        try {
+            LanguageServiceAccessor.getInstance(file.getProject())
+                    .getLanguageServers(file.getVirtualFile(), null, null)
+                    .get(5000, TimeUnit.MILLISECONDS);
+        } catch (Exception e) {
+            fail(e.getMessage());
+        }
+
+        EditorTestUtil.buildInitialFoldingsInBackground(editor);
```
SCWells72, line 60: Again, force population of the folding model in test mode.

src/test/java/com/redhat/devtools/lsp4ij/features/selectionRange/TypeScriptSelectionRangeTest.java
```diff
@@ -91,9 +94,8 @@ public void testSelectionRanges_qualifierExpression() {
 
     public void testSelectionRanges_callExpression() {
         assertSelectionRanges(
-                "demo.ts",
-                // language=typescript
-                "console.log('message');",
+                DEMO_TS_FILE_NAME,
```
angelozerr, line 97: It is just my opinion, but I find using constant is not easy to read, because you need to jump to the constant to know the value.

But if you prefer , keep like this.

src/test/java/com/redhat/devtools/lsp4ij/features/foldingRange/TypeScriptCodeBlockProviderTest.java
```diff
@@ -0,0 +1,70 @@
...
+                            "endLine": 3
+                          },
+                          {
+                            "startLine": 1,
+                            "endLine": 2
+                          }
+                        ]
+                        """,
+                new int[][]{
+                        // From "console" to the begin and end of the "demo()" function's braced block
```
angelozerr, line 53: Instead of expect some offset, I think will more readable to expect a string which have the same content of the editor but with some annotations like <fold1> ends with </fold1>

It follows the same idea with the <caret> that yiu can use in a test (see for instance completion test, where I use <caret> instead the offset number

angelozerr: Here a sample with folding ranges:

```java
public void testFoldingRanges() {
        assertFoldingRanges(
                "demo.ts",
                """
                export class Demo {
                    demo() {
                        console.log('demo');
                    }
                }
                """,
                // language=json
                """
                        [
                          {
                            "startLine": 0,
                            "endLine": 3
                          },
                          {
                            "startLine": 1,
                            "endLine": 2
                          }
                        ]
                        """,
                """
                <fold1>export class Demo {
                    demo() {
                        console.log('demo');
                    }
                }</fold1>
                """
        );
    }
}
```

SCWells72: I've changed the tests to provide a balance between readability and maintainability. Tokens are now used, but directly in the test file source, and they're both extracted for offset information and stripped from the file body for the actual `PsiFile` creation. That way you're only maintaining one version of the file body with tokens representing key locations.

Let me know if that doesn't address your concerns.

angelozerr:
Impressive!

PR #687, opened 2024-12-11
feat: Allow **Ctrl+Space** to toggle whether or not word completions are displayed

SCWells72:
Minor enhancement to the earlier LSP code completion enhancement such that word completions are not shown on auto-completion or odd completion trigger counts. This means that pressing **Ctrl+Space** repeatedly will now toggle word completions on and off but they won't be shown when there are LSP-based completions on auto-complete or initial explicit completion trigger.

Here's an example of the behavior with this change where you can see that there are not word completions by default, but pressing **Ctrl+Space** will toggle them on and off:

LSP_Word_Completions_Toggle


angelozerr:
Thanks @SCWells72

PR #688, opened 2024-12-11
fix: Implementations must return `null` and not an empty array to indicate that they cannot service the request

SCWells72:
Changed the return value to `null` when we cannot service the request via LSP. That matches the contract according to its JavaDoc.


angelozerr:
Thanks

PR #691, opened 2024-12-12
fix: Restoring VFS refresh of bundled JSON schema file

SCWells72:
Unfortunately I can't use `reloadPsi()` for two reasons:
1. There's not guaranteed to be a non-null `Project` when the JSON schema is loaded because it's loaded via a project-independent extension point.
2. `reloadPsi()` uses `FileManagerImpl` which is annotated as `@ApiStatus.Internal` and therefore almost certainly won't be accepted by JetBrains in a new plugin version submission. That likely needs to change to do something different, perhaps just what I'm doing here with a two-phase VFS refresh.


angelozerr:
Thanks @SCWells72 !

PR #692, opened 2024-12-12
fix: Added support for non-standard `member` token type reported by the TypeScript language server for `static` methods

SCWells72:
This is probably best demonstrated visually. Here's the syntax highlighting without this fix:

image

and here it is with this fix:

image

Note the lack of highlighting for the declaration and usage of `static` method `demo` in the first image and the proper highlighting of both in the second image.


angelozerr:
@SCWells72 I understand that the result is very nice, but I would like to really avoid hard code custom token for each language server. 

My idea for user defined language server is to have 

 * a new tab `Semantic Highlight` which cotains a JSON editor
 * to configure like vscode https://code.visualstudio.com/api/language-extensions/semantic-highlight-guide#custom-token-types-and-modifiers

angelozerr:
I see that you rebase your PR but as I have tried to explain, I would like to have a more generic support to do that like vscode provides.

Hope you will understand.

SCWells72:
> I see that you rebase your PR but as I have tried to explain, I would like to have a more generic support to do that like vscode provides.
> 
> Hope you will understand.

Oh, I'm just rebasing all open branches to keep things fresh for when I return to them. I'm not actively working on these right now. I'll change this to a draft PR to prevent any further confusion.

SCWells72:
Closing this as the simple approach here has been rejected and the more extensive proposed alternative is probably out-of-scope for me anytime soon.

angelozerr:
I am sorry @SCWells72 but I hope you understand my point of view. My fear is that each ls will want to add their own custom tokens

SCWells72:
> I am sorry @SCWells72 but I hope you understand my point of view. My fear is that each ls will want to add their own custom tokens

No, I understand. However, if you truly want to be strict about that stance, then the current special handling for `label` needs to be changed as well as it already violates that concept. I made the changes that I made for those PR based on the existing precedent around `label`.

PR #694, opened 2024-12-12
fix: Changed `file` to be `@Nullable` since a `null` value could be provided to the method

SCWells72:
Otherwise this could result in an effective NPE thrown by the `@NotNull` annotation at runtime.


angelozerr:
Good catch @SCWells72 !

PR #703, opened 2024-12-13
feat: Added a file view provider for LSP files that yields fake PSI elements for words.

SCWells72:
I'll switch this to draft as I'm just looking to use this as a reference point for discussion.

### DIFF
--- a/src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPDocumentFileViewProviderFactory.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPDocumentFileViewProviderFactory.java
@@ -1,0 +1,78 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Red Hat, Inc.
+ * Distributed under license by Red Hat, Inc. All rights reserved.
+ * This program is made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution,
+ * and is available at https://www.eclipse.org/legal/epl-v20.html
+ *
+ * Contributors:
+ * Red Hat, Inc. - initial API and implementation
+ ******************************************************************************/
+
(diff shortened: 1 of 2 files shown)


angelozerr:
My first impression is that you are using documentSymbol as AST.  

I understand that it could fix a lot of problem with IJ integration but my fear is about performance. For little file, documentSymbol is fast but for large file it can be extremely slow and can consume a lotnof CPU.

A very good sample is with our XML language server used in vscode-xml. If your xml file is large Outline which consumes documentSymbol can be very slow.

The second problem is that language server doesnt provide documentSymbol or provides it without detailled ast nodes.

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPDocumentFileViewProviderFactory.java
```diff
@@ -0,0 +1,78 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Red Hat, Inc.
```
angelozerr, line 2: 2024

SCWells72:
> I love the idea but I need to play with the PR to see if it fixes several issues.

Oh, it almost certainly needs quite a bit of work to make it do what it should _and_ (more importantly) not do anything it shouldn't. That's why this is still in draft state. I may take a look at it again a bit later this week if I free up from some of my current obligations.

angelozerr:
> > I love the idea but I need to play with the PR to see if it fixes several issues.
> 
> Oh, it almost certainly needs quite a bit of work to make it do what it should _and_ (more importantly) not do anything it shouldn't. That's why this is still in draft state. I may take a look at it again a bit later this week if I free up from some of my current obligations.


Please ping me when you will think the PR will be ready.

SCWells72:
Closing this for now as making everything else play nicely with these changes is probably more than I can take on anytime soon.

Issue #447, opened 2024-08-12
Unable to see code completion after the colon in comments.


CppCXY:
LSP4IJ: 
![completion1](https://github.com/user-attachments/assets/9a4ac221-4c17-4d0c-a410-1a697f75bffb)

Vscode:

![completion2](https://github.com/user-attachments/assets/30a2077b-80b8-48df-a620-947a5049af04)

but I can see completion response:
```log
[Trace - 16:51:13] Received response 'textDocument/completion - (190)' in 0ms.
Result: [
  {
    "label": "unused",
    "kind": 20,
    "detail": "diagnostic code",
    "deprecated": false
  },
  {
    "label": "undefined-global",
    "kind": 20,
    "detail": "diagnostic code",
    "deprecated": false
  },
  {
    "label": "need-import",
    "kind": 20,
    "detail": "diagnostic code",
    "deprecated": false
  },
  {
    "label": "deprecated",
    "kind": 20,
    "detail": "diagnostic code",
    "deprecated": false
  },
  {
    "label": "access-private-member",
    "kind": 20,
    "detail": "diagnostic code",
    "deprecated": false
  },
  {
    "label": "access-protected-member",
    "kind": 20,
    "detail": "diagnostic code",
    "deprecated": false
  },
  {
    "label": "access-package-member",
    "kind": 20,
    "detail": "diagnostic code",
    "deprecated": false
  },
  {
    "label": "missing-parameter",
    "kind": 20,
    "detail": "diagnostic code",
    "deprecated": false
  },
  {
    "label": "no-discard",
    "kind": 20,
    "detail": "diagnostic code",
    "deprecated": false
  },
  {
    "label": "disable-global-define",
    "kind": 20,
    "detail": "diagnostic code",
    "deprecated": false
  },
  {
    "label": "local-const-reassign",
    "kind": 20,
    "detail": "diagnostic code",
    "deprecated": false
  },
  {
    "label": "type-not-found",
    "kind": 20,
    "detail": "diagnostic code",
    "deprecated": false
  },
  {
    "label": "undefined-field",
    "kind": 20,
    "detail": "diagnostic code",
    "deprecated": false
  },
  {
    "label": "inject-field-fail",
    "kind": 20,
    "detail": "diagnostic code",
    "deprecated": false
  }
]

```

CppCXY:
you can direct install EmmyLua2 in intellij idea
[...]

PR #705, opened 2024-12-14
feat: Improved completion sorting that takes into account the current word and prefix (strict then camel-hump)

SCWells72:
Opening this initially as a draft PR, but this brings over some of the completion ordering enhancements that I've added to the completion contributors in my own plugin. The idea is that the initial offered completions should be as close to what the user is wanting as possible given the available information at the time that completions are triggered.

First, completion sorting now takes into account configured language case sensitivity.

Next, if completions are triggered in the context of an existing identifier, any matches for that identifier should be listed first with exact matches (again, taking into account case-sensitivity) before starts-with matches:

image

Next, the prefix is considered with exact prefix matches listed before camel-hump prefix matches:

**Exact match**
image

**Camel-hump match**
image

Here's another example from CSS since all of the above are from TypeScript:

image

If this looks good -- taking into account feedback, of course -- I'll update `CompletionItemComparatorTest` with comprehensive unit tests for the new comparator logic.


SCWells72:
@angelozerr mentioned that sorting should be using `sortText`, so it's worth providing context on why the new contextual aspects of sorting use `label` instead:

You expressly don't want to do that because it defeats the entire point of the changes I made.

The problem you're solving is that you're trying to insert some word of some sort, typically a keyword. If you sort by the sort text -- which is often something like an integer value as shown in the image below (being sorted as text which has its own issues) -- the resulting sort order doesn't take into account insertion context at all, e..g:

image

If you look at how IntelliJ IDEA and other JetBrains IDEs sort completions, they take into account the same things I am here, but also things like assignability and declaration proximity which we don't have (easily) from LSP, and of course also machine learning statistics around previously-selected completions in a given context.

With this change it now tries to sort by the keyword/identifier you're trying to insert by looking at the word that already exists at the point of insertion and/or the partial word prefix that's already been typed. And if none of that applies, it degrades gracefully to the original sorting behavior anyway...well, now with proper consideration of language grammar case-sensitivity.

The best way to get a feel for what this change does is to try it. Try completions with and without this change and see the difference in accuracy in terms of what you're actually trying to insert at the caret. I think you'll find that, with these changes, it's quite accurate, and without, you have to type more characters before the `PrefixMatcher` filter kicks in enough to narrow the offered completions down to what you originally wanted.

SCWells72:
@angelozerr **after the holidays** (i.e., "don't even look at this anytime soon while you should be spending time with family and friends"), I'd love to start working this one toward a merge. I've (hopefully) removed any concerns about behavioral regressions by making this an opt-in client-side configuration option aside from the case-sensitivty sorting improvements that likely should have been included with the prior code completion case-sensitivity work.

I'll add some additional unit test cases for the comparator changes and that should make this whole.

src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPCompletionFeature.java
```diff
@@ -97,7 +97,7 @@ public boolean isCompletionSupported(@NotNull PsiFile file) {
      * @param file the file.
      * @return true the file associated with a language server can support resolve completion and false otherwise.
      */
-    public boolean isResolveCompletionSupported(@Nullable PsiFile file) {
+    public boolean isResolveCompletionSupported(@NotNull PsiFile file) {
```
SCWells72, line 100: All callers were passing non-null values and the `isResolveCompletionSupported()` expects a non-null argument, so I changed the nullability annotation to match reality.

src/main/java/com/redhat/devtools/lsp4ij/features/completion/LSPCompletionContributor.java
```diff
@@ -119,12 +119,19 @@ private void addCompletionItems(@NotNull CompletionParameters parameters,
             items.addAll(completionList.getItems());
         }
 
-        // Sort by item.sortText
-        items.sort(completionProposalComparator);
+        PsiFile originalFile = parameters.getOriginalFile();
+        LSPClientFeatures clientFeatures = languageServer.getClientFeatures();
+
+        // Sort the completions as appropriate based on client configuration
+        boolean useContextAwareSorting = clientFeatures.getCompletionFeature().useContextAwareSorting(originalFile);
```
SCWells72, line 126: The new features are now opt-in. When disabled (which is the default), the only difference vs. prior behavior should be properly respecting the language case-sensitivity config setting.

src/main/java/com/redhat/devtools/lsp4ij/server/definition/launching/ClientConfigurationSettings.java
```diff
@@ -20,7 +20,17 @@
...
 public class ClientConfigurationSettings {
     /**
-     * Client-side code workspace symbol settings.
+     * Client-side code completion settings.
+     */
+    public static class ClientConfigurationCompletionSettings {
+        /**
+         * Whether or not client-side context-aware completion sorting should be used. Defaults to false.
+         */
+        public boolean useContextAwareSorting = false;
```
SCWells72, line 29: The opt-in config option.

src/test/java/com/redhat/devtools/lsp4ij/features/completion/CompletionItemComparatorTest.java
```diff
@@ -16,10 +16,11 @@
 import java.util.ArrayList;
 import java.util.List;
 
-import static org.junit.jupiter.api.Assertions.*;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 public class CompletionItemComparatorTest {
-    private final CompletionItemComparator comparator = new CompletionItemComparator();
+    private final CompletionItemComparator comparator = new CompletionItemComparator(null, null, false);
```
SCWells72, line 23: I'll add more test cases here to confirm proper behavior of the new context-aware features of this comparator.

angelozerr:
@SCWells72 I promise you that I will review your PR after my holiday.

A thing that it would be very nice is to improve the compute of the prefix (when textedit is nul, if I remember the prefix is computed for each completion item which is very badl) and to provide an api to customize it to fix https://github.com/redhat-developer/lsp4ij/issues/447#issuecomment-2286334837

SCWells72:
This will be the second PR that I plan to run through review. I'll add unit tests for the enhanced comparator today if possible and switch this from draft to ready-for-review.

SCWells72:
@angelozerr I've added extensive unit tests for the comparator so this should now be ready for review.

src/main/java/com/redhat/devtools/lsp4ij/server/definition/launching/UserDefinedCompletionFeature.java
```diff
@@ -0,0 +1,31 @@
+/*******************************************************************************
+ * Copyright (c) 2024 Red Hat Inc. and others.
```
angelozerr, line 2: 2025

angelozerr (changes requested):

src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPCompletionFeature.java
angelozerr, line 100: Please add this new method api in the docs Lsp api markdown

src/main/java/com/redhat/devtools/lsp4ij/server/definition/launching/UserDefinedCompletionFeature.java
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Adds client-side code completion configuration features.
+ */
+public class UserDefinedCompletionFeature extends LSPCompletionFeature {
+
+    @Override
+    public boolean useContextAwareSorting(@NotNull PsiFile file) {
+        UserDefinedLanguageServerDefinition serverDefinition = (UserDefinedLanguageServerDefinition) getClientFeatures().getServerDefinition();
+        ClientConfigurationSettings clientConfiguration = serverDefinition.getLanguageServerClientConfiguration();
+        return clientConfiguration != null ? clientConfiguration.completion.useContextAwareSorting : super.useContextAwareSorting(file);
```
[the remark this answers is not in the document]
SCWells72: As this is an opt-in feature, I didn't update any of the templates to specify the new config setting. The default value is, of course, `false`/disabled. Personally I think that most folks would want it, but I didn't want to assume that by setting it to true in the templates. If you agree that I should, I'll be happy to do so.

angelozerr:
@SCWells72 I have played with your PR and I love the idea but the sort have some strange behaviour, see the following demo:

CompletionSortBug

src/main/java/com/redhat/devtools/lsp4ij/server/definition/launching/UserDefinedCompletionFeature.java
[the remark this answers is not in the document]
SCWells72: I pushed a commit that enables this by default, so I don't think any template updates are needed at this point.

SCWells72:
> @SCWells72 I have played with your PR and I love the idea but the sort have some strange behaviour, see the following demo:

This should now be addressed.

src/main/java/com/redhat/devtools/lsp4ij/features/completion/LSPCompletionContributor.java
SCWells72: Actually the new feature is now enabled by default.

PR #707, opened 2024-12-16
fix: Trying to integrate diagnostics into bulk code inspection results (issue 476)

SCWells72:
This is not yet complete, but I'm opening a draft PR to discuss what needs to be done to complete it properly. I'll add questions in the PR itself.

### DIFF
--- a/src/main/java/com/redhat/devtools/lsp4ij/LanguageServerWrapper.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/LanguageServerWrapper.java
@@ -635,8 +635,8 @@ public class LanguageServerWrapper implements Disposable {
      *                 by using a blocking read action.
      * @return the completable future with the language server instance or null.
      */
-    CompletableFuture<@Nullable LanguageServer> connect(@NotNull VirtualFile file,
-                                                        @Nullable Document document) {
+    public CompletableFuture<@Nullable LanguageServer> connect(@NotNull VirtualFile file,
+                                                               @Nullable Document document) {
         if (file != null && file.exists()) {
             return connect(LSPIJUtils.toUri(file), file, document);
         }
--- a/src/main/java/com/redhat/devtools/lsp4ij/features/codeAction/LSPLazyCodeActionIntentionAction.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/features/codeAction/LSPLazyCodeActionIntentionAction.java
@@ -62,13 +62,13 @@ public class LSPLazyCodeActionIntentionAction implements IntentionAction {
     @Override
     public @IntentionName @NotNull String getText() {
         loadCodeActionIfNeeded();
-        return title;
+        return title != null ? title : "";
     }
 
     @Override
     public @NotNull @IntentionFamilyName String getFamilyName() {
         loadCodeActionIfNeeded();
-        return familyName;
+        return familyName != null ? familyName : "";
     }
 
     @Override
--- a/src/main/java/com/redhat/devtools/lsp4ij/features/diagnostics/LSPDiagnosticAnnotator.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/features/diagnostics/LSPDiagnosticAnnotator.java
@@ -18,21 +18,31 @@ import com.intellij.lang.annotation.AnnotationHolder;
 import com.intellij.lang.annotation.ExternalAnnotator;
 import com.intellij.openapi.editor.Document;
 import com.intellij.openapi.editor.Editor;
+import com.intellij.openapi.progress.ProcessCanceledException;
 import com.intellij.openapi.progress.ProgressManager;
 import com.intellij.openapi.util.Key;
+import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.psi.PsiFile;
-import com.redhat.devtools.lsp4ij.LSPIJUtils;
-import com.redhat.devtools.lsp4ij.LSPVirtualFileData;
-import com.redhat.devtools.lsp4ij.LanguageServersRegistry;
-import com.redhat.devtools.lsp4ij.LanguageServiceAccessor;
+import com.redhat.devtools.lsp4ij.*;
+import com.redhat.devtools.lsp4ij.client.ExecuteLSPFeatureStatus;
+import com.redhat.devtools.lsp4ij.client.indexing.ProjectIndexingManager;
 import com.redhat.devtools.lsp4ij.features.AbstractLSPExternalAnnotator;
 import org.eclipse.lsp4j.Diagnostic;
+import org.eclipse.lsp4j.services.LanguageServer;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.net.URI;
 import java.util.Collections;
 import java.util.List;
+import java.util.concurrent.CancellationException;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+
+import static com.redhat.devtools.lsp4ij.internal.CompletableFutures.isDoneNormally;
+import static com.redhat.devtools.lsp4ij.internal.CompletableFutures.waitUntilDone;
 
 /**
  * Intellij {@link ExternalAnnotator} implementation which get the current LSP diagnostics for a given file and translate
@@ -40,19 +50,34 @@ import java.util.List;
  */
 public class LSPDiagnosticAnnotator extends AbstractLSPExternalAnnotator<Boolean, Boolean> {
 
+    private static final Logger LOGGER = LoggerFactory.getLogger(LSPDiagnosticAnnotator.class);
+
     private static final Key<Boolean> APPLIED_KEY = Key.create("lsp.diagnostic.annotator.applied");
 
+    private final boolean forBulkInspection;
+
     public LSPDiagnosticAnnotator() {
+        this(false);
+    }
+
+    LSPDiagnosticAnnotator(boolean forBulkInspection) {
         super(APPLIED_KEY);
+        this.forBulkInspection = forBulkInspection;
+    }
+
+    @Nullable
+    @Override
+    public Boolean collectInformation(@NotNull PsiFile file) {
+        if (!LanguageServersRegistry.getInstance().isFileSupported(file)) {
+            return Boolean.FALSE;
+        }
+        return Boolean.TRUE;
     }
 
     @Nullable
     @Override
     public Boolean collectInformation(@NotNull PsiFile file, @NotNull Editor editor, boolean hasErrors) {
-        if (!LanguageServersRegistry.getInstance().isFileSupported(file)) {
-            return Boolean.FALSE;
-        }
-        return Boolean.TRUE;
+        return collectInformation(file);
     }
 
     @Override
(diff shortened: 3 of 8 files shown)


src/main/java/com/redhat/devtools/lsp4ij/LanguageServerWrapper.java
```diff
@@ -635,8 +635,8 @@ public boolean canOperate(Project project) {
      *                 by using a blocking read action.
      * @return the completable future with the language server instance or null.
      */
-    CompletableFuture<@Nullable LanguageServer> connect(@NotNull VirtualFile file,
-                                                        @Nullable Document document) {
+    public CompletableFuture<@Nullable LanguageServer> connect(@NotNull VirtualFile file,
```
SCWells72, line 638: Had to make this `public` so that I could connect files that have not ever been opened in editor tabs to the respective language servers. See `LSPDiagnosticAnnotator#connectIfNecessary()` for more details.

src/main/java/com/redhat/devtools/lsp4ij/features/codeAction/LSPLazyCodeActionIntentionAction.java
```diff
@@ -62,13 +62,13 @@ public void setLazyCodeActions(LSPLazyCodeActionProvider lazyCodeActions) {
     @Override
     public @IntentionName @NotNull String getText() {
         loadCodeActionIfNeeded();
-        return title;
+        return title != null ? title : "";
```
SCWells72, line 65: These are annotated as `@NotNull` but I was seeing runtime errors where they could be null. I've made them defensive, but perhaps these intentions should not ever be created if they don't have sufficient info to be used?

src/main/java/com/redhat/devtools/lsp4ij/features/diagnostics/LSPDiagnosticAnnotator.java
```diff
@@ -18,43 +18,68 @@
...
  * Intellij {@link ExternalAnnotator} implementation which get the current LSP diagnostics for a given file and translate
  * them into Intellij {@link com.intellij.lang.annotation.Annotation}.
  */
 public class LSPDiagnosticAnnotator extends AbstractLSPExternalAnnotator<Boolean, Boolean> {
 
+    private static final Logger LOGGER = LoggerFactory.getLogger(LSPDiagnosticAnnotator.class);
+
     private static final Key<Boolean> APPLIED_KEY = Key.create("lsp.diagnostic.annotator.applied");
 
+    private final boolean forBulkInspection;
```
SCWells72, line 57: This allows the `LSPDiagnosticAnnotatorInspection` to create an instance of the annotator that knows that it's running as part of a bulk/batch inspection run.

src/main/java/com/redhat/devtools/lsp4ij/features/diagnostics/LSPDiagnosticAnnotator.java
```diff
@@ -18,43 +18,68 @@
...
+
+    LSPDiagnosticAnnotator(boolean forBulkInspection) {
         super(APPLIED_KEY);
+        this.forBulkInspection = forBulkInspection;
     }
 
     @Nullable
     @Override
-    public Boolean collectInformation(@NotNull PsiFile file, @NotNull Editor editor, boolean hasErrors) {
+    public Boolean collectInformation(@NotNull PsiFile file) {
```
SCWells72, line 70: This missing signature of `collectInformation()` needs to be present as it's called during bulk/batch inspection runs. The original just delegates to this one with the same implementation as before.

src/main/java/com/redhat/devtools/lsp4ij/features/diagnostics/LSPDiagnosticAnnotator.java
```diff
@@ -66,28 +91,86 @@ public void doApply(@NotNull PsiFile file, Boolean applyAnnotator, @NotNull Anno
...
         for (var ls : servers) {
-            LSPVirtualFileData data = ls.getLSPVirtualFileData(fileUri);
-            if (data != null) {
-                // The file is mapped with the current language server
-                var ds = data.getDiagnosticsForServer();
-                // Loop for LSP diagnostics to transform it to Intellij annotation.
-                for (Diagnostic diagnostic : ds.getDiagnostics()) {
-                    ProgressManager.checkCanceled();
-                    createAnnotation(diagnostic, document, file, ds, holder);
+            // TODO: Is there not a way to check whether the server is valid for this file here?
```
SCWells72, line 106: I noticed that `LSPDiagnosticFeature#isSupported()` always returns `true`, so this block is being executed for language servers that don't support the file. Is there some way to check whether the language server supports the file, e.g., based on its mapping configuration?

src/main/java/com/redhat/devtools/lsp4ij/features/diagnostics/LSPDiagnosticAnnotator.java
```diff
@@ -66,28 +91,86 @@ public void doApply(@NotNull PsiFile file, Boolean applyAnnotator, @NotNull Anno
...
+            // Force a connection for batch inspection
+            boolean needsDisconnect = forBulkInspection && connectIfNecessary(ls, file);
```
SCWells72, line 108: In a bulk inspection run, connect the file to the language server if not already connected and note that we did so so that we can disconnect the file when we're finished with it.

src/main/java/com/redhat/devtools/lsp4ij/features/diagnostics/LSPDiagnosticAnnotator.java
```diff
@@ -66,28 +91,86 @@ public void doApply(@NotNull PsiFile file, Boolean applyAnnotator, @NotNull Anno
...
+            try {
+                // Check again as the connection above may not have succeeded
+                if (ls.isConnectedTo(fileUri)) {
+                    LSPVirtualFileData data = ls.getLSPVirtualFileData(fileUri);
+                    if (data != null) {
+                        // The file is mapped with the current language server
+                        var ds = data.getDiagnosticsForServer();
+                        // Loop for LSP diagnostics to transform it to Intellij annotation.
+                        // TODO: When running for bulk inspection, we need to wait for diagnostics here; not sure how to
```
SCWells72, line 117: This is the main remaining issue...diagnostics are not guaranteed to be populated by the time `getDiagnosticsForServer()` returns. Is there a way to pull diagnostics explcitly? If not, is there a (good) way to wait for the diagnostics to be published? We really need them sychronously during bulk inspection runs.

src/main/java/com/redhat/devtools/lsp4ij/features/diagnostics/LSPDiagnosticAnnotator.java
```diff
@@ -66,28 +91,86 @@ public void doApply(@NotNull PsiFile file, Boolean applyAnnotator, @NotNull Anno
...
             }
         }
+
+        return false;
     }
 
     private static void createAnnotation(@NotNull Diagnostic diagnostic,
                                          @NotNull Document document,
-                                         @Nullable PsiFile file,
+                                         @NotNull PsiFile file,
```
SCWells72, line 173: All existing callers of this had already guaranteed a non-null file, so I changed the parameter nullability annotation accordingly since the code is using it as if it were non-null anyway.

src/main/java/com/redhat/devtools/lsp4ij/features/diagnostics/LSPDiagnosticAnnotatorInspection.java
```diff
@@ -0,0 +1,45 @@
...
+    @Override
+    public boolean runForWholeFile() {
+        return true;
+    }
+
+    @Override
+    public ProblemDescriptor @Nullable [] checkFile(@NotNull PsiFile file,
+                                                    @NotNull InspectionManager manager,
+                                                    boolean isOnTheFly) {
+        if (!isOnTheFly) {
```
SCWells72, line 39: Only call the annotator when invoked for a bulk inspection run.

src/main/java/com/redhat/devtools/lsp4ij/features/diagnostics/LSPDiagnosticHandler.java
```diff
@@ -93,7 +94,10 @@ private void updateDiagnostics(@NotNull PublishDiagnosticsParams params, @NotNul
             // Trigger Intellij validation to execute
             // {@link LSPDiagnosticAnnotator}.
             // which translates LSP Diagnostics into Intellij Annotation
-            DaemonCodeAnalyzer.getInstance(project).restart(psiFile);
+            Editor editor = LSPIJUtils.editorForElement(psiFile);
```
SCWells72, line 97: No need to restart the analyzer if the file isn't in an editor tab, and that will be the case for many files during a bulk inspection run if we stick with reacting to published diagnostic notifications.

src/main/java/com/redhat/devtools/lsp4ij/features/diagnostics/LSPDiagnosticsForServer.java
```diff
@@ -77,7 +77,9 @@ private Map<Diagnostic, LSPLazyCodeActions> toMap(List<Diagnostic> diagnostics,
         List<Diagnostic> sortedDiagnostics = diagnostics
                 .stream()
                 .sorted((d1, d2) -> {
-                    if (Ranges.containsRange(d1.getRange(), d2.getRange())) {
+                    if (Objects.equals(d1.getRange(), d2.getRange())) {
```
SCWells72, line 80: This comparator didn't have any logic for equivalent ranges, so I added it.

src/main/resources/META-INF/plugin.xml
```diff
@@ -249,6 +249,17 @@ L
                 id="LSPDiagnosticAnnotator"
                 language=""
                 implementationClass="com.redhat.devtools.lsp4ij.features.diagnostics.LSPDiagnosticAnnotator"/>
+        <localInspection
```
SCWells72, line 252: This is the inspection-based adapter for the external annotator.

src/main/java/com/redhat/devtools/lsp4ij/features/diagnostics/LSPDiagnosticAnnotator.java
```diff
@@ -66,28 +91,86 @@ public void doApply(@NotNull PsiFile file, Boolean applyAnnotator, @NotNull Anno
...
+                        //  do that since it seems it's all based on publishDiagnostics notifications
+                        for (Diagnostic diagnostic : ds.getDiagnostics()) {
+                            ProgressManager.checkCanceled();
+                            createAnnotation(diagnostic, document, file, ds, holder);
+                        }
+                    }
+                }
+            } finally {
+                if (needsDisconnect) {
+                    ls.disconnect(fileUri, false);
```
SCWells72, line 127: I don't think we can/should disconnect until we know that we have diagnostics for the file, but assuming we add some kind of synchronous wait for them above, this should still be what want to release resources that were only needed for the batch run.

src/main/java/com/redhat/devtools/lsp4ij/features/diagnostics/LSPDiagnosticAnnotator.java
SCWells72: It looks like we just need to implement a client for `TextDocumentService#diagnostic()` to compute/retrieve diagnostics synchronously in this situation.

src/main/java/com/redhat/devtools/lsp4ij/LanguageServerWrapper.java
angelozerr: If I understand correctly your code, you need to call connect for file which are not opened, right?

If it that we need to find an another way because connect call LSP didOpen notification which means "the file is opened in an editor".

SCWells72: Yeah, this approach is a dead end. I tried to add support for [pull diagnostics](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_pullDiagnostics) instead as that would do exactly what we need, but unfortunately that capability is not very well supported, at least among the language servers that I tried. The TypeScript server doesn't support it, and the CSS server said it did but then threw an exception when it received the request.

I think this overall feature may need to wait until pull diagnostics are more broadly supported.

SCWells72:
Closing because it's not really possible -- at least in any efficient manner -- without pull diagnostics, and pull diagnostics aren't supported well enough by existing language servers yet.

PR #711, opened 2024-12-17
fix: Addressed issues with LSP Find Usages implementation (issue 708)

SCWells72:
I'm unfortunately unable to reproduce the actual error seen by the reporting user in both 2023.2 or 2024.3, but wrapping the entire call in a read action should resolve all downstream requirements for the read lock.

### DIFF
--- a/src/main/java/com/redhat/devtools/lsp4ij/usages/LSPExternalReferencesFinder.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/usages/LSPExternalReferencesFinder.java
@@ -26,6 +26,7 @@ import com.intellij.psi.PsiFile;
 import com.intellij.psi.PsiReference;
 import com.intellij.psi.impl.source.resolve.reference.PsiReferenceUtil;
 import com.intellij.psi.search.PsiSearchHelper;
+import com.intellij.psi.search.SearchScope;
 import com.intellij.psi.search.UsageSearchContext;
 import com.intellij.util.Processor;
 import com.redhat.devtools.lsp4ij.LSPIJUtils;
@@ -48,12 +49,27 @@ public final class LSPExternalReferencesFinder {
     /**
      * Processes all external references for the LSP4IJ element at the offset in the specified file.
      *
-     * @param file      the file for which the element at the specified offset should be processed for external references
-     * @param offset    the offset of the element in the file
-     * @param processor the external reference processor
+     * @param file        the file for which the element at the specified offset should be processed for external references
+     * @param offset      the offset of the element in the file
+     * @param processor   the external reference processor
      */
     public static void processExternalReferences(@NotNull PsiFile file,
                                                  int offset,
+                                                 @NotNull Processor<PsiReference> processor) {
+        processExternalReferences(file, offset, ReadAction.compute(file::getUseScope), processor);
+    }
+
+    /**
+     * Processes all external references for the LSP4IJ element at the offset in the specified file.
+     *
+     * @param file        the file for which the element at the specified offset should be processed for external references
+     * @param offset      the offset of the element in the file
+     * @param searchScope the search scope
+     * @param processor   the external reference processor
+     */
+    public static void processExternalReferences(@NotNull PsiFile file,
+                                                 int offset,
+                                                 @NotNull SearchScope searchScope,
                                                  @NotNull Processor<PsiReference> processor) {
         VirtualFile virtualFile = file.getVirtualFile();
         if (virtualFile != null) {
@@ -61,12 +77,13 @@ public final class LSPExternalReferencesFinder {
             TextRange wordTextRange = document != null ? LSPIJUtils.getWordRangeAt(document, file, offset) : null;
             if (wordTextRange != null) {
                 LSPPsiElement wordElement = new LSPPsiElement(file, wordTextRange);
-                String wordText = wordElement.getText();
+                String wordText = ReadAction.compute(wordElement::getText);
                 if (StringUtil.isNotEmpty(wordText)) {
                     processExternalReferences(
                             file,
                             wordText,
                             wordTextRange,
+                            searchScope,
                             ProgressManager.getInstance().getProgressIndicator(),
                             processor
                     );
@@ -78,6 +95,7 @@ public final class LSPExternalReferencesFinder {
     private static void processExternalReferences(@NotNull PsiFile file,
                                                   @NotNull String wordText,
                                                   @NotNull TextRange wordTextRange,
+                                                  @NotNull SearchScope searchScope,
                                                   @Nullable ProgressIndicator progressIndicator,
                                                   @NotNull Processor<PsiReference> processor) {
         VirtualFile virtualFile = file.getVirtualFile();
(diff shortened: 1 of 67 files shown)


src/main/java/com/redhat/devtools/lsp4ij/usages/LSPExternalReferencesFinder.java
```diff
@@ -48,25 +49,41 @@ private LSPExternalReferencesFinder() {
...
+     * Processes all external references for the LSP4IJ element at the offset in the specified file.
+     *
+     * @param file        the file for which the element at the specified offset should be processed for external references
+     * @param offset      the offset of the element in the file
+     * @param searchScope the search scope
+     * @param processor   the external reference processor
+     */
+    public static void processExternalReferences(@NotNull PsiFile file,
+                                                 int offset,
+                                                 @NotNull SearchScope searchScope,
```
SCWells72, line 72: For Find Usages, we need the configured search scope.

src/main/java/com/redhat/devtools/lsp4ij/usages/LSPExternalReferencesFinder.java
```diff
@@ -48,25 +49,41 @@ private LSPExternalReferencesFinder() {
...
-     * @param offset    the offset of the element in the file
-     * @param processor the external reference processor
+     * @param file        the file for which the element at the specified offset should be processed for external references
+     * @param offset      the offset of the element in the file
+     * @param processor   the external reference processor
      */
     public static void processExternalReferences(@NotNull PsiFile file,
                                                  int offset,
                                                  @NotNull Processor<PsiReference> processor) {
+        processExternalReferences(file, offset, ReadAction.compute(file::getUseScope), processor);
```
SCWells72, line 59: This is the original signature that derives the search scope from the file.

src/main/java/com/redhat/devtools/lsp4ij/usages/LSPExternalReferencesFinder.java
```diff
@@ -48,25 +49,41 @@ private LSPExternalReferencesFinder() {
...
+                                                 @NotNull SearchScope searchScope,
+                                                 @NotNull Processor<PsiReference> processor) {
         VirtualFile virtualFile = file.getVirtualFile();
         if (virtualFile != null) {
             Document document = LSPIJUtils.getDocument(virtualFile);
             TextRange wordTextRange = document != null ? LSPIJUtils.getWordRangeAt(document, file, offset) : null;
             if (wordTextRange != null) {
                 LSPPsiElement wordElement = new LSPPsiElement(file, wordTextRange);
-                String wordText = wordElement.getText();
+                String wordText = ReadAction.compute(wordElement::getText);
```
SCWells72, line 80: Tactical read actions where required.

src/main/java/com/redhat/devtools/lsp4ij/usages/LSPUsageSearcher.java
```diff
@@ -53,11 +55,30 @@ public class LSPUsageSearcher extends CustomUsageSearcher {
...
+        if (file == null) {
+            return;
+        }
+
+        VirtualFile virtualFile = file.getVirtualFile();
+        if (virtualFile == null) {
+            return;
+        }
+
+        Project project = element.getProject();
```
SCWells72, line 69: Don't proceed unless there's at least one configured language server definition for this file that supports usages. That really addresses Joachim's core issue where it was doing this work in a file that shouldn't have been considered by LSP4IJ at all.

src/main/java/com/redhat/devtools/lsp4ij/usages/LSPUsageSearcher.java
```diff
@@ -98,7 +115,12 @@ public void processElementUsages(@NotNull PsiElement element, @NotNull Processor
         }
 
         // For completeness' sake, also collect external usages to LSP (pseudo-)elements
-        LSPExternalReferencesFinder.processExternalReferences(file, element.getTextOffset(), reference -> processor.process(new UsageInfo2UsageAdapter(new UsageInfo(reference))));
+        LSPExternalReferencesFinder.processExternalReferences(
+                file,
+                ReadAction.compute(element::getTextOffset),
```
SCWells72, line 120: This is the specific call that was causing the exception in Joachim's report. Now we get the text offset in a read action.

src/main/java/com/redhat/devtools/lsp4ij/usages/LSPUsageSearcher.java
```diff
@@ -98,7 +115,12 @@ public void processElementUsages(@NotNull PsiElement element, @NotNull Processor
         }
 
         // For completeness' sake, also collect external usages to LSP (pseudo-)elements
-        LSPExternalReferencesFinder.processExternalReferences(file, element.getTextOffset(), reference -> processor.process(new UsageInfo2UsageAdapter(new UsageInfo(reference))));
+        LSPExternalReferencesFinder.processExternalReferences(
+                file,
+                ReadAction.compute(element::getTextOffset),
+                options.searchScope,
+                reference -> processor.process(new UsageInfo2UsageAdapter(new UsageInfo(reference)))
```
angelozerr, line 122: If I remember you can have a ReadAction problem here. (See the first call process.process which compute a ReadAction

SCWells72: Do you mean that `new UsageInfo2UsageAdapter(new UsageInfo(reference))` should be enclosed in a `ReadAction` here like in the other similar constructor calls in the same method. I don't _think_ that's necessary because the `UsageInfo2UsageAdapter` constructor already includes its own `ReadAction`, and the `UsageInfo` constructor also looks fine. Having said that, I'm totally fine wrapping that in a `ReadAction` both for consistency and out of a sense of caution if you'd like. Let me know.

SCWells72: I went ahead and added it.

src/main/java/com/redhat/devtools/lsp4ij/usages/LSPUsageSearcher.java
angelozerr: Great!

angelozerr:
@SCWells72 please fix conflicts.

src/main/java/com/redhat/devtools/lsp4ij/usages/LSPUsageSearcher.java
angelozerr: Yes it is that, I don't remember how to reproduce it, but I had to add ReadAction#compute to fix the problem. See https://github.com/redhat-developer/lsp4ij/blob/1783e367a3190af823c6f8a5766881133fef3d67/src/main/java/com/redhat/devtools/lsp4ij/usages/LSPUsageSearcher.java#L62

angelozerr:
@SCWells72 instead of having ReadAction run/compute in several locations, perhaps it should better to have just one ReadAction, no?

SCWells72:
That's how I originally had it. However, read and write actions do take out locks, and it's best to limit the scope of those locks if possible. That's why I've added these in a more fine-grained manner.

SCWells72:
Going to open a new PR for this.

PR #712, opened 2024-12-17
fix: Addressed issues with LSP Find Usages implementation (issue 708)

SCWells72:
This is a fresh version of #711. See that PR for conversation history.


angelozerr:
LGTM, thanks @SCWells72 !

PR #716, opened 2024-12-20
feat: `codeBlockProvider` improvements

SCWells72:
Opening this as a draft PR initially to drive discussion, but I'm hoping it will be promoted to a working PR pretty quickly.

This work was driven by some inconsistent behavior and shortcomings I saw with the LSP formatting range-based code block provider, and I figured that by also using the information from the LSP selection range capability, it should be possible to address most if not all of those issues. In practice, that does actually seem to be the case.

This set of changes also includes an update to derive actual language-specific brace pairs from TextMate (`textmate` language) or the IDE's abstract file type (`TEXT` language), but due to a breaking change in the TextMate API in the oldest supported IDE versions, the TextMate aspect is currently commented out. Once support for those oldest versions (the breaking change was made in September, 2023 and it stable in all subsequent IDE versions) has been removed, TextMate support can be restored. Alternatively I could use reflection to support both until there's no longer a need for the older API.

Here's a demo of what these changes enable:

LSP4IJ_Code_Block_Provider_Demo

Specifically the IDE should be able to infer a much more authoritative and accurate set of code blocks based on the current caret position by using selection range information first and folding range information if not found via selection ranges.

The **Move Caret to Code Block Start** and **Move Caret to Code Block End** actions should take you to the correct location based on the current caret. I've enabled the IDE's Presentation Assistant in the demo above so that the OS-specific keybindings for these actions are easily found.

I'll add other details alongside the PR changes themselves.


src/main/java/com/redhat/devtools/lsp4ij/features/codeBlockProvider/LSPCodeBlockProvider.java
```diff
@@ -0,0 +1,201 @@
...
+import org.eclipse.lsp4j.SelectionRange;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.*;
+
+/**
+ * Code block provider that uses information from {@link LSPSelectionRangeSupport} and {@link LSPFoldingRangeBuilder}.
+ */
+public class LSPCodeBlockProvider implements CodeBlockProvider {
```
SCWells72, line 32: Now that this is no longer strictly tied to the folding range feature -- which it still uses, but alongside the selection range feature now -- I moved it into its own package. Unfortunately there are enough other changes that Git doesn't see it as a move so there's not a good before/after diff here.

src/main/java/com/redhat/devtools/lsp4ij/features/codeBlockProvider/LSPCodeBlockProvider.java
```diff
@@ -0,0 +1,201 @@
...
+            openBraceChar = LSPCodeBlockUtils.getCodeBlockStartChar(file, closeBraceChar);
+        } else if ((offset < (documentLength - 1)) && LSPCodeBlockUtils.isCodeBlockEndChar(file, documentChars.charAt(offset + 1))) {
+            closeBraceOffset = offset + 1;
+            closeBraceChar = documentChars.charAt(offset + 1);
+            openBraceChar = LSPCodeBlockUtils.getCodeBlockStartChar(file, closeBraceChar);
+        }
+
+        // Try to find it first using the selection ranges which tend to be more accurate; we must use the effective
+        // offset for selection ranges to act as if we're in the adjusted braced block
+        TextRange codeBlockRange = getUsingSelectionRanges(
```
SCWells72, line 72: Selection ranges from the current offset generally provide more detailed code block information, so try finding the closest containing block using that feature first.

src/main/java/com/redhat/devtools/lsp4ij/features/codeBlockProvider/LSPCodeBlockProvider.java
```diff
@@ -0,0 +1,201 @@
...
+                openBraceOffset,
+                closeBraceChar,
+                closeBraceOffset
+        );
+        if (codeBlockRange != null) {
+            return codeBlockRange;
+        }
+
+        // Failing that, try to find it using the folding ranges
+        return getUsingFoldingRanges(
```
SCWells72, line 86: If we couldn't find a good code block from the selection ranges, try to find one using the folding ranges.

src/main/java/com/redhat/devtools/lsp4ij/features/codeBlockProvider/LSPCodeBlockProvider.java
```diff
@@ -0,0 +1,201 @@
...
+            for (TextRange textRange : textRanges) {
+                if (textRange.getLength() > 1) {
+                    int startOffset = textRange.getStartOffset();
+                    int endOffset = textRange.getEndOffset();
+
+                    char startChar = documentChars.charAt(startOffset);
+                    char endChar = documentChars.charAt(endOffset - 1);
+
+                    // If aligned on an open brace and this ends with the expected close brace, use it
+                    if ((startOffset == openBraceOffset)) {
```
SCWells72, line 129: The first two conditions here are for when the caret starts exactly on an open or close brace(/bracket/paren) and the evaluated range begins/ends on that offset and ends/begins with the appropriate paired brace. Basically these are an exact brace match.

src/main/java/com/redhat/devtools/lsp4ij/features/codeBlockProvider/LSPCodeBlockProvider.java
```diff
@@ -0,0 +1,201 @@
...
+                    // If aligned on a close brace and this starts with the expected open brace, use it
+                    else if (((endOffset - 1) == closeBraceOffset)) {
+                        if ((openBraceChar != null) && (openBraceChar == startChar)) {
+                            return textRange;
+                        } else {
+                            return null;
+                        }
+                    }
+                    // Otherwise see if it starts and ends with a known brace pair and we'll find the "closest" below
+                    else if ((openBraceOffset == -1) && (closeBraceOffset == -1) && LSPCodeBlockUtils.isCodeBlockStartChar(file, startChar)) {
```
SCWells72, line 145: Otherwise we see whether or not the selection range starts and ends with a supported brace pair and, if so, we add it to the list of containing code blocks.

src/main/java/com/redhat/devtools/lsp4ij/features/codeBlockProvider/LSPCodeBlockProvider.java
```diff
@@ -0,0 +1,201 @@
...
+                        Character pairedCloseBraceChar = LSPCodeBlockUtils.getCodeBlockEndChar(file, startChar);
+                        if ((pairedCloseBraceChar != null) && (pairedCloseBraceChar == endChar)) {
+                            containingTextRanges.add(textRange);
+                        }
+                    }
+                }
+            }
+
+            // Return the closest (i.e., smallest) containing text range
+            if (!ContainerUtil.isEmpty(containingTextRanges)) {
```
SCWells72, line 155: And then we return the smallest containing code block if at least one was found.

src/main/java/com/redhat/devtools/lsp4ij/features/codeBlockProvider/LSPCodeBlockProvider.java
```diff
@@ -0,0 +1,201 @@
...
+        return null;
+    }
+
+    @Nullable
+    private static TextRange getUsingFoldingRanges(@NotNull PsiFile file,
+                                                   @NotNull Document document,
+                                                   int offset,
+                                                   @Nullable Character openBraceChar,
+                                                   @Nullable Character closeBraceChar) {
+        List<FoldingRange> foldingRanges = LSPFoldingRangeBuilder.getFoldingRanges(file);
```
SCWells72, line 170: This is effectively the same logic that was used previously to derive code blocks from folding ranges.

src/main/java/com/redhat/devtools/lsp4ij/features/codeBlockProvider/LSPCodeBlockUtils.java
```diff
@@ -0,0 +1,141 @@
...
+import org.jetbrains.plugins.textmate.psi.TextMateFile;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Utilities for deriving information about code blocks.
+ */
+@ApiStatus.Internal
+public final class LSPCodeBlockUtils {
```
SCWells72, line 34: I derived another utility class specifically for code block-related utility methods such as `isCodeBlockStart/EndChar()` and `getCodeBlockStart/EndChar()`. This is also what tries to use the file type-specific information to derive supported brace pairs. It works for both TextMate files and plain text files backed by `AbstractFileType` information, but as stated a few times already, the TextMate aspect is currently commented out due to a breaking API change across the supported IDE versions.

src/main/java/com/redhat/devtools/lsp4ij/features/codeBlockProvider/LSPCodeBlockUtils.java
```diff
@@ -0,0 +1,141 @@
...
+        return codeBlockStartChar != null ? getBracePairsFwd(file).get(codeBlockStartChar) : null;
+    }
+
+    @NotNull
+    private static Map<Character, Character> getBracePairsFwd(@NotNull PsiFile file) {
+        return CachedValuesManager.getCachedValue(file, () -> {
+            Map<Character, Character> bracePairs = file instanceof TextMateFile ? LSPIJTextMateUtils.getBracePairs(file) :
+                    file.getFileType() instanceof AbstractFileType ? getAbstractFileTypeBracePairs(file) :
+                            null;
+            if (bracePairs == null) {
```
SCWells72, line 109: Degrade gracefully to the previous behavior where we assumed that the brace pairs were the default curly braces, square brackets, and parens.

src/main/java/com/redhat/devtools/lsp4ij/features/codeBlockProvider/LSPCodeBlockUtils.java
```diff
@@ -0,0 +1,141 @@
...
+     */
+    @Nullable
+    @ApiStatus.Internal
+    public static Character getCodeBlockEndChar(@NotNull PsiFile file, @Nullable Character codeBlockStartChar) {
+        return codeBlockStartChar != null ? getBracePairsFwd(file).get(codeBlockStartChar) : null;
+    }
+
+    @NotNull
+    private static Map<Character, Character> getBracePairsFwd(@NotNull PsiFile file) {
+        return CachedValuesManager.getCachedValue(file, () -> {
```
SCWells72, line 105: I've confirmed that the caching here is properly evicted if cached files' file types are changed, e.g., from TextMate to plain text or vice-versa.

src/main/java/com/redhat/devtools/lsp4ij/features/codeBlockProvider/LSPCodeBlockUtils.java
```diff
@@ -0,0 +1,141 @@
...
+        });
+    }
+
+    @NotNull
+    private static Map<Character, Character> getBracePairsBwd(@NotNull PsiFile file) {
+        return CachedValuesManager.getCachedValue(file, () -> Result.create(ContainerUtil.reverseMap(getBracePairsFwd(file)), file));
+    }
+
+    @Nullable
+    private static Map<Character, Character> getAbstractFileTypeBracePairs(@NotNull PsiFile file) {
```
SCWells72, line 122: This is how supported brace pairs are derived for plain text files backed by `AbstractFileType`s.

src/main/java/com/redhat/devtools/lsp4ij/features/foldingRange/LSPCodeBlockProvider.java
SCWells72, line 1: This was moved to the `codeBlockProvider` package and was already described above.

src/main/java/com/redhat/devtools/lsp4ij/features/foldingRange/LSPFoldingRangeBuilder.java
```diff
@@ -49,20 +51,11 @@ public class LSPFoldingRangeBuilder extends CustomFoldingBuilder {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(LSPFoldingRangeBuilder.class);
 
-    // NOTE: JetBrains has maintained a long assumption that these are the primary structural block delimiters via
```
SCWells72, line 52: Moved into `LSPCodeBlockUtils`.

src/main/java/com/redhat/devtools/lsp4ij/features/foldingRange/LSPFoldingRangeBuilder.java
```diff
@@ -72,7 +65,7 @@ protected void buildLanguageFoldRegions(@NotNull List<FoldingDescriptor> descrip
         List<FoldingRange> foldingRanges = getFoldingRanges(file);
         if (!ContainerUtil.isEmpty(foldingRanges)) {
             for (FoldingRange foldingRange : foldingRanges) {
-                TextRange textRange = getTextRange(foldingRange, document);
+                TextRange textRange = getTextRange(foldingRange, file, document);
```
SCWells72, line 68: We need to pass the PSI file so that we can derive file type-specific brace pairs.

src/main/java/com/redhat/devtools/lsp4ij/features/foldingRange/LSPFoldingRangeBuilder.java
```diff
@@ -176,24 +175,6 @@ static TextRange getTextRange(
         return textRange;
     }
 
-    static boolean isOpenBraceChar(@Nullable Character character) {
```
SCWells72, line 179: Move into `LSPCodeBlockUtils`.

src/main/java/com/redhat/devtools/lsp4ij/features/foldingRange/LSPFoldingRangeBuilder.java
```diff
@@ -153,9 +152,9 @@ static TextRange getTextRange(
         Character startChar = start > 0 ? documentChars.charAt(start - 1) : null;
         if ((startChar != null) && ((openBraceChar == null) || (startChar == openBraceChar))) {
             // If necessary, infer the braces for this block
-            if ((openBraceChar == null) && isOpenBraceChar(startChar)) {
+            if ((openBraceChar == null) && LSPCodeBlockUtils.isCodeBlockStartChar(file, startChar)) {
```
SCWells72, line 155: These were moved but also renamed to be less "brace"-specific in their naming.

src/main/java/com/redhat/devtools/lsp4ij/features/selectionRange/LSPExtendWordSelectionHandler.java
```diff
@@ -71,89 +57,36 @@ public boolean canSelect(@NotNull PsiElement element) {
...
                  parentSelectionRange != null;
                  parentSelectionRange = parentSelectionRange.getParent()) {
-                TextRange parentSelectionTextRange = getTextRange(parentSelectionRange, document);
+                TextRange parentSelectionTextRange = LSPSelectionRangeSupport.getTextRange(parentSelectionRange, document);
                 textRanges.addAll(expandToWholeLinesWithBlanks(editorText, parentSelectionTextRange));
             }
         }
         return new ArrayList<>(textRanges);
     }
-
```
SCWells72, line 105: Moved into `LSPSelectionRangeSupport` so that it can be used by `LSPCodeBlockProvider`.

src/main/java/com/redhat/devtools/lsp4ij/features/selectionRange/LSPSelectionRangeSupport.java
```diff
@@ -33,12 +51,72 @@
...
+
     private Integer previousOffset;
 
     public LSPSelectionRangeSupport(@NotNull PsiFile file) {
         super(file);
     }
 
+    @NotNull
+    @ApiStatus.Internal
+    public static List<SelectionRange> getSelectionRanges(@NotNull PsiFile file,
```
SCWells72, line 64: This was moved from `LSPExtendWordSelectionHandler`.

src/test/java/com/redhat/devtools/lsp4ij/fixtures/LSPCodeBlockProviderFixtureTestCase.java
```diff
@@ -42,12 +43,18 @@ protected LSPCodeBlockProviderFixtureTestCase(String... fileNamePatterns) {
...
-                                   @NotNull String mockFoldingRangesJson) {
+                                   @NotNull String mockFoldingRangesJson,
+                                   @NotNull String mockSelectionRangesJson) {
         MockLanguageServer.INSTANCE.setTimeToProceedQueries(100);
+
         List<FoldingRange> mockFoldingRanges = JSONUtils.getLsp4jGson().fromJson(mockFoldingRangesJson, new TypeToken<List<FoldingRange>>() {
         }.getType());
         MockLanguageServer.INSTANCE.setFoldingRanges(mockFoldingRanges);
 
+        List<SelectionRange> mockSelectionRanges = JSONUtils.getLsp4jGson().fromJson(mockSelectionRangesJson, new TypeToken<List<SelectionRange>>() {
```
SCWells72, line 54: Use the mock selection ranges.

src/main/java/com/redhat/devtools/lsp4ij/features/codeBlockProvider/LSPCodeBlockProvider.java
```diff
@@ -0,0 +1,251 @@
...
+                        file,
+                        document,
+                        openBraceChar,
+                        closeBraceChar
+                );
+                if ((textRange != null) && (textRange.getLength() > 1)) {
+                    int startOffset = Math.max(0, textRange.getStartOffset() - 1);
+                    int endOffset = Math.min(documentLength - 1, textRange.getEndOffset());
+                    // These ranges can tend to add whitespace at the end, so trim that before looking for braces
+                    while ((endOffset > startOffset) && Character.isWhitespace(documentChars.charAt(endOffset))) {
```
SCWells72, line 206: We sometimes have to perform a pseudo-"trim" on the tail of the folding range to find the closing brace.

src/test/java/com/redhat/devtools/lsp4ij/features/foldingRange/TypeScriptCodeBlockProviderTest.java
SCWells72, line 1: This ends up being a big diff because I had to add mock selection range responses and I added considerably more test cases for more complex code blocks. I'd recommend just taking a look at the test class itself in the IDE instead of here.

src/main/java/com/redhat/devtools/lsp4ij/LSPIJTextMateUtils.java
```diff
@@ -0,0 +1,92 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.Map;
+
+/**
+ * Utilities for working with TextMate files in LSP4IJ.
+ */
+@ApiStatus.Internal
+public final class LSPIJTextMateUtils {
```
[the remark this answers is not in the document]
SCWells72: Note that I've hit this exact same issue with 2023.2 in another PR that I'll be submitting shortly. Any chance that you're thinking about dropping support for 2023.2 anytime soon? That would definitely help with some TextMate-related stuff because of these changes between 2023.2 and 2023.3+.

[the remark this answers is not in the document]
angelozerr: Ok let upgrade that.

Please do that in this PR or an another PR and please update the README also.

src/main/java/com/redhat/devtools/lsp4ij/LSPIJTextMateUtils.java
```diff
@@ -0,0 +1,92 @@
...
+
+    /**
+     * Returns the simple/single-character brace pairs for the file if it's a TextMate file.
+     *
+     * @param file the PSI file
+     * @return the simple brace pairs for the file if it's a TextMate file; otherwise null
+     */
+    @Nullable
+    @ApiStatus.Internal
+    public static Map<Character, Character> getBracePairs(@NotNull PsiFile file) {
```
angelozerr, line 39: This method should be removed, right?

src/main/java/com/redhat/devtools/lsp4ij/features/codeBlockProvider/LSPCodeBlockProvider.java
```diff
@@ -0,0 +1,251 @@
+/*******************************************************************************
+ * Copyright (c) 2024 Red Hat, Inc.
```
angelozerr, line 2: 2025

angelozerr:
@SCWells72 ypu did a lot of changes and it is hard for me to imagine usecases.

Do you think it is possible to write several tests which cover your changes?

SCWells72:
> @SCWells72 ypu did a lot of changes and it is hard for me to imagine usecases.
> 
> Do you think it is possible to write several tests which cover your changes?

I did. Look at the test changes.  They're extensive enough that they don't show up in the diff by default, but you can show them. I roughly doubled the number of test cases with specific examples of the use cases enabled by these changes.

src/main/java/com/redhat/devtools/lsp4ij/LSPIJTextMateUtils.java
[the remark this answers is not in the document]
SCWells72: > Ok let upgrade that.
> 
> Please do that in this PR or an another PR and please update the README also.

That upgrade could be a non-trivial amount of work based on whether tests, etc., need to be updated. Let's keep that separate from this work, and I'm happy to do a follow-on PR that enables this logic once LSP4IJ no longer supports 2023.2.

src/main/java/com/redhat/devtools/lsp4ij/LSPIJTextMateUtils.java
SCWells72: It's up to you. I'd hate to lose it if LSP4IJ is going to be updated to drop support for 2023.2 soon and it should be restored, and it would likely be lost through a squash merge. If you'd like it removed since it's a no-op in its current form, I can do so, but it should hopefully only be commented out for a short time until 2023.2 support is dropped. Let me know what you'd prefer.

src/main/java/com/redhat/devtools/lsp4ij/features/codeBlockProvider/LSPCodeBlockProvider.java
SCWells72: This class already existed in 2024 and was moved across packages. It should still be 2024.

src/main/java/com/redhat/devtools/lsp4ij/LSPIJTextMateUtils.java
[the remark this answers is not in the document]
SCWells72: For what it's worth, I just tried to update the build for 2023.3 instead of 2023.2 in the context of this PR, and while I can build and run the plugin, the tests fail with:

```
Execution failed for task ':instrumentCode'.
> taskdef class com.intellij.ant.InstrumentIdeaExtensions cannot be found
   using the classloader AntClassLoader[]
```

It's possible that the build needs to be updated for the Gradle 2.0 plugin SDK to target higher versions. Either way, it seems like that upgrade should be handled as its own PR.

SCWells72:
@angelozerr the PR still shows "Changes requested", but I think I've addressed all feedback. Please let me know if you're waiting for me on something here or have what you need to proceed with the next phase of the review and/or merge.

angelozerr:
@SCWells72 could you please record a demo which shows in action the fix that PR provides.

I have tried to set a breakpoint in the LSPCodeBlockProvider and try to format code, try to uses selection ranges, (with Ctrl+W) but I go never in the LSPCodeBlockProvider?

SCWells72:
@angelozerr Apologies for the confusion. This change isn't about formatting. It's about extending the capabilities of the `codeBlockProvider` EP implementation so that it can identify code blocks for the active caret more acccurately and extensively by consulting the selection ranges from the caret first, then the formatting ranges from that same location if a good code block couldn't be identified by selection range.

The best way to understand it is to look at the changes to `TypeScriptCodeBlockProviderTest` where I've added explicit test cases for identification of code blocks that weren't possible with only folding ranges:

https://github.com/redhat-developer/lsp4ij/pull/716/files#diff-5d3b7b59c0f9a335ac1124e7ecb1c29d5a7a46e1e136144c042b931638f5b589

or:

https://github.com/redhat-developer/lsp4ij/blob/8503f91fc919d362e65c9f3537d7f25c85ad601d/src/test/java/com/redhat/devtools/lsp4ij/features/foldingRange/TypeScriptCodeBlockProviderTest.java

If you want to try it within the editor, try with the following:

```typescript
export class Foo {
    bar() {
        let condition: boolean = true;
        if (condition) {
            invokePromise({foo: '', bar: 10})
                .then(() => {
                    console.log('demo');
                })
                .catch((error) => {
                    console.error(error);
                });
        }
    }
}

function invokePromise(params: {foo: string, bar: number;}): Promise<any> {
    console.log(params);
    return Promise.call(params);
}
```

and you can see that the Go To Matching Brace actions work properly for each of the brace pairs including the `.then` and `.catch` callback functions now whereas they didn't with just the formatting ranges-based implementation.

Once this is merged, there are follow-on PRs that add support for improved formatting including client-side on-type formatting using this improved code block information, but this one doesn't have anything to do with formatting.

Let me know if that doesn't clarify the goals/behavior of this PR.

angelozerr:
Thanks @SCWells72 !

PR #737, opened 2025-01-10
feat: Allow disabling of editor templates when accepting method/function invocation completions with a single argument (fixes #718)

### LINKED ISSUE #718: When code completions with a single tab stop are inserted, it should be possible to disable the resulting template

Image

SCWells72:
As stated/shown in #718, the editor template that is currently started when accepting a code completion that is a method/function invocation with a single argument is quite disruptive to the standard editor flow. This PR adds a new client configuration option, `completion.useTemplateForSingleArgument`, that, when disabled (enabled by default for backward-compatibility), instead just places the caret into the invocation arguments so that the user can populate them as they would in any of JetBrains' natively-supported language editors. I've also updated all existing language server definition templates to disable editor templates in this case so that the "default" behavior for newly-created language server definitions is the smoother/more natural editor flow, but existing language server definitions will not see a change in behavior.


SCWells72:
This is pretty dirty from all the merging/rebasing. I'll create a new branch/PR with just the changes.

PR #738, opened 2025-01-10
feat: Allow disabling of editor templates when accepting method/function invocation completions (fixes #718)

### LINKED ISSUE #718: When code completions with a single tab stop are inserted, it should be possible to disable the resulting template

SCWells72:
As stated/shown in #718, the editor template that is currently started when accepting a code completion that is a method/function invocation ~~with a single argument~~ is quite disruptive to the standard editor flow. ~~This PR adds a new client configuration option, `completion.useTemplateForSingleArgument`, that, when disabled (enabled by default for backward-compatibility), instead just places the caret into the invocation arguments so that the user can populate them as they would in any of JetBrains' natively-supported language editors. I've also updated all existing language server definition templates to disable editor templates in this case so that the "default" behavior for newly-created language server definitions is the smoother/more natural editor flow, but existing language server definitions will not see a change in behavior.~~

**UPDATE:** This has now been generalized to allow disabling editor templates for _any_ invocation-only snippet. An invocation-only snippet is one where all non-end (`$0`) variables are found as a comma-/whitespace-delimited list in invocation parentheses. If other invocation patterns are found in LSP languages, recognition support can be added for them as well.

The configuration option is now `completion.useTemplateForInvocationOnlySnippet` to reflect this more general behavior. As before, it is disabled by default in code but enabled by default in the bundled language server templates.

This also properly handles the situation of a single no-value variable snippet by forgoing an editor template and moving to the correct caret position for that variable with no need for any client config change.


src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPCompletionProposal.java
```diff
@@ -155,12 +167,23 @@ private void updateCompletionItemFromResolved() {
         this.item = resolved;
     }
 
+    @Contract("null -> false")
```
SCWells72, line 170: I added `@Contract` annotations here because the IDE's nullability inspection was having trouble determining the relationships between the params and the results and was raising false inspection positives.

src/main/java/com/redhat/devtools/lsp4ij/server/definition/ClientConfigurableLanguageServerDefinition.java
```diff
@@ -0,0 +1,31 @@
...
+
+package com.redhat.devtools.lsp4ij.server.definition;
+
+import com.redhat.devtools.lsp4ij.server.definition.launching.ClientConfigurationSettings;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Common interface for language server definitions that support client configuration.
+ */
+public interface ClientConfigurableLanguageServerDefinition {
```
SCWells72, line 23: This is the part that probably warrants the most discussion, and it also exists in the upcoming PR for client-side on-type formatting. Basically we need to be able to change client config settings from unit tests, but prior to this change, only `UserDefinedLanguageServerDefinition` has had access to client configuration. However, we need `MockLanguageServerDefinition` to have client configuration as well, and ideally without having to push that notion up into `LanguageServerDefinition`. As a result, I added this common interface for `LanguageServerDefinition` implementations that are client-configurable, and I implemented that interface on both `UserDefinedLanguageServerDefinition` and `MockLanguageServerDefinition`. There will obviously be a few other related changes below, but hopefully that sets the context. And I'm certainly open to other options for how client config could be exposed in unit tests, but this certainly works nicely and without any major ripple-effect.

src/main/java/com/redhat/devtools/lsp4ij/server/definition/launching/UserDefinedCompletionFeature.java
```diff
@@ -24,8 +25,15 @@ public class UserDefinedCompletionFeature extends LSPCompletionFeature {
 
     @Override
     public boolean useContextAwareSorting(@NotNull PsiFile file) {
-        UserDefinedLanguageServerDefinition serverDefinition = (UserDefinedLanguageServerDefinition) getClientFeatures().getServerDefinition();
+        ClientConfigurableLanguageServerDefinition serverDefinition = (ClientConfigurableLanguageServerDefinition) getClientFeatures().getServerDefinition();
```
SCWells72, line 28: Previously it was deemed safe to blindly cast to `UserDefinedLanguageServerDefinition` so I'm doing the same here with `ClientConfigurableLanguageServerDefinition`. However, I would feel better checking the type before casting with this change. Thoughts? If you agree, I'll update all such usages to be more defensive.

angelozerr:
@SCWells72 I wonder if we should have a settings for that? I have tried the Java editor of JetBrains and it works like this.

src/main/java/com/redhat/devtools/lsp4ij/server/definition/launching/UserDefinedLanguageServerDefinition.java
```diff
@@ -34,7 +35,7 @@
  * {@link com.redhat.devtools.lsp4ij.server.definition.LanguageServerDefinition} implementation to start a
  * language server with a process command defined by the user.
  */
-public class UserDefinedLanguageServerDefinition extends LanguageServerDefinition {
+public class UserDefinedLanguageServerDefinition extends LanguageServerDefinition implements ClientConfigurableLanguageServerDefinition {
```
SCWells72, line 38: Here's where it has been updated to implement the client-configurable interface.

src/main/resources/templates/clojure-lsp/clientSettings.json
```diff
@@ -1,7 +1,8 @@
 {
   "caseSensitive": true,
   "completion": {
-    "useContextAwareSorting": true
+    "useContextAwareSorting": true,
+    "useTemplateForSingleArgument": false
```
SCWells72, line 5: Even though editor templates are enabled by default for backward-compatibility, I updated all of the existing language server definition templates to disable editor templates for single-arg invocations because they're so disruptive to the editor typing experience. This is the same approach we used for `useContextAwareSorting` to avoid changing behavior for existing language server definitions while making newly-created ones use the preferred behavior.

src/test/java/com/redhat/devtools/lsp4ij/fixtures/LSPCompletionClientConfigFixtureTestCase.java
```diff
@@ -0,0 +1,153 @@
...
+                                              @NotNull String fileBodyAfter,
+                                              @NotNull String mockTextCompletionJson,
+                                              @NotNull String mockCompletionItemResolveJson,
+                                              @Nullable Consumer<ClientConfigurationSettings> clientConfigCustomizer) {
+        MockLanguageServer.INSTANCE.setTimeToProceedQueries(200);
+
+        CompletionList mockCompletionList = JSONUtils.getLsp4jGson().fromJson(mockTextCompletionJson, CompletionList.class);
+        MockLanguageServer.INSTANCE.setCompletionList(mockCompletionList);
+
+        CompletionItem mockCompletionItem = JSONUtils.getLsp4jGson().fromJson(mockCompletionItemResolveJson, CompletionItem.class);
```
SCWells72, line 80: I had to add the ability to specify the mock resolved completion item.

src/test/java/com/redhat/devtools/lsp4ij/fixtures/LSPCompletionClientConfigFixtureTestCase.java
```diff
@@ -0,0 +1,153 @@
...
+            fail(e.getMessage());
+        }
+
+        // Configure the language server's completion client configuration
+        LanguageServerItem languageServer = ContainerUtil.getFirstItem(languageServers);
+        assertNotNull(languageServer);
+
+        // Use a configurable completion feature
+        LSPClientFeatures clientFeatures = languageServer.getClientFeatures();
+        clientFeatures.setCompletionFeature(new UserDefinedCompletionFeature());
```
SCWells72, line 104: Make sure that the mock server is using a configurable completion feature.

src/test/java/com/redhat/devtools/lsp4ij/fixtures/LSPCompletionClientConfigFixtureTestCase.java
```diff
@@ -0,0 +1,153 @@
...
+
+        // Enable completion item resolution
+        languageServer.getServerCapabilities().setCompletionProvider(new CompletionOptions(true, null));
```
SCWells72, line 107: By default the mock completion provider's options don't enable resolution, so must enable that for this test.

src/test/java/com/redhat/devtools/lsp4ij/fixtures/LSPCompletionClientConfigFixtureTestCase.java
```diff
@@ -0,0 +1,153 @@
...
+
+        // Update client configuration as required for this test scenario
+        if (clientConfigCustomizer != null) {
```
SCWells72, line 110: And finally, get the client configuration from the server and apply any config changes for this test scenario.

src/test/java/com/redhat/devtools/lsp4ij/fixtures/LSPCompletionClientConfigFixtureTestCase.java
```diff
@@ -0,0 +1,153 @@
...
+            LanguageServerDefinition languageServerDefinition = languageServer.getServerDefinition();
+            assertInstanceOf(languageServerDefinition, ClientConfigurableLanguageServerDefinition.class);
+            ClientConfigurableLanguageServerDefinition configurableLanguageServerDefinition = (ClientConfigurableLanguageServerDefinition) languageServerDefinition;
+            ClientConfigurationSettings clientConfiguration = configurableLanguageServerDefinition.getLanguageServerClientConfiguration();
+            assertNotNull(clientConfiguration);
+            clientConfigCustomizer.accept(clientConfiguration);
+        }
+
+        // Move to the offset at which completion should be triggered
```
SCWells72, line 119: Hopefully the rest is pretty self-explanatory.

src/test/java/com/redhat/devtools/lsp4ij/mock/MockLanguageServer.java
```diff
@@ -162,6 +162,10 @@ public void setCompletionList(CompletionList completionList) {
 		this.textDocumentService.setMockCompletionList(completionList);
 	}
 
+	public void setCompletionItem(CompletionItem completionItem) {
```
SCWells72, line 165: Like I said above, need to be able to set the mock resolved completion item.

src/test/java/com/redhat/devtools/lsp4ij/mock/MockLanguageServerDefinition.java
```diff
@@ -12,16 +12,21 @@
...
 import com.redhat.devtools.lsp4ij.server.definition.LanguageServerDefinition;
+import com.redhat.devtools.lsp4ij.server.definition.launching.ClientConfigurationSettings;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 /**
  * {@link LanguageServerDefinition} implementation to register the mock language server.
  */
-public class MockLanguageServerDefinition extends LanguageServerDefinition {
+public class MockLanguageServerDefinition extends LanguageServerDefinition implements ClientConfigurableLanguageServerDefinition {
```
SCWells72, line 24: The mock language server is now client-configurable as well.

src/test/java/com/redhat/devtools/lsp4ij/mock/MockLanguageServerDefinition.java
```diff
@@ -12,16 +12,21 @@
...
 
     private static final String SERVER_ID = "mock-server-id";
 
+    private final ClientConfigurationSettings clientConfigurationSettings = new ClientConfigurationSettings();
```
SCWells72, line 28: And it just starts with a default client configuration.

src/test/java/com/redhat/devtools/lsp4ij/mock/MockTextDocumentService.java
```diff
@@ -35,6 +35,7 @@
 public class MockTextDocumentService implements TextDocumentService {
 
     private CompletionList mockCompletionList;
+    private CompletionItem mockCompletionItem;
```
SCWells72, line 38: Other stuff required to be able to mock the resolved completion item.

SCWells72:
@angelozerr this builds on the foundation established by the previous PR for completion sorting improvements, and it also establishes the pattern I've used for testing client configuration settings in a mock language server. Let me know if you have an questions/concerns/thoughts about what problems this is addressing and/or how it's addressing them.

SCWells72:
> @SCWells72 I wonder if we should have a settings for that? I have tried the Java editor of JetBrains and it works like this.

@angelozerr, I'm not sure that I understand this question. I've included client configuration settings for the change in this PR already. Or do you mean settings for something else?

angelozerr:
> > @SCWells72 I wonder if we should have a settings for that? I have tried the Java editor of JetBrains and it works like this.
> 
> @angelozerr, I'm not sure that I understand this question. I've included client configuration settings for the change in this PR already. Or do you mean settings for something else?

I mean that for me the current behavior is a bug and useTemplateForSingleArgument should be removed. What is the benefit to have this behaviour?

SCWells72:
> > > @SCWells72 I wonder if we should have a settings for that? I have tried the Java editor of JetBrains and it works like this.
> > 
> > 
> > @angelozerr, I'm not sure that I understand this question. I've included client configuration settings for the change in this PR already. Or do you mean settings for something else?
> 
> I mean that for me the current behavior is a bug and useTemplateForSingleArgument should be removed. What is the benefit to have this behaviour?

@angelozerr, oh, I see. Yes, I agree that the current behavior is a bug. If you're fine with me just removing the client configuration from this PR and making that the behavior, I'm all for that. Let me know if that's what you're recommending.

SCWells72:
@angelozerr switching back to draft because I have a good idea on how this can/should be generalized. I'll add those changes and then let you know when it's switched back to active state.

src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPCompletionProposal.java
```diff
@@ -65,6 +73,12 @@ public class LSPCompletionProposal extends LookupElement implements Pointer<LSPC
 
     private static final Logger LOGGER = LoggerFactory.getLogger(LSPCompletionProposal.class);
 
+    private static final Pattern TEMPLATE_VARIABLE_PATTERN = Pattern.compile("\\$(?:\\{\\d:.+?}|\\d+)");
```
SCWells72, line 76: As stated in the comment, these patterns should work well for languages where invocations take the form of parenthesized, comma-delimited argument lists. If supported languages with other invocation syntaxes are found, it should be pretty simple to extend the pattern to support them as well, but until such concrete examples are found, this should be a lightweight way to help determine that a code snippet is explicitly for an invocation argument list.

src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPCompletionProposal.java
```diff
@@ -100,14 +114,16 @@ public void handleInsert(@NotNull InsertionContext context) {
         updateCompletionItemFromResolved();
         Template template = null;
         if (item.getInsertTextFormat() == InsertTextFormat.Snippet) {
+            // Adjust the snippet content as appropriate based on client config
+            String snippetContent = adjustSnippetContent(getInsertText());
```
SCWells72, line 118: Now we adjust the LSP-provided snippet before parsing it into a template if appropriate based on client config and the snippet contents.

src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPCompletionProposal.java
```diff
@@ -100,14 +114,16 @@ public void handleInsert(@NotNull InsertionContext context) {
...
+
             // Insert text has snippet syntax, ex : ${1:name}
-            String snippetContent = getInsertText();
             // Get the indentation settings
             LspSnippetIndentOptions indentOptions = CompletionProposalTools.createLspIndentOptions(snippetContent, file);
             // Load the insert text to build:
             // - an IJ Template instance which will take care of replacement of placeholders
             // - the insert text without placeholders
-            template = SnippetTemplateFactory.createTemplate(snippetContent, context.getProject(), name -> getVariableValue(name), indentOptions);
+            template = SnippetTemplateFactory.createTemplate(snippetContent, context.getProject(), this::getVariableValue, indentOptions);
```
SCWells72, line 126: Just changed to a method reference. Same behavior.

src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPCompletionProposal.java
```diff
@@ -116,7 +132,12 @@ public void handleInsert(@NotNull InsertionContext context) {
         // Apply all text edits
         apply(context.getDocument(), context.getOffset(CompletionInitializationContext.SELECTION_END_OFFSET));
 
-        if (shouldStartTemplate(template)) {
+        // Just move to the first tab stop if there's a single invocation argument with no default value
+        if (shouldMoveToFirstTabStop(template)) {
```
SCWells72, line 136: Now we only move to the first tab stop for a template with a single non-end variable that has no value. That happens independent of client config because it doesn't make sense to start an editor template for a single tab stop with no default value. Instead just place the caret in that same location and let the user type what they need.

src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPCompletionProposal.java
```diff
@@ -155,12 +176,57 @@ private void updateCompletionItemFromResolved() {
         this.item = resolved;
     }
 
+    @NotNull
+    private String adjustSnippetContent(@NotNull String snippetContent) {
```
SCWells72, line 180: Hopefully this method is pretty self-explanatory, but basically if client config says not to use editor templates for invocation arg-only completion snippets, see if the snippet contains what appears to be an invocation arg list with template variables, and if so **and** if there are no non-end variables outside of that invocation arg list, just update the snippet to be `<beforeArgList>($0)<afterArgList>` with any other `$0` removed. That way when the snippet is parsed into a template, it will just result in the caret being moved into the invocation arg parens.

src/main/resources/templates/clojure-lsp/clientSettings.json
```diff
@@ -1,7 +1,8 @@
 {
   "caseSensitive": true,
   "completion": {
-    "useContextAwareSorting": true
+    "useContextAwareSorting": true,
+    "useTemplateForInvocationOnlyCodeSnippet": false
```
SCWells72, line 5: As before, disabling editor templates for invocation-only code snippets in all bundled templates so that newly-created language server definitions will have this (IMO preferred) behavior by defafult.

SCWells72:
@angelozerr this PR has been restored to an active state. Please read the comments I've added about how it has been generalized to support disabling of editor templates for _all_ invocation-style completion snippets instead of just the special case of a single invocation argument.

src/test/java/com/redhat/devtools/lsp4ij/features/completion/CssCompletionClientConfigTest.java
```diff
@@ -0,0 +1,445 @@
...
+ ******************************************************************************/
+
+package com.redhat.devtools.lsp4ij.features.completion;
+
+import com.redhat.devtools.lsp4ij.fixtures.LSPCompletionClientConfigFixtureTestCase;
+
+/**
+ * Tests code completion client configuration settings using a mock CSS language server.
+ */
+public class CssCompletionClientConfigTest extends LSPCompletionClientConfigFixtureTestCase {
```
SCWells72, line 19: The CSS test exercises whitespace-delimited argument list snippets as well as some complex snippets with variable values that are enumerated lists/choices.

PR #744, opened 2025-01-14
feat: Allow disabling of editor templates when accepting method/function invocation completions (fixes #718)

### LINKED ISSUE #718: When code completions with a single tab stop are inserted, it should be possible to disable the resulting template

SCWells72:
This is a fresh branch-based PR as a replacmeent for #738.

As stated/shown in #718, the editor template that is currently started when accepting a code completion that is a method/function invocation ~~with a single argument~~ is quite disruptive to the standard editor flow. ~~This PR adds a new client configuration option, `completion.useTemplateForSingleArgument`, that, when disabled (enabled by default for backward-compatibility), instead just places the caret into the invocation arguments so that the user can populate them as they would in any of JetBrains' natively-supported language editors. I've also updated all existing language server definition templates to disable editor templates in this case so that the "default" behavior for newly-created language server definitions is the smoother/more natural editor flow, but existing language server definitions will not see a change in behavior.~~

**UPDATE:** This has now been generalized to allow disabling editor templates for _any_ invocation-only snippet. An invocation-only snippet is one where all non-end (`$0`) variables are found as a comma-/whitespace-delimited list in invocation parentheses. If other invocation patterns are found in LSP languages, recognition support can be added for them as well.

The configuration option is now `completion.useTemplateForInvocationOnlySnippet` to reflect this more general behavior. As before, it is disabled by default in code but enabled by default in the bundled language server templates.

This also properly handles the situation of a single no-value variable snippet by forgoing an editor template and moving to the correct caret position for that variable with no need for any client config change.

### DIFF
--- a/src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPCompletionProposal.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPCompletionProposal.java
@@ -16,8 +16,11 @@ import com.intellij.codeInsight.completion.CompletionInitializationContext;
 import com.intellij.codeInsight.completion.InsertionContext;
 import com.intellij.codeInsight.completion.LookupElementListPresenter;
 import com.intellij.codeInsight.lookup.*;
+import com.intellij.codeInsight.template.Expression;
+import com.intellij.codeInsight.template.Result;
 import com.intellij.codeInsight.template.Template;
 import com.intellij.codeInsight.template.TemplateManager;
+import com.intellij.codeInsight.template.impl.Variable;
 import com.intellij.model.Pointer;
 import com.intellij.model.Symbol;
 import com.intellij.openapi.editor.Document;
@@ -26,10 +29,12 @@ import com.intellij.openapi.editor.EditorModificationUtil;
 import com.intellij.openapi.progress.ProcessCanceledException;
 import com.intellij.openapi.progress.ProgressManager;
 import com.intellij.openapi.util.TextRange;
+import com.intellij.openapi.util.text.StringUtil;
 import com.intellij.platform.backend.documentation.DocumentationResult;
 import com.intellij.platform.backend.documentation.DocumentationTarget;
 import com.intellij.platform.backend.presentation.TargetPresentation;
 import com.intellij.psi.PsiFile;
+import com.intellij.util.containers.ContainerUtil;
 import com.redhat.devtools.lsp4ij.LSPIJUtils;
 import com.redhat.devtools.lsp4ij.LanguageServerItem;
 import com.redhat.devtools.lsp4ij.commands.CommandExecutor;
@@ -41,6 +46,7 @@ import com.redhat.devtools.lsp4ij.internal.StringUtils;
 import org.eclipse.lsp4j.*;
 import org.eclipse.lsp4j.jsonrpc.messages.Either;
 import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.Contract;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.slf4j.Logger;
@@ -50,6 +56,8 @@ import java.util.*;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import static com.redhat.devtools.lsp4ij.features.completion.snippet.LspSnippetVariableConstants.*;
 import static com.redhat.devtools.lsp4ij.features.documentation.LSPDocumentationHelper.convertToHtml;
@@ -64,6 +72,12 @@ import static com.redhat.devtools.lsp4ij.internal.CompletableFutures.waitUntilDo
 public class LSPCompletionProposal extends LookupElement implements Pointer<LSPCompletionProposal>, Symbol, DocumentationTarget {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(LSPCompletionProposal.class);
+
+    // These patterns should match code snippets that look like parenthesized, comma-/whitespace-delimited invocation arg lists
+    private static final Pattern SNIPPET_VARIABLE_PATTERN = Pattern.compile("\\$(?:\\{\\d[:|].+?}|\\d+)");
+    // TODO: What supported language grammars would not be supported by these pattern?
+    private static final Pattern INVOCATION_ARGS_SNIPPET_PATTERN = Pattern.compile("\\(\\s*" + SNIPPET_VARIABLE_PATTERN.pattern() + "(?:(?:,\\s*|\\s*)" + SNIPPET_VARIABLE_PATTERN.pattern() + ")*\\s*\\)");
+    private static final String END_VARIABLE = "$0";
 
     private CompletionItem item; // can be replaced with resolved
     private final PsiFile file;
@@ -100,14 +114,16 @@ public class LSPCompletionProposal extends LookupElement implements Pointer<LSPC
         updateCompletionItemFromResolved();
         Template template = null;
         if (item.getInsertTextFormat() == InsertTextFormat.Snippet) {
+            // Adjust the snippet content as appropriate based on client config
+            String snippetContent = adjustSnippetContent(getInsertText());
+
             // Insert text has snippet syntax, ex : ${1:name}
-            String snippetContent = getInsertText();
             // Get the indentation settings
             LspSnippetIndentOptions indentOptions = CompletionProposalTools.createLspIndentOptions(snippetContent, file);
             // Load the insert text to build:
             // - an IJ Template instance which will take care of replacement of placeholders
             // - the insert text without placeholders
-            template = SnippetTemplateFactory.createTemplate(snippetContent, context.getProject(), name -> getVariableValue(name), indentOptions);
+            template = SnippetTemplateFactory.createTemplate(snippetContent, context.getProject(), this::getVariableValue, indentOptions);
             // Update the TextEdit with the content snippet content without placeholders
             // ex : ${1:name} --> name
             updateInsertTextForTemplateProcessing(template.getTemplateText());
@@ -116,7 +132,12 @@ public class LSPCompletionProposal extends LookupElement implements Pointer<LSPC
         // Apply all text edits
         apply(context.getDocument(), context.getOffset(CompletionInitializationContext.SELECTION_END_OFFSET));
 
-        if (shouldStartTemplate(template)) {
+        // Just move to the first tab stop if there's a single invocation argument with no default value
+        if (shouldMoveToFirstTabStop(template)) {
+            int segmentOffset = template.getSegmentOffset(0);
+            EditorModificationUtil.moveCaretRelatively(editor, -template.getTemplateText().length());
+            EditorModificationUtil.moveCaretRelatively(editor, segmentOffset);
+        } else if (shouldStartTemplate(template)) {
             // LSP completion with snippet syntax, activate the inline template
             context.setAddCompletionChar(false);
             EditorModificationUtil.moveCaretRelatively(editor, -template.getTemplateText().length());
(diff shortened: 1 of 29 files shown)


src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPCompletionProposal.java
```diff
@@ -100,14 +114,16 @@ public void handleInsert(@NotNull InsertionContext context) {
         updateCompletionItemFromResolved();
         Template template = null;
         if (item.getInsertTextFormat() == InsertTextFormat.Snippet) {
+            // Adjust the snippet content as appropriate based on client config
+            String snippetContent = adjustSnippetContent(getInsertText());
```
SCWells72, line 118: Now we adjust the LSP-provided snippet before parsing it into a template if appropriate based on client config and the snippet contents.

src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPCompletionProposal.java
```diff
@@ -116,7 +132,12 @@ public void handleInsert(@NotNull InsertionContext context) {
         // Apply all text edits
         apply(context.getDocument(), context.getOffset(CompletionInitializationContext.SELECTION_END_OFFSET));
 
-        if (shouldStartTemplate(template)) {
+        // Just move to the first tab stop if there's a single invocation argument with no default value
+        if (shouldMoveToFirstTabStop(template)) {
```
SCWells72, line 136: Now we only move to the first tab stop for a template with a single non-end variable that has no value. That happens independent of client config because it doesn't make sense to start an editor template for a single tab stop with no default value. Instead just place the caret in that same location and let the user type what they need.

src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPCompletionProposal.java
```diff
@@ -155,12 +176,57 @@ private void updateCompletionItemFromResolved() {
         this.item = resolved;
     }
 
+    @NotNull
+    private String adjustSnippetContent(@NotNull String snippetContent) {
```
SCWells72, line 180: Hopefully this method is pretty self-explanatory, but basically if client config says not to use editor templates for invocation arg-only completion snippets, see if the snippet contains what appears to be an invocation arg list with template variables, and if so **and** if there are no non-end variables outside of that invocation arg list, just update the snippet to be `<beforeArgList>($0)<afterArgList>` with any other `$0` removed. That way when the snippet is parsed into a template, it will just result in the caret being moved into the invocation arg parens.

src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPCompletionProposal.java
```diff
@@ -155,12 +176,57 @@ private void updateCompletionItemFromResolved() {
...
+                String beforeInvocationArgs = snippetContent.substring(0, invocationArgsStartIndex + 1);
+                String afterInvocationArgs = snippetContent.substring(invocationArgsEndIndex - 1);
+                snippetContent = beforeInvocationArgs.replace(END_VARIABLE, "") + END_VARIABLE + afterInvocationArgs.replace(END_VARIABLE, "");
+            }
+        }
+
+        return snippetContent;
+    }
+
+    @Contract("null -> false")
```
SCWells72, line 212: I added `@Contract` annotations here and below because the IDE's nullability inspection was having trouble determining the relationships between the params and the results and was raising false inspection positives.

src/main/java/com/redhat/devtools/lsp4ij/server/definition/launching/UserDefinedCompletionFeature.java
```diff
@@ -24,8 +25,15 @@ public class UserDefinedCompletionFeature extends LSPCompletionFeature {
 
     @Override
     public boolean useContextAwareSorting(@NotNull PsiFile file) {
-        UserDefinedLanguageServerDefinition serverDefinition = (UserDefinedLanguageServerDefinition) getClientFeatures().getServerDefinition();
+        ClientConfigurableLanguageServerDefinition serverDefinition = (ClientConfigurableLanguageServerDefinition) getClientFeatures().getServerDefinition();
```
SCWells72, line 28: Previously it was deemed safe to blindly cast to `UserDefinedLanguageServerDefinition` so I'm doing the same here with `ClientConfigurableLanguageServerDefinition`. However, I would feel better checking the type before casting with this change. Thoughts? If you agree, I'll update all such usages to be more defensive.

src/main/resources/templates/clojure-lsp/clientSettings.json
```diff
@@ -1,7 +1,8 @@
 {
   "caseSensitive": true,
   "completion": {
-    "useContextAwareSorting": true
+    "useContextAwareSorting": true,
+    "useTemplateForInvocationOnlySnippet": false
```
SCWells72, line 5: I updated all of the templates to disable editor templates for invocation-only snippets similar to how I enabled context-aware sorting in the templates.

src/test/java/com/redhat/devtools/lsp4ij/features/completion/CssCompletionClientConfigTest.java
```diff
@@ -0,0 +1,445 @@
...
+ ******************************************************************************/
+
+package com.redhat.devtools.lsp4ij.features.completion;
+
+import com.redhat.devtools.lsp4ij.fixtures.LSPCompletionClientConfigFixtureTestCase;
+
+/**
+ * Tests code completion client configuration settings using a mock CSS language server.
+ */
+public class CssCompletionClientConfigTest extends LSPCompletionClientConfigFixtureTestCase {
```
SCWells72, line 19: The CSS test exercises whitespace-delimited argument list snippets as well as some complex snippets with variable values that are enumerated lists/choices.

src/test/java/com/redhat/devtools/lsp4ij/fixtures/LSPCompletionClientConfigFixtureTestCase.java
```diff
@@ -0,0 +1,155 @@
...
+                                              @NotNull String fileBodyAfter,
+                                              @NotNull String mockTextCompletionJson,
+                                              @NotNull String mockCompletionItemResolveJson,
+                                              @Nullable Consumer<ClientConfigurationSettings> clientConfigCustomizer) {
+        MockLanguageServer.INSTANCE.setTimeToProceedQueries(200);
+
+        CompletionList mockCompletionList = JSONUtils.getLsp4jGson().fromJson(mockTextCompletionJson, CompletionList.class);
+        MockLanguageServer.INSTANCE.setCompletionList(mockCompletionList);
+
+        CompletionItem mockCompletionItem = JSONUtils.getLsp4jGson().fromJson(mockCompletionItemResolveJson, CompletionItem.class);
```
SCWells72, line 79: I had to add the ability to specify the mock resolved completion item.

src/test/java/com/redhat/devtools/lsp4ij/fixtures/LSPCompletionClientConfigFixtureTestCase.java
```diff
@@ -0,0 +1,155 @@
...
+        try {
+            ContainerUtil.addAllNotNull(languageServers, LanguageServiceAccessor.getInstance(project)
+                    .getLanguageServers(file.getVirtualFile(), null, null)
+                    .get(5000, TimeUnit.MILLISECONDS));
+        } catch (Exception e) {
+            fail(e.getMessage());
+        }
+
+        // Configure the language server's completion client configuration
+        LanguageServerItem languageServer = ContainerUtil.getFirstItem(languageServers);
```
SCWells72, line 98: Now we need to get the language server that's going to be used and set it up properly for client config testing.

src/test/java/com/redhat/devtools/lsp4ij/fixtures/LSPCompletionClientConfigFixtureTestCase.java
```diff
@@ -0,0 +1,155 @@
...
+
+        // Use a configurable completion feature
+        LSPClientFeatures clientFeatures = languageServer.getClientFeatures();
+        clientFeatures.setCompletionFeature(new UserDefinedCompletionFeature());
+
+        // Enable completion item resolution
+        languageServer.getServerCapabilities().setCompletionProvider(new CompletionOptions(true, null));
+
+        // Update client configuration as required for this test scenario
+        if (clientConfigCustomizer != null) {
```
SCWells72, line 109: And finally, get the client configuration from the server and apply any config changes for this test scenario.

src/test/java/com/redhat/devtools/lsp4ij/fixtures/LSPCompletionClientConfigFixtureTestCase.java
```diff
@@ -0,0 +1,155 @@
...
+            LanguageServerDefinition languageServerDefinition = languageServer.getServerDefinition();
+            assertInstanceOf(languageServerDefinition, ClientConfigurableLanguageServerDefinition.class);
+            ClientConfigurableLanguageServerDefinition configurableLanguageServerDefinition = (ClientConfigurableLanguageServerDefinition) languageServerDefinition;
+            ClientConfigurationSettings clientConfiguration = configurableLanguageServerDefinition.getLanguageServerClientConfiguration();
+            assertNotNull(clientConfiguration);
+            clientConfigCustomizer.accept(clientConfiguration);
+        }
+
+        // Move to the offset at which completion should be triggered
```
SCWells72, line 118: Hopefully the rest is pretty self-explanatory.

src/test/java/com/redhat/devtools/lsp4ij/mock/MockLanguageServer.java
```diff
@@ -162,6 +162,10 @@ public void setCompletionList(CompletionList completionList) {
 		this.textDocumentService.setMockCompletionList(completionList);
 	}
 
+	public void setCompletionItem(CompletionItem completionItem) {
```
SCWells72, line 165: Like I said above, need to be able to set the mock resolved completion item.

src/test/java/com/redhat/devtools/lsp4ij/mock/MockLanguageServerDefinition.java
```diff
@@ -12,16 +12,21 @@
...
 import com.redhat.devtools.lsp4ij.server.definition.LanguageServerDefinition;
+import com.redhat.devtools.lsp4ij.server.definition.launching.ClientConfigurationSettings;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 /**
  * {@link LanguageServerDefinition} implementation to register the mock language server.
  */
-public class MockLanguageServerDefinition extends LanguageServerDefinition {
+public class MockLanguageServerDefinition extends LanguageServerDefinition implements ClientConfigurableLanguageServerDefinition {
```
SCWells72, line 24: The mock language server is now client-configurable as well.

src/test/java/com/redhat/devtools/lsp4ij/mock/MockLanguageServerDefinition.java
```diff
@@ -12,16 +12,21 @@
...
 
     private static final String SERVER_ID = "mock-server-id";
 
+    private final ClientConfigurationSettings clientConfigurationSettings = new ClientConfigurationSettings();
```
SCWells72, line 28: And it just starts with a default client configuration.

src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPCompletionProposal.java
```diff
@@ -102,12 +109,22 @@ public void handleInsert(@NotNull InsertionContext context) {
...
             // Load the insert text to build:
             // - an IJ Template instance which will take care of replacement of placeholders
             // - the insert text without placeholders
-            template = SnippetTemplateFactory.createTemplate(snippetContent, context.getProject(), name -> getVariableValue(name), indentOptions);
+            template = SnippetTemplateFactory.createTemplate(
+                    snippetContent,
+                    context.getProject(),
+                    this::getVariableValue,
+                    indentOptions,
+                    useTemplateForInvocationOnlySnippet
```
SCWells72, line 126: We now pass this state to the snippet-to-template transform.

src/main/java/com/redhat/devtools/lsp4ij/features/completion/SnippetTemplateFactory.java
```diff
@@ -33,15 +34,25 @@ public class SnippetTemplateFactory {
...
+                                                   LspSnippetIndentOptions indentOptions,
+                                                   boolean useTemplateForInvocationOnlySnippet) {
         Template template = TemplateManager.getInstance(project).createTemplate("", "");
         template.setInline(true);
-        LspSnippetParser parser = new LspSnippetParser(new SnippetTemplateLoader(template, variableResolver, indentOptions));
+        LspSnippetParser parser = new LspSnippetParser(new SnippetTemplateLoader(
+                template,
+                variableResolver,
+                indentOptions,
+                useTemplateForInvocationOnlySnippet
```
SCWells72, line 54: And propagate it to the template loader.

src/main/java/com/redhat/devtools/lsp4ij/features/completion/SnippetTemplateLoader.java
```diff
@@ -32,43 +37,126 @@
 public class SnippetTemplateLoader extends AbstractLspSnippetHandler {
 
     private final Template template;
+    private final boolean useTemplateForInvocationOnlySnippet;
 
-    private final List<String> existingVariables;
+    private final List<SnippetTemplateSegment> templateSegments = new LinkedList<>();
```
SCWells72, line 42: Instead of building the template on-the-fly, we accumulate the exact same information that we'd need to build the template, and we do it in `endSnippet()` instead. That allows us to have full context of the resulting template segments so we can see whether/how it can be transformed/simplified.

src/main/java/com/redhat/devtools/lsp4ij/features/completion/SnippetTemplateLoader.java
```diff
@@ -32,43 +37,126 @@
...
+                // If we find a variable not in an invocation, we can't simplify this template
+                else if ((templateSegment.isVariable() || templateSegment.isNamedVariable()) && !inInvocation) {
+                    hasVariablesOutsideInvocation = true;
+                    break;
+                }
+            }
 
+            // If we found an invocation and all variables were inside of it, use the simplified template segments
+            if (hasInvocation && !hasVariablesOutsideInvocation) {
+                effectiveTemplateSegments = simplifiedTemplateSegments;
```
SCWells72, line 110: If it looks like an invocation-only snippet, use the simplified segments below when building the actual template.

src/main/java/com/redhat/devtools/lsp4ij/features/completion/SnippetTemplateLoader.java
```diff
@@ -32,43 +37,129 @@
...
     @Override
     public void endSnippet() {
+        List<SnippetTemplateSegment> effectiveTemplateSegments = templateSegments;
+
+        // If we should not use a template for an invocation-only snippet, see whether this fits the pattern
+        if (!useTemplateForInvocationOnlySnippet) {
+            boolean hasInvocation = false;
+            // While it's unlikely to happen in a typical invocation arg list scenario, this helps ensure we're only
+            // reacting to the contents of balanced parentheses
+            int nestedInvocationCount = 0;
```
SCWells72, line 77: This is an additional safeguard against the unlikely possibility of nested parens in segments. We'll only come out of this logic deciding to use the simplified text segments if everything occurred within balanced invocation parens.

src/main/java/com/redhat/devtools/lsp4ij/features/completion/SnippetTemplateLoader.java
```diff
@@ -32,43 +37,129 @@
...
+    }
+
+    private static boolean isInvocationEnd(@NotNull String text) {
+        return StringUtil.trimLeading(text).startsWith(")");
     }
 
     @Override
     public void text(String text) {
-        template.addTextSegment(formatText(text));
+        templateSegments.add(SnippetTemplateSegment.textSegment(formatText(text)));
```
SCWells72, line 154: From here down, what you'll see is the change from building the template itself to accumulating what's needed to build the template in `endSnippet()`.

src/main/java/com/redhat/devtools/lsp4ij/features/completion/SnippetTemplateSegment.java
```diff
@@ -0,0 +1,132 @@
...
+
+import com.intellij.codeInsight.template.Expression;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * This is a simple wrapper for a {@link com.intellij.codeInsight.template.Template} segment that allows deferred
+ * creation of the corresponding template with all segment information retained.
+ */
+class SnippetTemplateSegment {
```
SCWells72, line 25: And this is the way we accumulate that state. I've basically added factory methods for each of the types of template segment that are created above and simple ways to check which type of segment each is so it can be used safely/properly when building the final template.

src/test/java/com/redhat/devtools/lsp4ij/features/completion/SnippetTemplateLoaderTest.java
```diff
@@ -0,0 +1,112 @@
...
+package com.redhat.devtools.lsp4ij.features.completion;
+
+import com.intellij.codeInsight.template.Template;
+import com.redhat.devtools.lsp4ij.fixtures.LSPCompletionFixtureTestCase;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Tests behavior of {@link SnippetTemplateLoader} via {@link SnippetTemplateFactory}.
+ */
+public class SnippetTemplateLoaderTest extends LSPCompletionFixtureTestCase {
```
SCWells72, line 24: Direct tests for `SnippetTemplateLoader` to exercise common and corner-case scenarios.

angelozerr:
It works like a charm, great job @SCWells72 !

PR #753, opened 2025-01-20
feat: Added support for `textDocument/onTypeFormatting` capability

SCWells72:
This PR adds support for the [`textDocument/onTypeFormatting`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_onTypeFormatting) LSP capability. To do so, it registers two EP implementations, a `typedHandler` for when non-**Enter** keys are typed and an `enterHandlerDelegate` for when **Enter** is typed. The logic is the exact same in both cases, basically checking whether or not a started language server for the file in which the character was typed supports that capability for the typed character and, if so, the corresponding `textDocument/onTypeFormatting` request is submitted to the server. Any text edits included in the response are then applied to the editor.

Note that so far I've only seen this capability provided by one language server, specifically [jdtls](https://github.com/eclipse-jdtls/eclipse.jdt.ls), and while that language server claims to support `}`, `;`, and `\n`:

```json
    "documentOnTypeFormattingProvider": {
      "firstTriggerCharacter": ";",
      "moreTriggerCharacter": [
        "\n",
        "}"
      ]
    }
```

I only see populated text edits returned for `}`. I've added unit tests for this specific level of observed support.


docs/LSPSupport.md
```diff
@@ -66,7 +66,7 @@ Current state of [Language Features]( https://microsoft.github.io/language-serve
  * ❌ [textDocument/pullDiagnostics](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_pullDiagnostics).
  * ✅ [textDocument/formatting](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_formatting) (see [implementation details](#formatting))
  * ✅ [textDocument/rangeFormatting](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_rangeFormatting) (see [implementation details](#formatting)).
- * ❌ [textDocument/onTypeFormatting](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_onTypeFormatting).
+ * ✅ [textDocument/onTypeFormatting](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_onTypeFormatting) (see [implementation details](#on-type-formatting)).
```
SCWells72, line 69: Updated the capability support matrix to show support for `textDocument/onTypeFormatting`.

docs/LSPSupport.md
```diff
@@ -427,6 +427,12 @@ Here is an example with the [TypeScript Language Server](./user-defined-ls/types
 [LSPFormattingOnlyService](https://github.com/redhat-developer/lsp4ij/blob/main/src/main/java/com/redhat/devtools/lsp4ij/features/formatting/LSPFormattingOnlyService.java) /
 [LSPFormattingAndRangeBothService](https://github.com/redhat-developer/lsp4ij/blob/main/src/main/java/com/redhat/devtools/lsp4ij/features/formatting/LSPFormattingAndRangeBothService.java) with the `formattingService` extension point.
 
+### On-Type Formatting
```
SCWells72, line 430: And added details on how that support is implemented.

src/main/java/com/redhat/devtools/lsp4ij/LSPFileSupport.java
```diff
@@ -66,6 +67,8 @@ public class LSPFileSupport extends UserDataHolderBase implements Disposable {
 
     private final LSPFormattingSupport formattingSupport;
 
+    private final LSPOnTypeFormattingSupport onTypeFormattingSupport;
```
SCWells72, line 70: Added capability support in all the necessary places.

src/main/java/com/redhat/devtools/lsp4ij/features/formatting/LSPServerSideOnTypeFormattingEnterHandler.java
```diff
@@ -0,0 +1,34 @@
...
+import com.intellij.openapi.actionSystem.DataContext;
+import com.intellij.openapi.editor.Editor;
+import com.intellij.psi.PsiFile;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Performs LSP-based on-type formatting if appropriate for the file when Enter is typed. For handling of other typed
+ * characters, see {@link LSPServerSideOnTypeFormattingTypedHandler}.
+ */
+public class LSPServerSideOnTypeFormattingEnterHandler extends EnterHandlerDelegateAdapter {
```
SCWells72, line 24: This is the `enterHandlerDelegate` that just immediately invokes `LSPServerSideOnTypeFormattingHelper.applyOnTypeFormatting()` with a typed character of `\n`. If text edits were applied, it short-circuits any further `enterHandlerDelegate` implementations.

src/main/java/com/redhat/devtools/lsp4ij/features/formatting/LSPServerSideOnTypeFormattingTypedHandler.java
```diff
@@ -0,0 +1,36 @@
...
+import com.intellij.openapi.editor.Editor;
+import com.intellij.openapi.project.Project;
+import com.intellij.psi.PsiFile;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Performs LSP-based on-type formatting if appropriate for the file and typed character. Note that this is not invoked
+ * for Enter/Newline; instead see {@link LSPServerSideOnTypeFormattingEnterHandler}.
+ */
+public class LSPServerSideOnTypeFormattingTypedHandler extends TypedHandlerDelegate {
```
SCWells72, line 24: This is the `typeHandler` that just immediately invokes `LSPServerSideOnTypeFormattingHelper.applyOnTypeFormatting()` with the typed character. If text edits were applied, it short-circuits any further `typedHandler` implementations.

src/main/java/com/redhat/devtools/lsp4ij/features/formatting/LSPServerSideOnTypeFormattingHelper.java
```diff
@@ -0,0 +1,151 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+
+import static com.redhat.devtools.lsp4ij.internal.CompletableFutures.isDoneNormally;
+import static com.redhat.devtools.lsp4ij.internal.CompletableFutures.waitUntilDone;
+
+/**
+ * Shared utility class for LSP on-type formatting.
+ */
+final class LSPServerSideOnTypeFormattingHelper {
```
SCWells72, line 46: This is what's called by both of the EP implementations to do the actual work.

src/main/java/com/redhat/devtools/lsp4ij/features/formatting/LSPServerSideOnTypeFormattingHelper.java
```diff
@@ -0,0 +1,151 @@
...
+     * @param file the PSI file
+     * @return the LSP on-type formatting information for the file if supported; otherwise null
+     */
+    @Nullable
+    private static Pair<DocumentOnTypeFormattingOptions, LSPOnTypeFormattingSupport> getOnTypeFormattingInfo(@NotNull PsiFile file) {
+        // On-type formatting shouldn't trigger a language server to start
+        Project project = file.getProject();
+        Set<LanguageServerWrapper> startedLanguageServers = LanguageServiceAccessor.getInstance(project).getStartedServers();
+        for (LanguageServerWrapper startedLanguageServer : startedLanguageServers) {
+            // TODO: Is there a better way to ask if this started language server supports the file?
```
SCWells72, line 135: This is the same as what I have in the client-side on-type formatting handler in the other PR/branch, so the same question holds. We don't want to suspend the IDE in response to the user typing a character to start a language server, so we first check whether there is a started language server for the current file. Is that the best way to do this? If not, what should I be doing instead?

src/main/java/com/redhat/devtools/lsp4ij/features/formatting/LSPServerSideOnTypeFormattingHelper.java
```diff
@@ -0,0 +1,151 @@
...
+            if (startedLanguageServer.isConnectedTo(LSPIJUtils.toUri(file))) {
+                LSPOnTypeFormattingFeature onTypeFormattingFeature = startedLanguageServer.getClientFeatures().getOnTypeFormattingFeature();
+                if (onTypeFormattingFeature.isEnabled(file) && onTypeFormattingFeature.isSupported(file)) {
```
SCWells72, line 138: Then we just make sure that the on-type formatting feature is enabled and supported and, if so, we return both the options -- which are needed to check whether or not the typed character is an on-type formatting trigger -- and the support -- which is needed to actually invoke the on-type formatting LSP feature.

src/main/resources/META-INF/plugin.xml
```diff
@@ -358,6 +358,16 @@ L
                 id="LSPFormattingAndRangeBothService"
                 implementation="com.redhat.devtools.lsp4ij.features.formatting.LSPFormattingAndRangeBothService"/>
 
+        <!-- LSP textDocument/onTypeFormatting support -->
+        <typedHandler
```
SCWells72, line 362: EP registrations to ensure that we handle both **Enter** and other characters that have been typed.

src/test/java/com/redhat/devtools/lsp4ij/features/formatting/JavaClientSideFormatOnCloseBraceTest.java
```diff
@@ -0,0 +1,137 @@
...
+
+import com.redhat.devtools.lsp4ij.fixtures.LSPServerSideOnTypeFormattingFixtureTestCase;
+
+/**
+ * Java-based server-side on-type formatting tests for format-on-close brace. Note that jdtls' support for on-type
+ * formatting does not seem to work well enough for format-on-statement terminator or newline even though those are
+ * included as supported on-type formatting characters in the language server initialization response. Both of those
+ * characters seem to yield empty text edit lists to be applied.
+ */
+public class JavaClientSideFormatOnCloseBraceTest extends LSPServerSideOnTypeFormattingFixtureTestCase {
```
SCWells72, line 22: The comment about pretty much says it, but basically I've only found support for this capability in the jdtls so far. The TypeScript, CSS, and Go language servers specifically don't support it, and I haven't yet set up language servers for other languages, but if others are found that support this capability, I'm happy to add tests for them as well.

Additionally while the initialization response for Java claims to support on-type formatting for the following trigger characters:

```json
    "documentOnTypeFormattingProvider": {
      "firstTriggerCharacter": ";",
      "moreTriggerCharacter": [
        "\n",
        "}"
      ]
    }
```

so far I've only seen a populated list of text edits to be applied returned when `}` is typed. As a result, I've only implemented a unit test for that specific combination of Java and close brace here.

src/test/java/com/redhat/devtools/lsp4ij/fixtures/LSPServerSideOnTypeFormattingFixtureTestCase.java
```diff
@@ -0,0 +1,133 @@
...
+import java.util.LinkedList;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Base class test case for server-side on-type formatting tests by emulating LSP 'textDocument/onTypeFormatting' responses.
+ */
+public abstract class LSPServerSideOnTypeFormattingFixtureTestCase extends LSPCodeInsightFixtureTestCase {
```
SCWells72, line 41: Base test class for on-type formatting unit tests.

src/test/java/com/redhat/devtools/lsp4ij/mock/MockTextDocumentService.java
```diff
@@ -177,7 +177,7 @@ public CompletableFuture<List<? extends TextEdit>> rangeFormatting(DocumentRange
 
     @Override
     public CompletableFuture<List<? extends TextEdit>> onTypeFormatting(DocumentOnTypeFormattingParams params) {
-        return CompletableFuture.completedFuture(null);
+        return CompletableFuture.completedFuture(mockFormattingTextEdits);
```
SCWells72, line 180: Have to return the registered mock `TextEdit` values for `onTypeFormatting`.

SCWells72:
@angelozerr you may want to review this one before the client-side one since it adds actual server-side LSP on-type formatting support. Having said that, I would very much like for both to get merged if possible because the client-side option is definitely valuable for those working against language servers that lack the `textDocument/onTypeFormatting` capability.

angelozerr:
> This PR adds support for the [`textDocument/onTypeFormatting`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_onTypeFormatting) LSP capability. To do so, it registers two EP implementations, a `typedHandler` for when non-**Enter** keys are typed and an `enterHandlerDelegate` for when **Enter** is typed. The logic is the exact same in both cases, basically checking whether or not a started language server for the file in which the character was typed supports that capability for the typed character and, if so, the corresponding `textDocument/onTypeFormatting` request is submitted to the server. Any text edits included in the response are then applied to the editor.
> 
> Note that so far I've only seen this capability provided by one language server, specifically [jdtls](https://github.com/eclipse-jdtls/eclipse.jdt.ls), and while that language server claims to support `}`, `;`, and `\n`:
> 
> ```json
>     "documentOnTypeFormattingProvider": {
>       "firstTriggerCharacter": ";",
>       "moreTriggerCharacter": [
>         "\n",
>         "}"
>       ]
>     }
> ```
> 
> I only see populated text edits returned for `}`. I've added unit tests for this specific level of observed support.

@rgrunber @datho7561 could you give us some usecase which uses other character than } like @SCWells72 said?

docs/LSPSupport.md
```diff
@@ -427,6 +427,12 @@ Here is an example with the [TypeScript Language Server](./user-defined-ls/types
 [LSPFormattingOnlyService](https://github.com/redhat-developer/lsp4ij/blob/main/src/main/java/com/redhat/devtools/lsp4ij/features/formatting/LSPFormattingOnlyService.java) /
 [LSPFormattingAndRangeBothService](https://github.com/redhat-developer/lsp4ij/blob/main/src/main/java/com/redhat/devtools/lsp4ij/features/formatting/LSPFormattingAndRangeBothService.java) with the `formattingService` extension point.
 
+### On-Type Formatting
+
```
angelozerr, line 431: If a file is associated to a language server which have ontypeformating capability, is client on type formatting is disabled?

We need to clarify the behavoir somewhere.

src/main/java/com/redhat/devtools/lsp4ij/features/formatting/LSPServerSideOnTypeFormattingHelper.java
angelozerr: Nice catch! I think you can do that with the standard getLanguageServers by using a before filter which matches ybe server status with started. 

In the filter you receive the languageserver item which defines a getServerStatus if I remember

datho7561:
> could you give us some usecase which uses other character than }

I can't think if anything off the top of my head. Quite often formatting only works well when there are no syntax errors in the code, and `;` `}` are pretty safe bets for that. Maybe `)`?

docs/LSPSupport.md
SCWells72: @angelozerr I'm not sure that I understand this question, but I'll try to answer as best I can. This capability is evaluated independent of other capabilities. If the language server claims to support the feature, this implementation will try to use it when one of the language server's reported trigger characters has been typed. Does that help?

src/main/java/com/redhat/devtools/lsp4ij/features/formatting/LSPServerSideOnTypeFormattingHelper.java
SCWells72: Okay. Do you think it's worth switching to that? Is there anything wrong with the approach taken here?

docs/LSPSupport.md
angelozerr: Sorry with my bad explanation.

You did an another PR which manage on type formatting on client side which uses selection ranges or folding.

Is this feature is disabled when the file is associated to a language server which supports ontypeformatting?

SCWells72:
> I can't think if anything off the top of my head. Quite often formatting only works well when there are no syntax errors in the code, and `;` `}` are pretty safe bets for that. Maybe `)`?

The language server specifies exactly which typed characters are considered on-type formatting triggers, e.g.:

```json
    "documentOnTypeFormattingProvider": {
      "firstTriggerCharacter": ";",
      "moreTriggerCharacter": [
        "\n",
        "}"
      ]
    },
```

When submitting an on-type formatting LSP request, you then specify the character that was typed and the position in the file at which it was typed. I don't think that it would make sense to submit a request with any character that wasn't listed specifically as an on-type formatting.

I do agree that this feature may well be quite limited by how well the server's parser can recover from syntax errors, but I would think that the specific characters listed above -- statement terminator (`;`), newline (`\n`), and close brace (`}`) would generally leave the file's contents in a syntactically valid state. However, I'm only seeing populated responses from jdtls (Java) for `documentText/onTypeFormatting` requests when `}` is typed and not when `;` or `\n` are typed.

I may try to look at a few other language servers, but it would be useful to know which support this particular capability. The TypeScript, CSS, and Go language servers do not. So far I've only found the Java language server to provide support, but I admit that I haven't looked at the language servers for Scala, Swift, C/C++/Objective-C, etc., yet.

docs/LSPSupport.md
SCWells72: Ah, that makes sense. No, right now these are two independent branches/PRs both based on `main`. Once the server-side PR is merged, I'll update the client-side PR accordingly. Instead of making them mutually-exclusive, though, I was thinking about making it so that the client-side typed handler is ordered before the server-side typed handler so that if the user preferred the behavior of client-side on-type formatting, it would supersede server-side on-type formatting if enabled. Obviously if client-side on-type formatting isn't enabled, it's a total no-op anyway.

SCWells72:
For what it's worth, I checked to see whether using the position from _before_ the character was typed might make any difference, but the results are the same. The specification is purposefully vague about what the position should actually be:

```typescript
/**
 * The position around which the on type formatting should happen.
 * This is not necessarily the exact position where the character denoted
 * by the property `ch` got typed.
 */
position: Position;
```

so I think that using the position _after_ the character has been typed should be fine, especially since I see no actual difference in the responses between the before and after positions. Bummer...

CppCXY:
> For what it's worth, I checked to see whether using the position from _before_ the character was typed might make any difference, but the results are the same. The [specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#documentOnTypeFormattingParams) is purposefully vague about what the position should actually be:
> 
> ```ts
> /**
>  * The position around which the on type formatting should happen.
>  * This is not necessarily the exact position where the character denoted
>  * by the property `ch` got typed.
>  */
> position: Position;
> ```
> 
> so I think that using the position _after_ the character has been typed should be fine, especially since I see no actual difference in the responses between the before and after positions. Bummer...

I remember that vscode sends the location after typing because I have developed a language server that supports type formatting.

src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPOnTypeFormattingFeature.java
```diff
@@ -0,0 +1,77 @@
...
+ */
+@ApiStatus.Experimental
+public class LSPOnTypeFormattingFeature extends AbstractLSPDocumentFeature {
+
+    private OnTypeFormattingCapabilityRegistry onTypeFormattingCapabilityRegistry;
+
+    @Override
+    public boolean isEnabled(@NotNull PsiFile file) {
+        // Need to perform the same check as in LSPFormattingFeature.isEnabled() to ensure that LSP formatting should be used
+        if (!getClientFeatures().getFormattingFeature().isExistingFormatterOverrideable(file) &&
```
SCWells72, line 33: I believe this is the check that you'd mentioned needed to occur, correct? Let me know if it's not correct? Obviously I could just call `getClientFeatures().getFormattingFeature().isEnabled(file)` instead, but if that method were overridden in a derived class, we might disable on-type formatting when only the full formatter should be disabled. Let me know if you'd prefer that I just simplify it to that, though.

src/main/java/com/redhat/devtools/lsp4ij/features/formatting/LSPOnTypeFormattingSupport.java
```diff
@@ -0,0 +1,97 @@
...
+    protected CompletableFuture<List<TextEdit>> doLoad(DocumentOnTypeFormattingParams params, CancellationSupport cancellationSupport) {
+        PsiFile file = super.getFile();
+        return onTypeFormatting(file, params, cancellationSupport);
+    }
+
+    private static @NotNull CompletableFuture<List<TextEdit>> onTypeFormatting(@NotNull PsiFile file,
+                                                                               @NotNull DocumentOnTypeFormattingParams params,
+                                                                               @NotNull CancellationSupport cancellationSupport) {
+        return getLanguageServers(file,
+                f -> f.getFoldingRangeFeature().isEnabled(file),
```
angelozerr, line 57: Use `getOnTypeFormattingFeature()`

src/main/java/com/redhat/devtools/lsp4ij/features/formatting/LSPOnTypeFormattingSupport.java
```diff
@@ -0,0 +1,97 @@
...
+                f -> f.getFoldingRangeFeature().isSupported(file))
```
angelozerr, line 58: Same thing here

src/main/java/com/redhat/devtools/lsp4ij/features/formatting/LSPOnTypeFormattingSupport.java
```diff
@@ -0,0 +1,97 @@
...
+                .thenComposeAsync(languageServers -> {
+                    // Here languageServers is the list of language servers which matches the given file
+                    // and which have on-type formatting capability
+                    if (languageServers.isEmpty()) {
+                        return CompletableFuture.completedStage(Collections.emptyList());
```
angelozerr, line 63: CompletableFuture.completableFuture (to be consistent with other code)

src/main/java/com/redhat/devtools/lsp4ij/features/formatting/LSPServerSideOnTypeFormattingHelper.java
```diff
@@ -0,0 +1,151 @@
...
+
+    /**
+     * Returns the LSP on-type formatting information for the provided file if supported. This includes the on-type
+     * formatting options and the on-type formatting support.
+     *
+     * @param file the PSI file
+     * @return the LSP on-type formatting information for the file if supported; otherwise null
+     */
+    @Nullable
+    private static Pair<DocumentOnTypeFormattingOptions, LSPOnTypeFormattingSupport> getOnTypeFormattingInfo(@NotNull PsiFile file) {
```
angelozerr, line 130: Instead of using Pair, I suggest that you define a private record.

src/main/java/com/redhat/devtools/lsp4ij/server/capabilities/OnTypeFormattingCapabilityRegistry.java
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * Server capability registry for 'textDocument/onTypeFormatting'.
+ */
+public class OnTypeFormattingCapabilityRegistry extends TextDocumentServerCapabilityRegistry<DocumentOnTypeFormattingRegistrationOptions> {
+
+    private static final @NotNull Predicate<@NotNull ServerCapabilities> SERVER_CAPABILITIES_PREDICATE = sc ->
+            sc.getDocumentOnTypeFormattingProvider() != null;
+
+    public OnTypeFormattingCapabilityRegistry(@NotNull LSPClientFeatures clientFeatures) {
+        super(clientFeatures, EditorFeatureType.FOLDING);
```
angelozerr, line 36: super(clientFeatures);

src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPOnTypeFormattingFeature.java
angelozerr: Great! Thanks.

src/main/java/com/redhat/devtools/lsp4ij/features/formatting/LSPServerSideOnTypeFormattingHelper.java
```diff
@@ -0,0 +1,151 @@
...
+     *
+     * @param charTyped the type character
+     * @param editor    the editor
+     * @param file      the PSI file
+     * @return true if on-type formatting was applied; otherwise false
+     */
+    static boolean applyOnTypeFormatting(char charTyped,
+                                         @NotNull Editor editor,
+                                         @NotNull PsiFile file) {
+        Pair<DocumentOnTypeFormattingOptions, LSPOnTypeFormattingSupport> onTypeFormattingInfo = getOnTypeFormattingInfo(file);
```
angelozerr, line 65: Please remove this getOnTypeFormattingInfo method

src/main/java/com/redhat/devtools/lsp4ij/features/formatting/LSPServerSideOnTypeFormattingHelper.java
```diff
@@ -0,0 +1,151 @@
...
+            return false;
+        }
+
+        DocumentOnTypeFormattingOptions onTypeFormattingOptions = onTypeFormattingInfo.getFirst();
+        LSPOnTypeFormattingSupport onTypeFormattingSupport = onTypeFormattingInfo.getSecond();
+
+        // Make sure the typed character should trigger on-type formatting for this language
+        Set<String> triggerCharacters = new LinkedHashSet<>();
+        ContainerUtil.addIfNotNull(triggerCharacters, onTypeFormattingOptions.getFirstTriggerCharacter());
+        ContainerUtil.addAllNotNull(triggerCharacters, onTypeFormattingOptions.getMoreTriggerCharacter());
```
angelozerr, line 76: Remove the code until here.

src/main/java/com/redhat/devtools/lsp4ij/features/formatting/LSPServerSideOnTypeFormattingHelper.java
```diff
@@ -0,0 +1,151 @@
...
+        String charTypedAsString = String.valueOf(charTyped);
+        if (!triggerCharacters.contains(charTypedAsString)) {
```
angelozerr, line 78: Remove this condition

src/main/java/com/redhat/devtools/lsp4ij/features/formatting/LSPServerSideOnTypeFormattingHelper.java
angelozerr, line 130: Remove this method

src/main/java/com/redhat/devtools/lsp4ij/features/formatting/LSPOnTypeFormattingSupport.java
```diff
@@ -0,0 +1,97 @@
...
+    @Override
+    protected CompletableFuture<List<TextEdit>> doLoad(DocumentOnTypeFormattingParams params, CancellationSupport cancellationSupport) {
+        PsiFile file = super.getFile();
+        return onTypeFormatting(file, params, cancellationSupport);
+    }
+
+    private static @NotNull CompletableFuture<List<TextEdit>> onTypeFormatting(@NotNull PsiFile file,
+                                                                               @NotNull DocumentOnTypeFormattingParams params,
+                                                                               @NotNull CancellationSupport cancellationSupport) {
+        return getLanguageServers(file,
```
angelozerr, line 56: Replace with:

```java
getLanguageServers(file,
                f -> f.getServerStatus() == ServerStatus.started &&
                                     f.getOnTypeFormattingFeature().isEnabled(file) &&
                                    f.getOnTypeFormattingFeature().isTriggerCharactersSupported(charedType),
```

src/main/java/com/redhat/devtools/lsp4ij/features/formatting/LSPOnTypeFormattingSupport.java
```diff
@@ -0,0 +1,97 @@
...
+        return super.getFeatureData(params);
+    }
+
+    @Override
+    protected CompletableFuture<List<TextEdit>> doLoad(DocumentOnTypeFormattingParams params, CancellationSupport cancellationSupport) {
+        PsiFile file = super.getFile();
+        return onTypeFormatting(file, params, cancellationSupport);
+    }
+
+    private static @NotNull CompletableFuture<List<TextEdit>> onTypeFormatting(@NotNull PsiFile file,
```
angelozerr, line 53: Add a new parameter `String charedType`

src/main/java/com/redhat/devtools/lsp4ij/features/formatting/LSPOnTypeFormattingSupport.java
angelozerr: to know how to implement `isTriggerCharactersSupported`, I suggest you do like completion https://github.com/redhat-developer/lsp4ij/blob/9bb7c8d4b2ba70f6e902d739355fdb4754ade3f7/src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPCompletionFeature.java#L111

src/main/java/com/redhat/devtools/lsp4ij/features/formatting/LSPOnTypeFormattingSupport.java
SCWells72: Wasn't necessary because it's already part of `DocumentOnTypeFormattingParams`.

src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPOnTypeFormattingFeature.java
```diff
@@ -0,0 +1,96 @@
...
+ */
+@ApiStatus.Experimental
+public class LSPOnTypeFormattingFeature extends AbstractLSPDocumentFeature {
+
+    private OnTypeFormattingCapabilityRegistry onTypeFormattingCapabilityRegistry;
+
+    @Override
+    public boolean isEnabled(@NotNull PsiFile file) {
+        // This feature should only be enabled if the server is already started
+        if (getServerStatus() != ServerStatus.started) {
```
angelozerr, line 34: Please add a comment why you do that (for permanence raison)

angelozerr (changes requested):

src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPOnTypeFormattingFeature.java
SCWells72: Done.

angelozerr:
Code is excellent and the features works like a charm. Amazing!

Thanks so much @SCWells72 !

PR #767, opened 2025-01-24
feat: Automatic folding of file header comment and imports regions if so configured in the IDE's code folding settings (fixes #749)

### LINKED ISSUE #749: Add support for folding-by-default of certain regions
JetBrains' IDEs provide support for automatic folding for certain regions, some language-agnostic and some language-specific, e.g.:

Image

The `textDocument/foldingRange` response does provide some limited categorization of folding ranges, e.g., the following from the TypeScript language server:

```json
[
  {
    "startLine": 5,
    "endLine": 55,
    "kind": "imports"
  },
]
```

and the following from the Java language server (jdtls):

```json
[
  {
    "startLine": 0,
    "endLine": 9,
    "kind": "comment"
  },
  {
    "startLine": 21,
    "endLine": 23,
    "kind": "comment"
  },
  {
    "startLine": 13,
    "endLine": 19,
    "kind": "imports"
  },
]
```

It would be nice if these could either be automatically mapped to the corresponding IDE folding regions, e.g., `kind: imports` => **Settings | Editor | General | Code Folding | Fold by default | General | Imports**, or if minimally there could be some other way to specify which values for `kind` should be folded by default, e.g., through client configuration.

SCWells72:
JetBrains' IDEs support the notion of configurable code folding ranges that should be collapsed by default when a file is first opened in an editor tab:

image

LSP's [`textDocument/foldingRange`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_foldingRange) capability allows a language server to include a [`kind`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#foldingRangeKind) attribute for each folding range with standard values of `comment`, `imports`, and `region`.

The `imports` kind maps cleanly to the IDE's standard notion of **General | Imports**, and a folding range with the `comment` kind that starts on the first line of a file can be mapped to the IDE's standard notion of **General | File header**. This PR includes changes to do just that.

At present, only two language servers of the ones I've tried -- which is admittedly only a small subset -- include `kind` values for the returned folding ranges. The [Java language server](https://github.com/eclipse-jdtls/eclipse.jdt.ls) specifies folding ranges as both `comment` and `imports`, and the [TypeScript language server](https://github.com/typescript-language-server/typescript-language-server) specifies folding ranges as `imports`. I've tested these changes against both including unit test updates accordingly.

Note that the IDE is not guaranteed to fold regions that identify as collapsed-by-default when an editor tab is opened. Such inquiries by the IDE are very much subject to `ProcessCanceledException`, and the previous explicit folding state of a given region is also retained by the IDE when the same file is reopened. This is true of first-class-supported languages as well. As a result, if testing manually, don't be surprised if you don't see the expected ranges collapsed by default when a file is opened. Instead open the IDE settings and toggle the respective settings off and on to confirm the proper behavior reliably.

Here is a demonstration of the feature in a Java source file where both the file header comment and imports are collapsed:

LSP_Collapse_By_Default_Java

and here is a demonstration in a TypeScript source file where only imports are collapsed:

LSP_Collapse_By_Default_TypeScript


src/main/java/com/redhat/devtools/lsp4ij/features/foldingRange/LSPFoldingRangeBuilder.java
```diff
@@ -67,12 +69,15 @@ protected void buildLanguageFoldRegions(@NotNull List<FoldingDescriptor> descrip
             for (FoldingRange foldingRange : foldingRanges) {
                 TextRange textRange = getTextRange(foldingRange, file, document);
                 if ((textRange != null) && (textRange.getLength() > 0)) {
-                    String collapsedText = foldingRange.getCollapsedText();
-                    if (collapsedText != null) {
-                        descriptors.add(new FoldingDescriptor(root.getNode(), textRange, null, collapsedText));
-                    } else {
-                        descriptors.add(new FoldingDescriptor(root.getNode(), textRange));
-                    }
+                    descriptors.add(new FoldingDescriptor(
```
SCWells72, line 72: No need for the branching logic anymore as we're calling a `FoldingDescriptor` constructor signature that allows a `null` value for `collapsedText`. We also pass a flag denoting whether or not the region should be collapsed by default which is, of course, the whole point of the changes in this PR.

src/main/java/com/redhat/devtools/lsp4ij/features/foldingRange/LSPFoldingRangeBuilder.java
```diff
@@ -194,6 +199,29 @@ protected String getLanguagePlaceholderText(@NotNull ASTNode node, @NotNull Text
         return null;
     }
 
+    private static boolean isCollapsedByDefault(@NotNull FoldingRange foldingRange) {
+        String foldingRangeKind = foldingRange.getKind();
+        if (foldingRangeKind != null) {
+            CodeFoldingSettings codeFoldingSettings = CodeFoldingSettings.getInstance();
+
+            // Imports
+            if (codeFoldingSettings.COLLAPSE_IMPORTS &&
```
SCWells72, line 208: If the IDE is configured to collapse imports by default and this folding region's `kind` is `imports`, it should be collapsed by default.

src/main/java/com/redhat/devtools/lsp4ij/features/foldingRange/LSPFoldingRangeBuilder.java
```diff
@@ -194,6 +199,29 @@ protected String getLanguagePlaceholderText(@NotNull ASTNode node, @NotNull Text
...
+            CodeFoldingSettings codeFoldingSettings = CodeFoldingSettings.getInstance();
+
+            // Imports
+            if (codeFoldingSettings.COLLAPSE_IMPORTS &&
+                FoldingRangeKind.Imports.equals(foldingRangeKind)) {
+                return true;
+            }
+
+            // File header comments
+            else if (codeFoldingSettings.COLLAPSE_FILE_HEADER &&
```
SCWells72, line 214: If the IDE is configured to collapse file header comments by default and this folding region's `kind` is `comment` and the region begins on the very first line of the file, it should be collapsed by default. Note that ideally we'd also check that the comment itself begins in the first column of the first line, but the language servers seem to report the folding region at beginning in the last non-whitespace column of the first line of the region, and we don't have sufficient information about the language's start/end block comment string patterns to perform a search backward from the corresponding offset. Worst-case scenario, if the file happens to look something like this:

```typescript
```typescript
console.log('Log something');/* Block comment
                                starts on the
                                first line but
                                not first column
                              */
console.log('Log something else');
```

the comment would be folded so that it looks like:

```typescript
console.log('Log something');/*[...]
console.log('Log something else');
```

I think that's **very** unlikely, and the resulting effect isn't bad, just perhaps unexpected.

src/test/java/com/redhat/devtools/lsp4ij/features/foldingRange/JavaFoldingRangeTest.java
```diff
@@ -0,0 +1,121 @@
...
+
+package com.redhat.devtools.lsp4ij.features.foldingRange;
+
+import com.intellij.codeInsight.folding.CodeFoldingSettings;
+import com.redhat.devtools.lsp4ij.fixtures.LSPFoldingRangeFixtureTestCase;
+
+/**
+ * Folding range tests by emulating LSP 'textDocument/foldingRange' responses from jdtls.
+ */
+public class JavaFoldingRangeTest extends LSPFoldingRangeFixtureTestCase {
```
SCWells72, line 20: I added a test for Java, but it's not as comprehensive as the one for TypeScript as it would otherwise just be almost exactly the same. I did include tests specifically for its support for both `comment` and `imports` folding range kinds, though, and I confirm that regions are collapsed-by-default as appropriate based on the respective IDE configuration settings.

src/test/java/com/redhat/devtools/lsp4ij/features/foldingRange/TypeScriptFoldingRangeTest.java
```diff
@@ -75,4 +76,49 @@ export class Demo {<start1>
                         """
         );
     }
+
+    // Collapse-by-default tests
```
SCWells72, line 80: And I added more tests to the existing TypeScript test suite specifically for its support for the `imports` folding range kind, and as above, I confirm that regions are collapsed-by-default as appropriate based on the respective IDE configuration setting.

src/test/java/com/redhat/devtools/lsp4ij/fixtures/LSPFoldingRangeFixtureTestCase.java
```diff
@@ -48,12 +50,40 @@ protected LSPFoldingRangeFixtureTestCase(String... fileNamePatterns) {
...
+     * @param fileBody                       the file body with embedded tag pairs for expected fold regions
+     * @param mockFoldingRangesJson          the mock <code>textDocument/foldingRange</code> response
+     * @param collapsedByDefaultRangeNumbers the optional 1-based numbers of the expected ranges which should be
+     *                                       collapsed initially
+     */
     protected void assertFoldingRanges(@NotNull String fileName,
                                        @NotNull String fileBody,
-                                       @NotNull String mockFoldingRangesJson) {
+                                       @NotNull String mockFoldingRangesJson,
+                                       @NotNull Integer... collapsedByDefaultRangeNumbers) {
```
SCWells72, line 67: And the big change here is that I had to add a way to verify which folding ranges are expected to be collapsed initially.

src/test/java/com/redhat/devtools/lsp4ij/fixtures/LSPFoldingRangeFixtureTestCase.java
```diff
@@ -48,12 +50,40 @@ protected LSPFoldingRangeFixtureTestCase(String... fileNamePatterns) {
...
         MockLanguageServer.INSTANCE.setTimeToProceedQueries(100);
         List<FoldingRange> mockFoldingRanges = JSONUtils.getLsp4jGson().fromJson(mockFoldingRangesJson, new TypeToken<List<FoldingRange>>() {
         }.getType());
+        // Sort for stability
+        mockFoldingRanges.sort((foldingRange1, foldingRange2) -> {
```
SCWells72, line 72: The folding ranges weren't always returned in a consistent or stable order, so I order them explicitly here so that folding ranges from LSP and fold regions from the IDE are in the same order. You'll see another sort below for the latter.

src/test/java/com/redhat/devtools/lsp4ij/fixtures/LSPFoldingRangeFixtureTestCase.java
```diff
@@ -90,6 +120,9 @@ protected void assertFoldingRanges(@NotNull String fileName,
             String expectedPlaceholderText = StringUtil.isNotEmpty(mockCollapsedText) ? mockCollapsedText : "...";
             String actualPlaceholderText = actualFoldRegion.getPlaceholderText();
             assertEquals(expectedPlaceholderText, actualPlaceholderText);
+
+            // Check the initial expansion state
+            assertEquals("Incorrect expansion state for region " + (i + 1) + ".", ArrayUtil.contains(i + 1, collapsedByDefaultRangeNumbers), !actualFoldRegion.isExpanded());
```
SCWells72, line 125: This is where it checks whether each fold region is in the correct initial expanded vs. collapsed state.

src/test/java/com/redhat/devtools/lsp4ij/fixtures/LSPFoldingRangeFixtureTestCase.java
```diff
@@ -167,6 +200,9 @@ private static List<TextRange> getExpectedTextRanges(@NotNull String fileBody) {
             int endOffset = endOffsets.get(i);
             expectedTextRanges.add(TextRange.create(startOffset, endOffset));
         }
+
+        // Sort for stability
+        expectedTextRanges.sort(TextRangeUtil.RANGE_COMPARATOR);
```
SCWells72, line 205: As mentioned above, sorting for consistency and stability so that we can have parallel lists of mock folding ranges and fold region text ranges.

SCWells72:
@angelozerr, this is the next one that's ready for review. Hopefully it'll be pretty straightforward. Let me know any feedback you have on it. When this one is complete, I should have the client-side on-type formatting PR ready for full review.

src/main/java/com/redhat/devtools/lsp4ij/features/foldingRange/LSPFoldingRangeBuilder.java
```diff
@@ -194,6 +199,29 @@ protected String getLanguagePlaceholderText(@NotNull ASTNode node, @NotNull Text
         return null;
     }
 
+    private static boolean isCollapsedByDefault(@NotNull FoldingRange foldingRange) {
```
angelozerr, line 202: I think it would be very nice to provide the capability to customize collapsed folding.

I suggest that you move (as non static) in the `LSPFoldingFeatures#isCollapsedByDefault(@NotNull FoldingRange foldingRange) ` and update also the LSPApi.md

docs/LSPApi.md
```diff
@@ -374,10 +374,11 @@ public class MyLSPDocumentSymbolFeature extends LSPDocumentSymbolFeature {
...
+| boolean isSupported(PsiFile file)                                     | Returns `true` if the LSP feature is supported for the given file and `false` otherwise. <br/>This supported state is called after starting the language server, which matches the file and user with the LSP server capabilities. | Check the server capability                                                                                                                                                                         |
+| boolean isCollapsedByDefault(PsiFile file, FoldingRange foldingRange) | Returns `true` if the folding range in the given file should be collapsed by default and `false` otherwise.                                                                                                                        | `true` for file header comments and imports when the respective values for `kind` are specified by the language server and those regions are configured as collapsed by default in the IDE settings |
```
SCWells72, line 381: Added `isCollapsedByDefault()` to the API for this feature.

src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPFoldingRangeFeature.java
```diff
@@ -62,4 +65,42 @@ public void setServerCapabilities(@Nullable ServerCapabilities serverCapabilitie
...
+     *     <li><code>kind</code> is {@link FoldingRangeKind#Comment}, {@link CodeFoldingSettings#COLLAPSE_FILE_HEADER}
+     *         is true, and the folding range starts on the file's first line
+     *     </li>
+     * </ul>
+     *
+     * @param file         the PSI file
+     * @param foldingRange the folding range
+     * @return true if the folding range should be collapsed by default; otherwise false
+     */
+    public boolean isCollapsedByDefault(@NotNull PsiFile file, @NotNull FoldingRange foldingRange) {
```
SCWells72, line 85: The default implementation is exactly as it was in the other location and is documented well here.

src/main/java/com/redhat/devtools/lsp4ij/features/foldingRange/LSPFoldingRange.java
```diff
@@ -0,0 +1,52 @@
...
+
+package com.redhat.devtools.lsp4ij.features.foldingRange;
+
+import org.eclipse.lsp4j.FoldingRange;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Augments {@link FoldingRange} with add a flag denoting whether or not it should be collapsed by default.
+ */
+class LSPFoldingRange extends FoldingRange {
```
SCWells72, line 20: Pretty much what it says, but this allows us to capture whether or not the folding range is collapsed by default while building the list of folding ranges from each language server. Note that it's package-protected visibility to prevent API contamination.

src/main/java/com/redhat/devtools/lsp4ij/features/foldingRange/LSPFoldingRangeBuilder.java
```diff
@@ -67,12 +67,15 @@ protected void buildLanguageFoldRegions(@NotNull List<FoldingDescriptor> descrip
...
-                        descriptors.add(new FoldingDescriptor(root.getNode(), textRange));
-                    }
+                    descriptors.add(new FoldingDescriptor(
+                            root.getNode(),
+                            textRange,
+                            null,
+                            Collections.emptySet(),
+                            false,
+                            foldingRange.getCollapsedText(),
+                            (foldingRange instanceof LSPFoldingRange lspFoldingRange) && lspFoldingRange.isCollapsedByDefault()
```
SCWells72, line 77: And now we just derive whether or not the range should be collapsed by default from the folding range itself as set, and only if it's our augmented subclass of `FoldingRange` with that information.

docs/LSPSupport.md
```diff
@@ -371,6 +371,26 @@ Here is an example with the [TypeScript Language Server](./user-defined-ls/types
 
 ![textDocument/foldingRange](./images/lsp-support/textDocument_foldingRange.gif)
 
+#### Automatic code folding
```
SCWells72, line 374: Added a new sub-section on automatic code folding.

docs/LSPSupport.md
```diff
@@ -371,6 +371,26 @@ Here is an example with the [TypeScript Language Server](./user-defined-ls/types
...
+
+* Those with `kind=imports` when **Imports** are configured to be folded by default in the IDE.
+* Those with `kind=comment` found on the first line of a file when **File header** is configured to be folded by default in the IDE. 
+
+Here is an example of collapse by default behavior with the [Java Language Server](https://github.com/eclipse-jdtls/eclipse.jdt.ls)
+showing automatic folding of both the file header comment and imports when the file is opened:
+
+![textDocument/foldingRange_collapseByDefault](./images/lsp-support/textDocument_foldingRange_collapseByDefault.gif)
+
+#### Code block provider
```
SCWells72, line 392: And added a sub-section header for the code block provider since it's beneath the code folding section.

angelozerr:
Amazing work @SCWells72 ! Thanks so much!

Issue #606, opened 2024-11-21
Inlay hint occasionally displays double.


CppCXY:
![Image](https://github.com/user-attachments/assets/3ca05a33-e515-4d3e-a502-c5fecff59a47)

heanrum:
![Image](https://github.com/user-attachments/assets/c92552e0-e5da-4cfa-8863-a0bea71cb4d6)

angelozerr:
My main problem is that I cannot reproduce it, but I suspect that some IJ version changes the session behavior which is used to store a flag which says "inlayhint are already processed).

Today LSP4IJ uses deprecated inlay hint, we should move to the new API.

I have no time for the moment to work on this migration, but any contribution are welcome!

Issue #382, opened 2024-06-21
Inlay hints look too small


InSyncWithFoo:
At a 14px font size, inlay hints seem too small in the middle of code parts:

![](https://github.com/redhat-developer/lsp4ij/assets/122007197/c1349b98-9a04-4206-97b5-3d2070efa63f)

Compare that with Kotlin's in IntelliJ IDEA:

![](https://github.com/redhat-developer/lsp4ij/assets/122007197/6f28a77e-3941-4a8c-b169-09871f688c0b)

...or, better yet, PyCharm's own inlay hints:

![](https://github.com/redhat-developer/lsp4ij/assets/122007197/1aa18ad5-1cc1-4346-8b6c-4e3c3ec46808)

The hints' font sizes are the same (~10px), but the latter looks better due to ~6px of top and bottom padding.

Seems that I'll have to inherit from `LSPInlayHintsProvider` and override its `.toPresentation()`?

fbricon:
so you mean the padding around inlayhints?

fbricon:
If you know how to fix it, you can open a PR on LSPInlayHintsProvider ;-)

angelozerr:
> According to its documentation comment, the former is likely to be deprecated in the future. The latter was first made available [in 2023.1](https://youtrack.jetbrains.com/issue/IJPL-29785).

Indeed we need to use [hints.declarative.InlayHintsProvider](https://github.com/JetBrains/intellij-community/blob/999e342ee6f3083a53be2ee99f5348e0acfda9f0/platform/lang-api/src/com/intellij/codeInsight/hints/declarative/InlayHintsProvider.kt) as you say. I use the new API for Codelens (CodeVision) but I had no time to migrate InlaHint with the new API.

Any contribution are welcome!

angelozerr:
@InSyncWithFoo I think we can close the issue, right?

PR #786, opened 2025-01-30
feat: Inlay hints refactor

FalsePattern:
## Goals
- ~~Completely remove the dependency on the deprecated inlay hints API.~~ non-goal for this PR, it's needed for color hints
- Migrate inlay hints support to the declarative API
- Fix the double inlay hints bug

## Fixes
- https://github.com/redhat-developer/lsp4ij/issues/606, primary goal of this PR
- https://github.com/redhat-developer/lsp4ij/issues/382, as a side effect of the rendering being completely delegated to the intellij gui backend 
image

## Current draft regressions
- ~~Inlay hint hover/click actions are not yet implemented~~ fixed
- ~~`textDocument/color` support is impossible through the declarative API due to its limitations, it only supports text nodes, and "trees" of text nodes~~ fixed, uses the old api

### DIFF
--- a/src/main/java/com/redhat/devtools/lsp4ij/LanguageServersRegistry.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/LanguageServersRegistry.java
@@ -10,8 +10,8 @@
  ******************************************************************************/
 package com.redhat.devtools.lsp4ij;
 
-import com.intellij.codeInsight.hints.NoSettings;
 import com.intellij.codeInsight.hints.ProviderInfo;
+import com.intellij.codeInsight.hints.declarative.InlayProviderInfo;
 import com.intellij.lang.Language;
 import com.intellij.lang.findUsages.EmptyFindUsagesProvider;
 import com.intellij.lang.findUsages.LanguageFindUsages;
@@ -63,6 +63,8 @@ public class LanguageServersRegistry {
             List<String> /* file extensions (ex : ts) */> languageIdFileExtensionsCache = new HashMap<>();
 
     private final Collection<LanguageServerDefinitionListener> listeners = new CopyOnWriteArrayList<>();
+
+    private final Map<String, List<InlayProviderInfo>> declarativeInlayHintsProviders = new HashMap<>();
 
     private final List<ProviderInfo<? extends Object>> inlayHintsProviders = new ArrayList<>();
 
@@ -190,8 +192,10 @@ public class LanguageServersRegistry {
         // the language received in InlayHintProviders is plain/text, we add it to support
         // LSP inlayHint, color for a file which is not linked to a language.
         distinctLanguages.add(PlainTextLanguage.INSTANCE);
-
-        // register LSPInlayHintsProvider + LSPColorProvider automatically
+        // register LSPInlayHintsProvider automatically
+        // for all languages associated with a language server.
+        updateDeclarativeInlayHintsProviders(distinctLanguages);
+        // register LSPColorProvider automatically
         // for all languages associated with a language server.
         updateInlayHintsProviders(distinctLanguages);
         // register LSPFindUsagesProvider automatically
@@ -199,13 +203,21 @@ public class LanguageServersRegistry {
         updateFindUsagesProvider(distinctLanguages);
     }
 
-    private void updateInlayHintsProviders(Set<Language> distinctLanguages) {
+    private void updateDeclarativeInlayHintsProviders(Set<Language> distinctLanguages) {
         LSPInlayHintsProvider lspInlayHintsProvider = new LSPInlayHintsProvider();
+        inlayHintsProviders.clear();
+        for (Language language : distinctLanguages) {
+            List<InlayProviderInfo> hints = new ArrayList<>();
+            hints.add(new InlayProviderInfo(lspInlayHintsProvider, LSPInlayHintsProvider.PROVIDER_ID, Collections.emptySet(), true, LanguageServerBundle.message("lsp.hints.declarative.provider.name")));
+            declarativeInlayHintsProviders.put(language.getID(), hints);
+        }
+    }
+
+    private void updateInlayHintsProviders(Set<Language> distinctLanguages) {
         LSPColorProvider lspColorProvider = new LSPColorProvider();
         inlayHintsProviders.clear();
         for (Language language : distinctLanguages) {
-            inlayHintsProviders.add(new ProviderInfo<NoSettings>(language, lspInlayHintsProvider));
-            inlayHintsProviders.add(new ProviderInfo<NoSettings>(language, lspColorProvider));
+            inlayHintsProviders.add(new ProviderInfo<>(language, lspColorProvider));
         }
     }
 
@@ -571,12 +583,16 @@ public class LanguageServersRegistry {
     }
 
     /**
-     * Returns the LSP codeLens / inlayHint inlay hint providers for all languages which are associated with a language server.
-     *
-     * @return the LSP codeLens / inlayHint inlay hint providers for all languages which are associated with a language server.
+     * @return the LSP codeLens / color inlay hint providers for all languages which are associated with a language server.
      */
     public List<ProviderInfo<? extends Object>> getInlayHintProviderInfos() {
         return inlayHintsProviders;
+    }
+    /**
+     * @return the LSP inlayHint inlay hint providers for all languages which are associated with a language server.
+     */
+    public Map<String, List<InlayProviderInfo>> getDeclarativeInlayHintProviderInfos() {
+        return declarativeInlayHintsProviders;
     }
 
     public record UpdateServerDefinitionRequest(@NotNull Project project,
--- a/src/main/java/com/redhat/devtools/lsp4ij/features/AbstractLSPInlayHintsProvider.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/features/AbstractLSPInlayHintsProvider.java
@@ -7,6 +7,7 @@
  *
  * Contributors:
  * Red Hat, Inc. - initial API and implementation
+ * FalsePattern - fixed duplicate inlay hints
  ******************************************************************************/
 package com.redhat.devtools.lsp4ij.features;
 
@@ -65,22 +66,15 @@ public abstract class AbstractLSPInlayHintsProvider implements InlayHintsProvide
         final long modificationStamp = psiFile.getModificationStamp();
         return new FactoryInlayHintsCollector(editor) {
 
-            private boolean processed;
-
             @Override
             public boolean collect(@NotNull PsiElement psiElement, @NotNull Editor editor, @NotNull InlayHintsSink inlayHintsSink) {
-                if (processed) {
-                    // Before IJ 2023-3, FactoryInlayHintsCollector#collect(PsiElement element.. is called once time with PsiFile as element.
-                    // Since IJ 2023-3, FactoryInlayHintsCollector#collect(PsiElement element.. is called several times for each token of the PsiFile
-                    // which causes the problem of codelens/inlay hint which are not displayed because there are too many call of LSP request codelens/inlayHint which are cancelled.
-                    // With IJ 2023-3 we need to collect LSP CodeLens/InlayHint just for the first call.
-                    return false;
+                if (!(psiElement instanceof PsiFile psiFile)) {
+                    return true;
                 }
-                processed = true;
                 Project project = psiFile.getProject();
                 if (project.isDisposed()) {
                     // InlayHint must not be collected
-                    return false;
+                    return true;
                 }
 
                 try {
@@ -105,7 +99,7 @@ public abstract class AbstractLSPInlayHintsProvider implements InlayHintsProvide
                 } catch (CancellationException e) {
                     // Do nothing
                 }
-                return false;
+                return true;
             }
         };
     }
--- a/src/main/java/com/redhat/devtools/lsp4ij/features/AbstractLSPInlayHintsProvider.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/features/AbstractLSPInlayHintsProvider.java
@@ -78,23 +78,13 @@ public abstract class AbstractLSPInlayHintsProvider implements InlayHintsProvide
                 }
 
                 try {
-                    final List<CompletableFuture> pendingFutures = new ArrayList<>();
+                    final List<CompletableFuture<?>> pendingFutures = new ArrayList<>();
                     doCollect(psiFile, editor, getFactory(), inlayHintsSink, pendingFutures);
                     if (!pendingFutures.isEmpty()) {
                         // Some LSP requests:
-                        // - textDocument/inlayHint, inlayHint/resolve
                         // - textDocument/colorInformation
                         // are pending, wait for their completion and refresh the inlay hints UI to render them
-                        CompletableFuture.allOf(pendingFutures.toArray(new CompletableFuture[0]))
-                                .thenApplyAsync(_unused -> {
-                                    // Check if PsiFile was not modified
-                                    if (modificationStamp == psiFile.getModificationStamp()) {
-                                        // All pending futures are finished, refresh the inlay hints
-                                        EditorFeatureManager.getInstance(project)
-                                                        .refreshEditorFeature(psiFile.getVirtualFile(), EditorFeatureType.INLAY_HINT, false);
-                                    }
-                                    return null;
-                                });
+                        EditorFeatureManager.getInstance(project).refreshEditorFeatureWhenAllDone(pendingFutures, modificationStamp, psiFile, EditorFeatureType.INLAY_HINT);
                     }
                 } catch (CancellationException e) {
                     // Do nothing
@@ -171,6 +161,6 @@ public abstract class AbstractLSPInlayHintsProvider implements InlayHintsProvide
                                       @NotNull Editor editor,
                                       @NotNull PresentationFactory factory,
                                       @NotNull InlayHintsSink inlayHintsSink,
-                                      @NotNull List<CompletableFuture> pendingFutures);
+                                      @NotNull List<CompletableFuture<?>> pendingFutures);
 
 }
--- a/src/main/java/com/redhat/devtools/lsp4ij/features/inlayhint/LSPInlayHintsProvider.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/features/inlayhint/LSPInlayHintsProvider.java
@@ -7,13 +7,11 @@
  *
  * Contributors:
  * Red Hat, Inc. - initial API and implementation
+ * FalsePattern - port to declarative inlay hint API
  ******************************************************************************/
 package com.redhat.devtools.lsp4ij.features.inlayhint;
 
-import com.intellij.codeInsight.hints.InlayHintsSink;
-import com.intellij.codeInsight.hints.presentation.InlayPresentation;
-import com.intellij.codeInsight.hints.presentation.PresentationFactory;
-import com.intellij.codeInsight.hints.presentation.SequencePresentation;
+import com.intellij.codeInsight.hints.declarative.*;
 import com.intellij.openapi.editor.Document;
 import com.intellij.openapi.editor.Editor;
 import com.intellij.openapi.progress.ProcessCanceledException;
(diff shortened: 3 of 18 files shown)


angelozerr:
> ## Goals
> - Completely remove the dependency on the deprecated inlay hints API.
> - Migrate inlay hints support to the declarative API
> 
> ## Fixes
> - https://github.com/redhat-developer/lsp4ij/issues/606, primary goal of this PR
> - https://github.com/redhat-developer/lsp4ij/issues/382, as a side effect of the rendering being completely delegated to the intellij gui backend 
> image
> 
> 
> ## Current draft regressions
> - Inlay hint hover/click actions are not yet implemented
> - `textDocument/color` support is impossible through the declarative API due to its limitations, it only supports text nodes, and "trees" of text nodes
> As a workaround, color hints should be migrated to the gutter bar, similarly to how intellij does color hints for `java.awt.Color` instances:
> image
> 
> 

Your pr looks promising. The click on inlay hint must be implemented because we use it a lot in our IJ quarkus.

For color information it is little annoying. I wonder how the native lsp support manage color?

FalsePattern:
From what i could gather, the intellij build-in color hints use the deprecated inlay hints api, and it seems to use a slightly different approach to avoid duplicates, via a custom cache.

I can split the inlay hints and the color hints so that the colors use the old api, while the text hints use the declarative api

FalsePattern:
I think i've found the root cause of the duplicate inlay hints, I will update the PR in a bit.

angelozerr:
> From what i could gather from decompiling IdeaU, for color hints they use the deprecated inlay hints api, and it seems to use a slightly different approach to avoid duplicates, via a custom cache.
> 

Hum it can be dangerous to do that and I dont want to have some trouble with Jetbrains. If you have a fix it is super nice but please dont do the same thing than Jetbrains.

> I can split the inlay hints and the color hints so that the colors use the old api, while the text hints use the declarative api

Indeed it was also my idea.

And what about click on inlay hint? If the new api inlay hint cannot support that, I fear that we cannot use it.

angelozerr:
> I think i've found the root cause of the duplicate inlay hints in the original implementation, I will update the PR in a bit. 

Great! Thanks so much.

FalsePattern:
I've split the declarative<->legacy implementations, using the legacy one for colors while using the declarative one for the colors. I've avoided referencing intellij code for it.

I also managed to implement clickable inlay hints via the new api by cheating a little and abusing object references and a weak hash map, unfortunately there is no way to pass the full context needed to the click handler logic by normal means. See: https://github.com/FalsePattern/lsp4ij/blob/main/src/main/java/com/redhat/devtools/lsp4ij/features/LSPDeclarativeInlayActionHandler.java and https://github.com/FalsePattern/lsp4ij/blob/main/src/main/java/com/redhat/devtools/lsp4ij/features/inlayhint/LSPInlayHintsProvider.java#L142-L144

src/main/java/com/redhat/devtools/lsp4ij/features/LSPDeclarativeInlayActionHandler.java
```diff
@@ -0,0 +1,31 @@
...
+
+import java.util.Map;
+import java.util.function.Consumer;
+
+public class LSPDeclarativeInlayActionHandler implements InlayActionHandler {
+    //There is no way to send complex data through inlay action payloads, so we cheat a little.
+    public static Map<InlayActionPayload, Consumer<Editor>> callbacks = new MapMaker().weakKeys().makeMap();
+
+    public static InlayActionPayload createAndRegisterPayload(String text, Consumer<Editor> callback) {
+        var payload = new StringInlayActionPayload(text);
```
angelozerr, line 18: What about extends StringInlayActionPayLoad to stores the callback and avoid using weak map?

FalsePattern: StringInlayActionPayload is a final class (non-open kotlin class), InlayActionPayload is a sealed interface, and all other classes extending InlayActionPayload are also final classes

FalsePattern: We *could* use a dynamic proxy style way to create a custom subclass of InlayActionPayload during runtime without tripping the marketplace verifier

angelozerr: I don't know how to that and if it is clean?

Main problem is callback will grow up a lot and never clean.

What about using a simple Map stored in the Editor#putUserData?

angelozerr: @SCWells72 have you some better idea?

FalsePattern: Fixed it in the latest push, I'm hijacking the `PsiPointerInlayActionPayload` with a custom subclass of `SmartPsiElementPointer` and using it as the callback carrier

FalsePattern:
All the issues caused by the migration should be resolved by now. At this point it's just the color inlay hints that need to be verified that they work, i validated the duplicate inlay hint and the clickable command inlay hints locally.

angelozerr:
> All the issues caused by the migration should be resolved by now. At this point it's just the color inlay hints that need to be verified that they work, i validated the duplicate inlay hint and the clickable command inlay hints locally.

Thanks so much, let me review it and test it.

If you have vscode, you can easily test the CSS support https://github.com/redhat-developer/lsp4ij/blob/main/docs/user-defined-ls/vscode-css-language-server.md

angelozerr:
> Works image

Indeed I have tested and it works.

src/main/java/com/redhat/devtools/lsp4ij/features/LSPDeclarativeInlayActionHandler.java
```diff
@@ -0,0 +1,72 @@
+package com.redhat.devtools.lsp4ij.features;
```
angelozerr, line 1: Add header

angelozerr, line 1: Move this class to `com.redhat.devtools.lsp4ij.features.inlayhint`

src/main/java/com/redhat/devtools/lsp4ij/features/LSPDeclarativeInlayActionHandler.java
```diff
@@ -0,0 +1,72 @@
...
+import com.intellij.openapi.vfs.VirtualFile;
+import com.intellij.psi.PsiElement;
+import com.intellij.psi.PsiFile;
+import com.intellij.psi.SmartPsiElementPointer;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.function.Consumer;
+
+public class LSPDeclarativeInlayActionHandler implements InlayActionHandler {
```
angelozerr, line 18: Add javadoc

src/main/java/com/redhat/devtools/lsp4ij/features/LSPDeclarativeInlayHintProvidersFactory.java
```diff
@@ -0,0 +1,51 @@
...
+ * Copyright (c) 2023 Red Hat, Inc.
+ * Distributed under license by Red Hat, Inc. All rights reserved.
+ * This program is made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution,
+ * and is available at http://www.eclipse.org/legal/epl-v20.html
+ *
+ * Contributors:
+ * Red Hat, Inc. - initial API and implementation
+ ******************************************************************************/
+package com.redhat.devtools.lsp4ij.features;
```
angelozerr, line 11: Move this class to `com.redhat.devtools.lsp4ij.features.inlayhint`

src/main/java/com/redhat/devtools/lsp4ij/features/LSPDeclarativeInlayHintProvidersFactory.java
```diff
@@ -0,0 +1,51 @@
+/*******************************************************************************
+ * Copyright (c) 2023 Red Hat, Inc.
```
angelozerr, line 2: 2025

src/main/java/com/redhat/devtools/lsp4ij/features/LSPDeclarativeInlayHintProvidersFactory.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.stream.Collectors;
+
+/**
+ * {@link InlayHintsProviderFactory inlay hint factory} implementation
+ * to register all languages mapped with a language server with {@link LSPInlayHintsProvider}
+ * to avoid for the external plugin to declare in plugin.xml the 'codeInsight.inlayProvider'.
+ */
+public class LSPDeclarativeInlayHintProvidersFactory implements InlayHintsProviderFactory {
+    @Override
+    public @Nullable InlayProviderInfo getProviderInfo(@NotNull Language language, @NotNull String s) {
```
angelozerr, line 34: Rename `s` with `providerId`

src/main/java/com/redhat/devtools/lsp4ij/features/inlayhint/LSPInlayHintsProvider.java
```diff
@@ -114,47 +115,41 @@ private static Range getViewPortRange(Editor editor) {
...
         elements.forEach(p -> {
             Either<String, List<InlayHintLabelPart>> label = p.second.inlayHint().getLabel();
             if (label.isLeft()) {
-                presentations.add(factory.smallText(label.getLeft()));
+                builder.text(label.getLeft(), null);
             } else {
                 int index = 0;
                 for (InlayHintLabelPart part : label.getRight()) {
-                    InlayPresentation text = createInlayPresentation(psiFile, editor, factory, p, index, part);
-                    if (part.getTooltip() != null && part.getTooltip().isLeft()) {
```
angelozerr, line 130: What about tooltip?

src/main/java/com/redhat/devtools/lsp4ij/LanguageServersRegistry.java
```diff
@@ -190,22 +193,32 @@ private void updateLanguages() {
...
         updateFindUsagesProvider(distinctLanguages);
     }
 
-    private void updateInlayHintsProviders(Set<Language> distinctLanguages) {
+    private void updateDeclarativeInlayHintsProviders(Set<Language> distinctLanguages) {
         LSPInlayHintsProvider lspInlayHintsProvider = new LSPInlayHintsProvider();
+        inlayHintsProviders.clear();
+        for (Language language : distinctLanguages) {
+            List<InlayProviderInfo> hints = new ArrayList<>();
+            hints.add(new InlayProviderInfo(lspInlayHintsProvider, "LSP4IJ", Collections.emptySet(), true, "Language Server"));
```
angelozerr, line 212: Replace `LSP4IJ` with `LSPInlayHintsProvider.PROVIDER_ID`

src/main/java/com/redhat/devtools/lsp4ij/features/AbstractLSPDeclarativeInlayHintsProvider.java
```diff
@@ -0,0 +1,101 @@
...
+ * Copyright (c) 2022 Red Hat, Inc.
+ * Distributed under license by Red Hat, Inc. All rights reserved.
+ * This program is made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution,
+ * and is available at http://www.eclipse.org/legal/epl-v20.html
+ *
+ * Contributors:
+ * Red Hat, Inc. - initial API and implementation
+ ******************************************************************************/
+package com.redhat.devtools.lsp4ij.features;
```
angelozerr, line 11: move to `...features.inlayhint` package.

src/main/java/com/redhat/devtools/lsp4ij/features/inlayhint/LSPInlayHintsProvider.java
```diff
@@ -45,16 +47,16 @@
 /**
  * LSP textDocument/inlayHint support.
  */
-public class LSPInlayHintsProvider extends AbstractLSPInlayHintsProvider {
+public class LSPInlayHintsProvider extends AbstractLSPDeclarativeInlayHintsProvider {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(LSPInlayHintsProvider.class);
 
+
```
angelozerr, line 54: define a constant

`public static final String LSPInlayHintsProvider.PROVIDER_ID = "LSPInlayHintsProvider";`

src/main/java/com/redhat/devtools/lsp4ij/features/inlayhint/LSPInlayHintsProvider.java
```diff
@@ -114,47 +115,41 @@ private static Range getViewPortRange(Editor editor) {
...
         if (hasCommand(part)) {
             // InlayHintLabelPart defines a Command, create a clickable inlay hint
-            int finalIndex = index;
-            text = factory.referenceOnHover(text, (event, translated) ->
-                    executeCommand(psiFile, editor, p.second.languageServer(), p.second.inlayHint(), finalIndex, event)
-            );
+            InlayActionPayload payload = LSPDeclarativeInlayActionHandler.createPayload(psiFile.getProject(), editor -> {
+                executeCommand(psiFile, editor, p.second.languageServer(), p.second.inlayHint(), index);
+            });
+            InlayActionData data = new InlayActionData(payload, "LSP4IJ");
```
angelozerr, line 148: `new InlayActionData(payload, LSPInlayHintsProvider.PROVIDER_ID);`

src/main/java/com/redhat/devtools/lsp4ij/internal/CompletableFutures.java
```diff
@@ -186,4 +189,20 @@ public void run(@NotNull ProgressIndicator indicator) {
         });
     }
 
+    public static void refreshEditorFeatureWhenAllDone(@NotNull List<CompletableFuture<?>> pendingFutures,
```
angelozerr, line 192: Move this method in `EditorFeatureManager` and add javadoc

src/main/java/com/redhat/devtools/lsp4ij/internal/editor/DeclarativeInlayHintsEditorFeature.java
```diff
@@ -0,0 +1,62 @@
...
+    public void clearEditorCache(@NotNull Editor editor, @NotNull Project project) {
+        DeclarativeInlayHintsPassFactory.Companion.scheduleRecompute(editor, project);
+    }
+
+    @Override
+    public void clearLSPCache(PsiFile file) {
+        // Evict the cache of LSP requests from inlayHint and color support
+        var fileSupport = LSPFileSupport.getSupport(file);
+        fileSupport.getInlayHintsSupport().cancel();
+        fileSupport.getColorSupport().cancel();
```
angelozerr, line 49: Remove this line code and move to `InlayHintsEditorFeature`

src/main/java/com/redhat/devtools/lsp4ij/LanguageServersRegistry.java
```diff
@@ -190,22 +192,32 @@ private void updateLanguages() {
...
         updateFindUsagesProvider(distinctLanguages);
     }
 
-    private void updateInlayHintsProviders(Set<Language> distinctLanguages) {
+    private void updateDeclarativeInlayHintsProviders(Set<Language> distinctLanguages) {
         LSPInlayHintsProvider lspInlayHintsProvider = new LSPInlayHintsProvider();
+        inlayHintsProviders.clear();
+        for (Language language : distinctLanguages) {
+            List<InlayProviderInfo> hints = new ArrayList<>();
+            hints.add(new InlayProviderInfo(lspInlayHintsProvider, LSPInlayHintsProvider.PROVIDER_ID, Collections.emptySet(), true, "Language Server"));
```
angelozerr, line 211: Use LanguageServerBundle.message('....) to avoid hard coding "Language Server" string

src/main/java/com/redhat/devtools/lsp4ij/features/AbstractLSPInlayHintsProvider.java
```diff
@@ -65,47 +66,30 @@ public final InlayHintsCollector getCollectorFor(@NotNull PsiFile psiFile,
...
             @Override
             public boolean collect(@NotNull PsiElement psiElement, @NotNull Editor editor, @NotNull InlayHintsSink inlayHintsSink) {
-                if (processed) {
-                    // Before IJ 2023-3, FactoryInlayHintsCollector#collect(PsiElement element.. is called once time with PsiFile as element.
-                    // Since IJ 2023-3, FactoryInlayHintsCollector#collect(PsiElement element.. is called several times for each token of the PsiFile
-                    // which causes the problem of codelens/inlay hint which are not displayed because there are too many call of LSP request codelens/inlayHint which are cancelled.
-                    // With IJ 2023-3 we need to collect LSP CodeLens/InlayHint just for the first call.
-                    return false;
+                if (!(psiElement instanceof PsiFile psiFile)) {
+                    return true;
```
angelozerr, line 72: Return true fixes the duplicate problem? Could you explain more how you have fixed duplicate inlay hints please.

FalsePattern: The `collect` method is called once per file, so the `processed` check can be replaced with a check for whether the `psiElement` is a `PsiFile`, and that will prevent the duplicate inlay hints bug.
Return true is there for consistency with other examples from open source intellij code (Apache 2.0), which always returns true, see: https://github.com/JetBrains/intellij-community/blob/9d33ede5a92b3d533fe7c1a906d6a425067902b0/plugins/markdown/core/src/org/intellij/plugins/markdown/editor/tables/ui/MarkdownTableInlayProvider.kt#L30

angelozerr:
@FalsePattern I'am testing your PR with our IJ Quarkus. When you open the first file, inlay hint doesn't appear. It is because our language server register inlay hint with dynamic capability which must refresh the editor

The root cause problem is https://github.com/redhat-developer/lsp4ij/blob/c8cffa0b1821285a5dbead162658ee13b17c462e/src/main/java/com/redhat/devtools/lsp4ij/server/capabilities/InlayHintCapabilityRegistry.java#L47

You need to change with 

```java
public InlayHintCapabilityRegistry(@NotNull LSPClientFeatures clientFeatures) {
        super(clientFeatures, EditorFeatureType.DECLARATIVE_INLAY_HINT );
    }
```

angelozerr:
In IJ Quarkus we have the following inlay hints which is composed with 2 parts:

```
[Trace - 10:09:02] Received response 'textDocument/inlayHint - (12)' in 6ms.
Result: [
  {
    "position": {
      "line": 0,
      "character": 11
    },
    "label": [
      {
        "value": ":"
      },
      {
        "value": "HttpServerRequest",
        "tooltip": "Open `io.vertx.core.http.HttpServerRequest` Java type.",
        "command": {
          "title": "Open `io.vertx.core.http.HttpServerRequest` Java type.",
          "command": "qute.command.java.definition",
          "arguments": [
            {
              "projectUri": "code-with-quarkus-web",
              "sourceType": "io.vertx.core.http.HttpServerRequest",
              "dataMethodInvocation": false
            }
          ]
        }
      }
    ],
    "kind": 1
  },
  {
    "position": {
      "line": 3,
      "character": 10
    },
    "label": [
      {
        "value": ":"
      },
      {
        "value": "String",
        "tooltip": "Open `java.lang.String` Java type.",
        "command": {
          "title": "Open `java.lang.String` Java type.",
          "command": "qute.command.java.definition",
          "arguments": [
            {
              "projectUri": "code-with-quarkus-web",
              "sourceType": "java.lang.String",
              "dataMethodInvocation": false
            }
          ]
        }
      }
    ],
    "kind": 1
  }
]

```

With your PR, it adds an extra space. Without your PR the inlay hint are merged (no extra space).

image

FalsePattern:
Should be fixed in the latest push. Unfortunately, declarative inlay hints don't support per-label tooltips, only a single tooltip for the whole inlay hint, so I merge the tooltip of each label inside a given inlay hint with newlines between them.

FalsePattern:
Also, note that intellij 2023.2 does not render hover text for declarative inlay hints, but it does in 2024.3, so that feature was added somewhere along the way

angelozerr:
Thanks! Let me try again.

I wonder if we could support paddingRight with this new InlayHint API ?

FalsePattern:
> I wonder if we could support paddingRight with this new InlayHint API ?
It's doable by adding some extra background-less inlay hints before/after the inlay hint itself, but the cursor can then land between these "padding hints", which might not be desirable
code:
image

results:
image
image
image

angelozerr:
Wow it is working like a charm!

image


Congrats @FalsePattern !

PR #802, opened 2025-02-04
fix: Fixed a problem with how LSP positions are converted into editor offsets when the position line is 0 and the position character is the pure offset into the document (fixes #798)

### LINKED ISSUE #798: Support language servers that report positions as pure offsets
I've encountered one language server that reports range position information as a pure offset into the file instead of as line/character, e.g.:

```json
{
  "uri": "file://...",
  "diagnostics": [
    {
      "range": {
        "start": {
          "line": 0,
          "character": 746
        },
        "end": {
          "line": 0,
          "character": 762
        }
      },
      "severity": 3,
      "code": "...",
      "source": "...",
      "message": "..."
    }
  ]
}
```

Right now `LSPIJUtils#toOffset()` tries to be defensive in the conversion from position to offset and ends up losing this valid way of calculating an offset from `line=0, character=offset`.

It's obviously an easy fix, but we'll want to make sure that there isn't anything downstream that is depending on the current behavior, or at least for those things that are, we'll want to address those dependencies.

SCWells72:
I've run into at least one language server, specifically [Salesforce's SLDS Validator LSP](https://github.com/forcedotcom/salesforcedx-vscode-slds/blob/master/lsp-2.0.12-executable.jar), that returns positions with no line number and a character that represents the pure offset into the document, e.g.:

```json
{
  "uri": "file://...",
  "diagnostics": [
    {
      "range": {
        "start": {
          "line": 0,
          "character": 746
        },
        "end": {
          "line": 0,
          "character": 762
        }
      },
      "severity": 3,
      "code": "...",
      "source": "...",
      "message": "..."
    }
  ]
}
```

The existing implementation of `LSPIJUtils#toOffset` was trying to guard against invalid offsets by constraining the result to the specified line. This PR includes a client config-based fix for that so that when a position of the form `[0,offset]` is returned, it is translated into `offset`, e.g.:

image


src/main/java/com/redhat/devtools/lsp4ij/LSPIJUtils.java
```diff
@@ -595,16 +596,16 @@ public static int toOffset(int line, int character, @NotNull Document document)
...
             return document.getTextLength();
         } else if (line < 0) {
             // The line number is negative, it defaults to 0.
             return 0;
         }
-        int lineOffset = document.getLineStartOffset(line);
-        int nextLineOffset = document.getLineEndOffset(line);
-        // If the character value is greater than the line length it defaults back to the line length
-        return Math.max(Math.min(lineOffset + character, nextLineOffset), lineOffset);
+        int lineStartOffset = document.getLineStartOffset(line);
```
SCWells72, line 605: This is the actual fix. Just convert the position into `lineStartOffset + column` and then constrain to the allowed boundaries of the document if necessary.

src/test/java/com/redhat/devtools/lsp4ij/LSPIJUtils_toOffsetTest.java
```diff
@@ -62,7 +63,7 @@ public void testInvalidCharacterWithOneLine() {
     }
 
     public void testInvalidCharacterWithTwoLines() {
-        assertOffset("foo\nbar", 0, 999999, 3, null);
+        assertOffset("foo\nbar", 0, 999999, 7, null);
```
SCWells72, line 66: Had to change this expected value because it's no longer constrained to the line but rather to the document.

src/test/java/com/redhat/devtools/lsp4ij/LSPIJUtils_toOffsetTest.java
```diff
@@ -74,6 +75,20 @@ public void testInvalidPositionWithTwoLine() {
...
     }
 
+    public void testPureOffsetPositions() {
+        String content = "foo\nbar\nbaz";
+        for (int character = 0; character < content.length(); character++) {
+            assertOffset(content, 0, character, character, content.charAt(character));
+        }
+    }
+
+    public void testInvalidPureOffsetPositions() {
```
SCWells72, line 85: Negative test that confirms that `line=0, column=<invalid>` results in the respective bounds of the content.

src/main/java/com/redhat/devtools/lsp4ij/LSPIJUtils.java
```diff
@@ -595,16 +596,16 @@ public static int toOffset(int line, int character, @NotNull Document document)
 
         // Adjust position line/character according to this comment https://github.com/microsoft/vscode-languageserver-node/blob/ed3cd0f78c1495913bda7318ace2be7f968008af/textDocument/src/main.ts#L26
         if (line >= document.getLineCount()) {
-            // The line number is greater than the number of lines in a document, it defaults back to the number of lines in the document.
+            // The line number is greater than the number of lines in a document, it defaults to the last valid offset in the document.
```
SCWells72, line 599: Just updated this comment to properly reflect the behavior.

src/test/java/com/redhat/devtools/lsp4ij/features/completion/QuteCompletionTest.java
```diff
@@ -24,70 +25,73 @@ public QuteCompletionTest() {
...
+                              "insertTextMode": 2,
+                              "textEdit": {
+                                "range": {
+                                  "start": {
+                                    "line": 1,
+                                    "character": 4
+                                  },
+                                  "end": {
+                                    "line": 1,
+                                    "character": 5
```
SCWells72, line 59: This was straight from the response captured in the LSP4IJ debug log. Not sure why it would have been `6` before for the same input, but `6` was failing with my fix. That's why I set up the Qute language server...so I could capture a new response and see if it was a mock data issue or a code issue. It turns out it was a mock data issue that is corrected here.

src/test/java/com/redhat/devtools/lsp4ij/features/completion/QuteCompletionTest.java
```diff
@@ -24,70 +25,73 @@ public QuteCompletionTest() {
...
+                                "newText": "{#if ${1:condition}}\\n\\t$0\\n{/if}"
+                              }
+                            },
+                            {
+                              "label": "for",
+                              "kind": 15,
+                              "detail": "Loop section with alias",
+                              "documentation": {
+                                "kind": "markdown",
+                                "value": "\\r\\n```\\r\\n{#for item in items}\\n\\t{item.name}\\n{/for}\\r\\n```\\r\\n\\r\\n\\r\\nSee [Loop section](https://quarkus.io/guides/qute-reference#loop_section) for more information."
```
SCWells72, line 71: It also looks like this was hand-edited from the original response because quite a bit of the required escaping was lost in the original version but has been restored properly here.

src/test/java/com/redhat/devtools/lsp4ij/features/completion/QuteCompletionTest.java
```diff
@@ -24,70 +25,73 @@ public QuteCompletionTest() {
...
+                              "insertTextMode": 2,
+                              "textEdit": {
+                                "range": {
+                                  "start": {
+                                    "line": 1,
+                                    "character": 4
+                                  },
+                                  "end": {
+                                    "line": 1,
+                                    "character": 5
```
SCWells72, line 84: Same here on the change from `6` originally to `5` from the newly-captured response.

src/test/java/com/redhat/devtools/lsp4ij/features/completion/QuteCompletionTest.java
```diff
@@ -24,70 +25,73 @@ public QuteCompletionTest() {
...
+                                  }
+                                },
+                                "newText": "{#for ${1:item} in ${2:items}}\\n\\t{${1:item}.${3:name}}$0\\n{/for}"
```
SCWells72, line 87: And again, some missing required escaping here.

SCWells72:
@angelozerr, I know this is coming in late for 0.10.0, but it seems like a good fix to include if you can get comfortable with it.

src/test/java/com/redhat/devtools/lsp4ij/LSPIJUtils_toOffsetTest.java
angelozerr: Mmmm this test comes from vscode if I rember.

SCWells72:
> Yes but do you know if there is a spec for that?

Hmmmm...good question, and the answer is of course "yes":

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#position

And this is probably the most relevant part:

> If the character value is greater than the line length it defaults back to the line length.

I can also see where the original LSP4IJ implementation is pretty much an exact port of the corresponding logic in VS Code:

https://github.com/microsoft/vscode-languageserver-node/blob/8e625564b531da607859b8cb982abb7cdb2fbe2e/textDocument/src/main.ts#L304

What I can't explain, however, is why this language server works properly in VS Code but not in LSP4IJ if they're translating `Position` to `offset` using the exact same logic.

Given that, I can certainly understand if you're wanting to leave this out. Perhaps this makes an even stronger case for me integrating this specific language server as an extension of LSP4IJ so I can change the reported `Position` values to be proper `line/character` pairs instead of `0/offset` that's being reported.

angelozerr:
> Given that, I can certainly understand if you're wanting to leave this out.

Yes please, after the release if you wish.

angelozerr:
If vscode supports this usecase, I think LSP4IJ should support it.

SCWells72:
> If vscode supports this usecase, I think LSP4IJ should support it.

100% agreed. I just don't understand how/why it supports this use case given the VS Code implementation of `offsetAt()` linked above.

SCWells72:
FYI, I've logged an issue against this specific language server and reached out to some of my internal contacts at Salesforce to see if someone on that team can provide insights into this unexpected and seemingly out-of-spec behavior:

https://github.com/forcedotcom/salesforcedx-vscode-slds/issues/141

Hopefully someone will weigh in soon, but it's very possible that no one will. If no one does in short order -- or if someone does and just says "that's how it works" -- I'd like to discuss acceptable/safe options for addressing this if possible.

src/test/java/com/redhat/devtools/lsp4ij/LSPIJUtils_toOffsetTest.java
angelozerr: The specs says 

`If the character value is greater than the line length it defaults back to the line length. `

https://github.com/microsoft/vscode-languageserver-node/blob/d810d51297c667bd3a3f46912eb849055beb8b6b/textDocument/src/main.ts#L40C5-L40C7

This test set line to 0 and in this case line length is equals to 3, right?

I have not found a test with this usecase. But perhaps the spec is not clear and we should  adopt your fix. The thing like I love in your PR i sthat you avoid calling getLineEndOffset.

I am not against to merge your PR,but before could you create an issue https://github.com/microsoft/vscode-languageserver-node/ with this usecase toclarify this usecase please.

SCWells72: I logged this to try to track down how this is working:

https://github.com/forcedotcom/salesforcedx-vscode-slds/issues/141

docs/LSPApi.md
```diff
@@ -60,6 +60,7 @@ public class MyLanguageServerFactory implements LanguageServerFactory {
...
 | boolean isCaseSensitive(PsiFile file)                        | Returns `true` if the language grammar for the given file is case-sensitive and `false` otherwise.                                                                                                                                                                                                                                                                                             | `false`           | 
+| boolean serverReportsOffsetOnlyPositions(PsiFile file)       | Returns `true` if the language server reports `Position` values as pure offsets and `false` otherwise.                                                                                                                                                                                                                                                                                         | `false`           |
```
SCWells72, line 63: This is the doc for the new client config option to tell LSP4IJ that this specific server reports `Position` values that are of the form `[0,offset]`.

src/main/java/com/redhat/devtools/lsp4ij/LSPIJUtils.java
```diff
@@ -595,12 +597,26 @@ public static int toOffset(int line, int character, @NotNull Document document)
...
             return document.getTextLength();
         } else if (line < 0) {
             // The line number is negative, it defaults to 0.
             return 0;
         }
+        // If the line number is 0 and the character is beyond the length of the first line, see if the language server
```
SCWells72, line 606: The new logic is only engaged if the provided `Position` has `line=0` and `character>lines[0].length`, and only then if the client config feature flag `serverReportsOffsetOnlyPositions` is enabled. Otherwise it is the exact same logic as before that was ported from VS Code.

src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPClientFeatures.java
```diff
@@ -154,6 +154,17 @@ public boolean isCaseSensitive(@NotNull PsiFile file) {
         return false;
     }
 
+    /**
+     * Determines whether or not the language server reports {@link org.eclipse.lsp4j.Position} values as pure offsets.
+     *
+     * @param file the file
+     * @return true if the file's language server reports offset-only positions; otherwise false
+     */
+    public boolean serverReportsOffsetOnlyPositions(@NotNull PsiFile file) {
```
SCWells72, line 163: The feature flag defaults to disabled.

src/main/java/com/redhat/devtools/lsp4ij/server/definition/launching/ClientConfigurationSettings.java
```diff
@@ -137,6 +137,15 @@ public static class ClientConfigurationWorkspaceSymbolSettings {
...
     public boolean caseSensitive = false;
 
+    /**
+     * Whether or not the language server reports {@link org.eclipse.lsp4j.Position} values as pure offsets, i.e.,
+     * <code>line=0, character=offset</code>. This is technically a deviation from the
+     * <a href="https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#position">specification</a>,
+     * but the behavior has been <a href="https://github.com/forcedotcom/salesforcedx-vscode-slds/issues/141">observed</a>
+     * in at least one language server.
+     */
+    public boolean serverReportsOffsetOnlyPositions = false;
```
SCWells72, line 147: This is the new client configuration option.

src/main/java/com/redhat/devtools/lsp4ij/server/definition/launching/UserDefinedClientFeatures.java
```diff
@@ -32,8 +33,15 @@ public UserDefinedClientFeatures() {
     }
 
     public boolean isCaseSensitive(@NotNull PsiFile file) {
-        UserDefinedLanguageServerDefinition serverDefinition = (UserDefinedLanguageServerDefinition) getServerDefinition();
+        ClientConfigurableLanguageServerDefinition serverDefinition = (ClientConfigurableLanguageServerDefinition) getServerDefinition();
```
SCWells72, line 36: Same change as in #804 to change this to assume client-configurable instead of user-defined to support testing.

src/main/java/com/redhat/devtools/lsp4ij/server/definition/launching/UserDefinedClientFeatures.java
```diff
@@ -32,8 +33,15 @@ public UserDefinedClientFeatures() {
...
         ClientConfigurationSettings clientConfiguration = serverDefinition.getLanguageServerClientConfiguration();
-        return (clientConfiguration != null) && clientConfiguration.caseSensitive;
+        return clientConfiguration != null ? clientConfiguration.caseSensitive : super.isCaseSensitive(file);
+    }
+
+    @Override
+    public boolean serverReportsOffsetOnlyPositions(@NotNull PsiFile file) {
```
SCWells72, line 42: And in client-configurable language server definitions, we try to derive the value for this setting from client config.

src/main/resources/jsonSchema/clientSettings.schema.json
```diff
@@ -11,6 +11,12 @@
       "description": "Whether or not the language grammar is case-sensitive.",
       "default": false
     },
+    "serverReportsOffsetOnlyPositions": {
```
SCWells72, line 14: Client config JSON schema update.

src/test/java/com/redhat/devtools/lsp4ij/LSPIJUtils_toOffsetTest.java
```diff
@@ -8,6 +8,7 @@
  * Contributors:
  * Red Hat, Inc. - initial API and implementation
  ******************************************************************************/
+
```
SCWells72, line 11: Whitespace-only changes after moving the tests for this change into their own test class below.

src/test/java/com/redhat/devtools/lsp4ij/LSPIJUtils_toOffset_serverReportsOffsetOnlyPositionsTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * Tests for {@link LSPIJUtils#toOffset(Position, Document)} when
+ * {@link ClientConfigurationSettings#serverReportsOffsetOnlyPositions} is enabled.
+ */
+public class LSPIJUtils_toOffset_serverReportsOffsetOnlyPositionsTest extends LSPCodeInsightFixtureTestCase {
+
+    public LSPIJUtils_toOffset_serverReportsOffsetOnlyPositionsTest() {
+        super("*.txt");
+        setClientConfigurable(true);
```
SCWells72, line 36: Again, this is from the same change as in #804 to allow tests to specify themselves as client-configurable.

src/test/java/com/redhat/devtools/lsp4ij/LSPIJUtils_toOffset_serverReportsOffsetOnlyPositionsTest.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    @Override
+    protected void setUp() throws Exception {
```
SCWells72, line 40: Need to get the language server to get the client config to enable the feature flag.

src/test/java/com/redhat/devtools/lsp4ij/LSPIJUtils_toOffset_serverReportsOffsetOnlyPositionsTest.java
```diff
@@ -0,0 +1,80 @@
...
+        LanguageServerDefinition languageServerDefinition = languageServer.getServerDefinition();
+        assertInstanceOf(languageServerDefinition, ClientConfigurableLanguageServerDefinition.class);
+        ClientConfigurableLanguageServerDefinition configurableLanguageServerDefinition = (ClientConfigurableLanguageServerDefinition) languageServerDefinition;
+        ClientConfigurationSettings clientConfigurationSettings = configurableLanguageServerDefinition.getLanguageServerClientConfiguration();
+        assertNotNull(clientConfigurationSettings);
+
+        clientConfigurationSettings.serverReportsOffsetOnlyPositions = true;
+    }
+
+    public void testOffsetOnlyPositions() {
```
SCWells72, line 66: Confirms that `[0,offset]=>offset` for all valid values of `offset` in the document.

src/test/java/com/redhat/devtools/lsp4ij/LSPIJUtils_toOffset_serverReportsOffsetOnlyPositionsTest.java
```diff
@@ -0,0 +1,80 @@
...
+        Document document = myFixture.getDocument(myFixture.getFile());
+        for (int character = 0; character < document.getTextLength(); character++) {
+            int offset = LSPIJUtils.toOffset(0, character, document);
+            assertEquals(character, offset);
+        }
+    }
+
+    public void testInvalidOffsetOnlyPositions() {
```
SCWells72, line 74: Confirms that negative and overflow offsets are constrained to `[0,documentLength]`.

src/test/java/com/redhat/devtools/lsp4ij/fixtures/LSPClientSideOnTypeFormattingFixtureTestCase.java
```diff
@@ -53,6 +53,7 @@ public abstract class LSPClientSideOnTypeFormattingFixtureTestCase extends LSPCo
 
     public LSPClientSideOnTypeFormattingFixtureTestCase(String... fileNamePatterns) {
         super(fileNamePatterns);
+        setClientConfigurable(true);
```
SCWells72, line 56: Again, have to explicitly make this test client-configurable now.

src/test/java/com/redhat/devtools/lsp4ij/fixtures/LSPCodeInsightFixtureTestCase.java
```diff
@@ -28,6 +28,7 @@ public abstract class LSPCodeInsightFixtureTestCase extends UsefulTestCase {
 
     private final String[] fileNamePatterns;
     private String languageId;
+    private boolean clientConfigurable = false;
```
SCWells72, line 31: Same changes as in #804 to explicitly denote tests as including client config.

src/main/java/com/redhat/devtools/lsp4ij/LSPIJUtils.java
```diff
@@ -595,12 +597,26 @@ public static int toOffset(int line, int character, @NotNull Document document)
...
+        // for this document should treat such a situation as an offset-only position
+        else if ((line == 0) && (character > document.getLineEndOffset(line))) {
+            VirtualFile virtualFile = LSPIJUtils.getFile(document);
+            Project project = virtualFile != null ? ProjectLocator.getInstance().guessProjectForFile(virtualFile) : null;
+            PsiFile file = project != null ? getPsiFile(virtualFile, project) : null;
+            if ((file != null) && LanguageServiceAccessor.getInstance(project).hasAny(
```
angelozerr, line 612: LSPIJUtils#toOffset is used for a given language server, so we should not using LanguageServiceAccessor here.

I think to have a clean API, LSPClientFeatures should expose toOffset and toPosition. It will allow to override it and manage offset,position according the language server.

The toOffset should have a new parameter serverReportsOffsetOnlyPositions and the toOffset of LSPClientFeatures should call LSPIJUtils#toOffset with this flag.

SCWells72: I'm not sure that I follow. Are you saying that you want every current call to `LSPIJUtils.toOffset()` to instead call `LSPClientFeatures.toOffset()` and similar with `toPosition()`? If so, the ripple effect of that change is going to be massive, and it's going to require getting the `LSPClientFeatures` all over the place where those calls currently exist. Apologies if I'm misunderstanding the guidance here...

SCWells72: I do, however, understand that you're saying that `hasAny()` is evaluating for multiple language servers for file and can result in an incorrect answer when `toOffset()` is called for a specific language server for which there are multiple registered language servers. Let me ponder that...

src/main/java/com/redhat/devtools/lsp4ij/LSPIJUtils.java
```diff
@@ -595,12 +596,20 @@ public static int toOffset(int line, int character, @NotNull Document document)
...
         // Adjust position line/character according to this comment https://github.com/microsoft/vscode-languageserver-node/blob/ed3cd0f78c1495913bda7318ace2be7f968008af/textDocument/src/main.ts#L26
         if (line >= document.getLineCount()) {
-            // The line number is greater than the number of lines in a document, it defaults back to the number of lines in the document.
+            // The line number is greater than the number of lines in a document, it defaults to the last valid offset in the document.
             return document.getTextLength();
         } else if (line < 0) {
             // The line number is negative, it defaults to 0.
             return 0;
         }
+        // If the line number is 0 and the character is beyond the length of the first line, treat it as an offset-only
```
SCWells72, line 605: Made the logic by which this exception takes place very specific, basically only when a `Position` is returned with `line=0` and `character>lines[0].length`.

SCWells72:
Reverted the previous commit to use client config and made the logic for deciding when to translate what seems to be an offset-only position into the reported offset much more specific to an _overflow_ character on _exactly the first line_ of a _multi-line_ file. It's still technically out of spec, but only for one specific situation where it's almost certainly supposed to behave in this manner.

src/main/java/com/redhat/devtools/lsp4ij/LSPIJUtils.java
```diff
@@ -595,12 +596,22 @@ public static int toOffset(int line, int character, @NotNull Document document)
...
         // Adjust position line/character according to this comment https://github.com/microsoft/vscode-languageserver-node/blob/ed3cd0f78c1495913bda7318ace2be7f968008af/textDocument/src/main.ts#L26
         if (line >= document.getLineCount()) {
-            // The line number is greater than the number of lines in a document, it defaults back to the number of lines in the document.
+            // The line number is greater than the number of lines in a document, it defaults to the last valid offset in the document.
             return document.getTextLength();
         } else if (line < 0) {
             // The line number is negative, it defaults to 0.
             return 0;
         }
+        // If the line number is 0 and the character is beyond the length of the first line in a multi-line document,
```
SCWells72, line 605: Now the logic to detect an offset-only position and translate it to exactly the reported offset is _very_ specific. It will only occur if:
1. `line=0`
2. `character>lines[0].length`
3. The document has more than one line.
Again, it's technically out-of-spec, but I imagine that it should behave as-intended for all but significantly ill-behaved language servers.

src/main/java/com/redhat/devtools/lsp4ij/LSPIJUtils.java
angelozerr: > I'm not sure that I follow. Are you saying that you want every current call to `LSPIJUtils.toOffset()` to instead call `LSPClientFeatures.toOffset()` and similar with `toPosition()`? If so, the ripple effect of that change is going to be massive, and it's going to require getting the `LSPClientFeatures` all over the place where those calls currently exist. Apologies if I'm misunderstanding the guidance here...

I agree that it is not a trivial refactoring, but imho we should do that like I did for uri.

Perhaps we should start that in an another PR to provide toOffset and toPosition in lsp client features. Instead of working with lsp client features we shoould work with an interface like I did for uri.

Having toOffset and toPosition in lsp client features will be powerfull

angelozerr: Indeed let see that after the release but I ghink toOffset and toPosition must be an api that you could override to manage your custme case with pure offset.

Lsp4ij should follow lsp spec and not manage specific usecases. Having toOffset as api can be override on your side znd manage the pure offset.

Issue #817, opened 2025-02-09
ZLS "organize @imports" code action results in missing lines


FalsePattern:
Forwarded bug report from ZigBrains, i've confirmed it on pure LSP4IJ without ZigBrains as well (commit 60adee60)

> When I press Alt+Enter for the Context Action menu and select "organize [@imports](https://github.com/imports)", the following code goes from:
> 
> ```
> const std = @import("std");
> const testing = std.testing;
> 
> const ast = @import("ast.zig");
> const Lexer = @import("lexer.zig").Lexer;
> const token = @import("token.zig");
> ```
> 
> to:
> 
> ```
> const testing = std.testing;
> 
> const ast = @import("ast.zig");
> const Lexer = @import("lexer.zig").Lexer;
> const token = @import("token.zig");
> 
> const std = @import("std");
> ```
> 
> When I select "organize [@imports](https://github.com/imports)" again, the code changes into:
> 
> ```
> 
> const ast = @import("ast.zig");
> const Lexer = @import("lexer.zig").Lexer;
> const token = @import("token.zig");
> 
> const testing = std.testing;
> ```
> 
> Note that the `@import` for `std` is now missing and is replace by a blank line.
> 
> System Details:
> 
>     * JetBrains: RustRover 2024.3.4
> 
>     * OS: Linux Mint 21.3 (Ubuntu 22.04)
> 
>     * Zig version: 0.14.0-dev.3062+ff551374a
> 
>     * ZLS version: 0.14.0-dev.369+c8bd727

FalsePattern:
As a test on a local repo, I've invoked the organize imports action on the following code:

![Image](https://github.com/user-attachments/assets/9dd1305f-fe49-442b-b643-139fa371bfa5)

Which resulted in the following TextEdit array sent by the language server:

![Image](https://github.com/user-attachments/assets/b0ad3498-8817-40e6-be5c-9aba7778b35f)

When sorted by `TEXT_EDITS_DESCENDING_COMPARATOR`, changes to this:

![Image](https://github.com/user-attachments/assets/07be6f15-fef5-4a61-8a0d-0053af0cb9d2)

And after applying, resulted in this code:

![Image](https://github.com/user-attachments/assets/2d4e1fa3-12b3-461f-947d-abb40e1f5076)

FalsePattern:
As a test, i've modified the comparator locally to sort "removal" (0-length) TextEdit instances before non-empty TextEdits:

[...]

PR #819, opened 2025-02-09
fix: Made code action edit apply logic order-independent

FalsePattern:
Resolves https://github.com/redhat-developer/lsp4ij/issues/817

### DIFF
--- a/src/main/java/com/redhat/devtools/lsp4ij/LSPIJUtils.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/LSPIJUtils.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2019 Red Hat, Inc.
+ * Copyright (c) 2019-2025 Red Hat, Inc.
  * Distributed under license by Red Hat, Inc. All rights reserved.
  * This program is made available under the terms of the
  * Eclipse Public License v2.0 which accompanies this distribution,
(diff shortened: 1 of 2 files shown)


angelozerr:
It seems your PR breaks some existing tests.

Please create also a test with your usecase like https://github.com/redhat-developer/lsp4ij/blob/main/src%2Ftest%2Fjava%2Fcom%2Fredhat%2Fdevtools%2Flsp4ij%2FLSPIJUtils_applyWorkspaceEdit_TypeScriptTest.java

angelozerr:
@FalsePattern i have planned to create a release tomorrow.

I fear that it will hard to fix this issue for tomorrow. Is it annoying?

FalsePattern:
The latest commit should work, it passes the tests on my system. The issue was that i tried to also use a RangeMarker for the caret, but the old logic is fine for that, so i reverted that change back to the "cumulative offset" logic in the for loop and only kept the RangeMarker logic for the edits themselves

FalsePattern:
Alright, it's ready to merge now, it seems to be working correctly no matter how much i shuffle the test file

angelozerr:
> Alright, it's ready to merge now, it seems to be working correctly no matter how much i shuffle the test file

Wow impressive ! I will review it asap.

@SCWells72 if you have time to review it, you are welcome :)

PR #826, opened 2025-02-10
fix: Prospective fix for an observed `ConcurrentModificationException` when multiple formatting edits are applied in quick succession while typing repeated close braces

SCWells72:
It looks like someone is reusing the same collection of text edits, so basically we now just create an owned, mutable (if needed) copy of the formatting text edits before applying them (which in one situation entails sorting/mutating them).


angelozerr:
Your PR prevents from concurrent modif exception but what about the result if zpply edits is executed twice?

angelozerr:
Thanks @SCWells72 !

Issue #835, opened 2025-02-11
LSP inlay hint request blocks the EDT


FalsePattern:
The `LSPInlayHintsProvider.doCollect` method blocks inside a read action while waiting for the LSP to respond, and in the case of that large file, the LSP can take a long time to complete, which causes IDE stutter every time inlay hints are refreshed (every few seconds when typing)
![image](https://github.com/user-attachments/assets/ca564a45-0cef-4f31-ade4-1aa893506fcf)

I've found a dirty workaround for this locally by always putting the future into the `pendingFutures` if it's not finished, and relying on the caching behaviour of the LSPInlayHintsSupport to provide the same future once the `refreshEditorFeatureWhenAllDone` triggers another inlay hint pass. This fixes the EDT freeze caused by the waitUntilDone call.
![image](https://github.com/user-attachments/assets/01efc023-6429-48fe-87ff-23b5ad6242cb)

As a proper fix, we could add a `boolean waitUntilDoneOrTimeout(future, psiFile, int timeoutMillis, List<CompletableFuture<?>> pendingFutures)` method to `CompletableFutures` so that inlay hints that complete fast (<50ms or similar) get done synchronously without this extra async step, but ones that take longer will get the async route.

PR #836, opened 2025-02-11
feat: Non-blocking LSP requests for inlay hints and color hints

FalsePattern:
Proposed solution for https://github.com/redhat-developer/lsp4ij/issues/835

### DIFF
--- a/src/main/java/com/redhat/devtools/lsp4ij/features/inlayhint/LSPInlayHintsProvider.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/features/inlayhint/LSPInlayHintsProvider.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2022 Red Hat, Inc.
+ * Copyright (c) 2022-2025 Red Hat, Inc.
  * Distributed under license by Red Hat, Inc. All rights reserved.
  * This program is made available under the terms of the
  * Eclipse Public License v2.0 which accompanies this distribution,
@@ -7,7 +7,7 @@
  *
  * Contributors:
  * Red Hat, Inc. - initial API and implementation
- * FalsePattern - port to declarative inlay hint API
+ * FalsePattern - port to declarative inlay hint API, non-blocking LSP request
  ******************************************************************************/
 package com.redhat.devtools.lsp4ij.features.inlayhint;
 
@@ -33,6 +33,7 @@ import java.util.List;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeoutException;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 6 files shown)


src/main/java/com/redhat/devtools/lsp4ij/internal/CompletableFutures.java
```diff
@@ -109,6 +110,23 @@ public static void waitUntilDone(@Nullable CompletableFuture<?> future,
...
+     * If the timeout is reached, the future is put into the <code>pendingFutures</code> list and the method returns false.
+     * If <code>pendingFutures</code> is null, the function will retry the wait indefinitely, blocking the thread.
+     *
+     * @param future         the future to wait.
+     * @param file           the Psi file.
+     * @param timeout        the amount of milliseconds to wait in {@link Future#get(long, TimeUnit)}
+     * @param pendingFutures the list to put the unfinished future into on timeout. If <code>null</code>,
+     *                       the timeout is retried indefinitely until successful.
+     */
+    public static boolean waitUntilDoneOrTimeout(@NotNull CompletableFuture<?> future,
```
angelozerr, line 126: Instead of having a new method, I think we should add an Integer timeout which can be null.

pendingFutures is a specific case that LSP inlay hints should manage.

src/main/java/com/redhat/devtools/lsp4ij/internal/CompletableFutures.java
```diff
@@ -132,6 +150,10 @@ public static void waitUntilDone(@Nullable CompletableFuture<?> future,
                     // This wait can block IJ, here we stop the wait (and we could lose some LSP feature)
                     throw new CancellationException("Some projects are indexing");
                 }
+                if (pendingFutures != null) {
+                    pendingFutures.add(future);
```
angelozerr, line 154: I wonder if it should be better to store pendingFutures in the editor data and use an HashSet to avoid adding the same future?

FalsePattern:
I've updated the code to use an overload for the waitUntilDone function instead, and replaced the `pendingFutures` system with sets. I'm not sure what you mean by "editor data".

src/main/java/com/redhat/devtools/lsp4ij/features/inlayhint/LSPInlayHintsProvider.java
```diff
@@ -61,7 +61,7 @@ protected void doCollect(@NotNull PsiFile psiFile,
 
         try {
             // Wait until the future is finished and stop the wait if there are some ProcessCanceledException.
-            waitUntilDone(future, psiFile);
+            waitUntilDone(future, psiFile, 25);
```
angelozerr, line 64: Is 25 ms is enough? Perhaps we should set 500 ms to cover cases of ls which give  a fast LSP response?

src/main/java/com/redhat/devtools/lsp4ij/features/inlayhint/LSPInlayHintsProvider.java
```diff
@@ -51,8 +52,7 @@ public class LSPInlayHintsProvider extends AbstractLSPDeclarativeInlayHintsProvi
     @Override
     protected void doCollect(@NotNull PsiFile psiFile,
                              @NotNull Editor editor,
-                             @NotNull InlayTreeSink inlayHintsSink,
-                             @NotNull List<CompletableFuture<?>> pendingFutures) {
```
angelozerr, line 55: Please keep the `@NotNull List<CompletableFuture<?>> pendingFutures)`  parameters. Class which implement doCollect must don't know how pendingFutures is stored.

angelozerr:
@FalsePattern let me do some fix in an other PR.

angelozerr:
Thanks so much @FalsePattern for your great idea.

I close this PR in favor of https://github.com/redhat-developer/lsp4ij/pull/857

PR #845, opened 2025-02-14
feat: TextMate-specific editor improvements

SCWells72:
Enables TextMate-specific functionality that had to be disabled while 2023.2 was still supported. This includes improved support for paired nested braces/brackets/parens while typing and derivation of supported quote characters from the TextMate metadata.

Here's a quick demo of how paired nested braces/brackets/parens behave _without_ this fix where nested braces/brackets/parens aren't properly paired as they're typed resulting in likely syntax errors:

LSP_TextMate_Paired_Nested_Braces_Issue

and how it behaves _with_ this fix where they are properly paired if unbalanced, and once balanced, typing the close character to the right of the caret just advances the caret as expected by the user:

LSP_TextMate_Paired_Nested_Braces_Fixed

PR #856, opened 2025-02-19
feat: TextMate-specific editor improvements

SCWells72:
This is a fresh branch/PR for #845 since that one is having trouble being merged via rebase.


angelozerr:
Thanks @SCWells72 !

PR #853, opened 2025-02-18
feat: Custom file view provider for LSP4IJ TextMate files that uses semantic token information to create a simple PSI "tree" (fixes #98, #531, and #813)

SCWells72:
Okay, I'll be adding quite a bit of detail to the changes themselves, but at a high-level, this adds a `FileViewProviderFactory` for LSP4IJ TextMate files and plain text/abstract file type files in Community Edition IDEs to simulate a flat PSI tree based the language server's reported semantic tokens for the file including whether those elements represent declarations, references, comments, string and numeric literals, etc.

Non-user-defined language server integrations can also opt-in for these features without changing their files' PSI element hierarchies by registering a file view provider factory for their files' language or file type. This has been confirmed to work properly via collaboration with @ericdallo and is documented in `LSPApi.md`.

The end result is that the following issues are fixed and features are added:

* For LSP files against servers that support sufficiently descriptive semantic tokens, only declarations and references to declarations are highlighted on **Ctrl/Cmd+Mouseover**, and nothing else is ever highlighted.
* For LSP files against servers that do _not_ support semantic tokens such as the CSS language server, the entire file is no longer highlighted on **Ctrl/Cmd+Mouseover**. This is also the behavior for LSP files against language servers that _do_ support semantic tokens when the semantic tokens for the file are not yet populated.
* The tooltip shown when hovering in this manner is now much more descriptive, showing the declaration type if known from the semantic tokens and their modifiers followed by the declaration name and, if hovering over a reference that resolves to a declaration in another file, the target file name is also shown. Also, the symbolic names are displayed as code in these tooltips.
* Attempting to navigate on a reference takes you to the declaration. Attempting to navigate on a declaration shows its usages.

This feature can be disabled for user-defined language servers via the client configuration setting `editor.enableSemanticTokensFileViewProvider`. That setting defaults to `true` for all language server definitions, but as stated above, unless a non-user-defined language server integration opts in explicitly by registering a file view provider, it is effectively disabled.

Here's a little demo:

LSP_File_View_Provider_Demo

Regarding concerns about memory bloat or slowness from something like this, I confirmed the following:

* IDE memory usage before opening large file was ~200MB (after multiple forced GCs).
* Opened a ~30K line TypeScript file (`lib.dom.d.ts`).
* IDE memory usage increased about 45MB (again, foricing GCs).
* Verified that file editing performance was normal.
* Verified that the new **Ctrl/Cmd+Mouseover** behavior worked properly and tracked accurately after making changes.
[...]

Issue #531, opened 2024-09-20
Open references if definition result has the same range than clicked offset


InSyncWithFoo:
Currently <kbd>Ctrl</kbd> + Click always send `textDocument/definition`, which might not be ideal when the definition itself is clicked. LSP4IJ integrators should be able to configure that programmatically.

(Filed as requested at [this issue](https://github.com/InSyncWithFoo/pyright-for-pycharm/issues/72).)

CppCXY:
in VSCode, when you ctrl + click on a reference, it first sends a textDocument/definition request. If the returned range overlaps with the clicked position, it then sends a textDocument/references request.

angelozerr:
Thanks so much @CppCXY for your information.

I see now more the benefit to consume references.

@InSyncWithFoo is it the same usecase in the reported issue from your project?

angelozerr:
> Yes, that's probably the best behaviour.

Indeed, but do you think it is th ebehavior than user would like to have?

angelozerr:
@InSyncWithFoo after debugging Java support in IJ, I undertand more how the go to declaration is working.

There are 2 means to manage go to declaration:

 * extends GotoDeclarationHandler like we did with LSPGotoDeclarationHandler. The advantage to use that is if there are some custom PsiFile (it is your case @InSyncWithFoo I think), the go to declaration will take the priority and if another plugin which manages go to declaration like PyCharm I think,the LSP go to definition will work. The inconevinant to use this GotoDeclarationHandler are:
   * the hyperlinked ignore the LSP location link and with textmate you have this issue  https://github.com/redhat-developer/lsp4ij/issues/98
   * there is no way to manage the click action, so it is impossible to Show the References panel.
* using the PsiFile to collect declarations / references. In the LSP context, we have no custom PsiFile BUT it is possible to get declarations / referernces by implementing PsiSymbolDeclarationProvider / ImplicitReferenceProvider. But there are 2 problems:
  *  ImplicitReferenceProvider#getImplicitReferences is called every time (when you do an hoverfor instance), so if we implement textDocument/definition in the getImplicitReferences, it will consume textDocument/definitionevery time. In our case we need to execute getImplicitReferences  only when a GoToAction is executed. So we could implement an action listener before/after to enable/disable the call of getImplicitReferences 
[...]

PR #882, opened 2025-03-11
feat: Custom file view provider for LSP4IJ TextMate files that uses semantic token information to create a simple PSI "tree" (fixes #98, #531, and #813)

### LINKED ISSUE #98: Wrong hyperlinked element (definition, documentLink) when there are no Psi
When a file is not parsed with a specific Psi parser (TEXT, texmate), the hyperlinked element takes the full content of the file.

Here is a demo with the ugly problem: 

WrongHyperlinked

To manage LSP definition we implement GotoDeclarationHandler which doesn't provide the capability to customize the source range.

I have tried to support LSP definition with other means and one idea was to implement https://github.com/JetBrains/intellij-community/blob/6c42d0fad1db66faa4e5549935ca3eab91c5b2ca/platform/core-api/src/com/intellij/model/psi/ImplicitReferenceProvider.java#L32 where we can customize the source range.

The only problem is that this method is called every time (when you are typing when the file is loaded, etc). I discovered with https://github.com/redhat-developer/lsp4ij/pull/356 that we can track which action is executed.

One idea is:

 * implement  https://github.com/JetBrains/intellij-community/blob/6c42d0fad1db66faa4e5549935ca3eab91c5b2ca/platform/core-api/src/com/intellij/model/psi/ImplicitReferenceProvider.java#L32
 * at the start of the method we could check (with the action listener) if the last executed action is GoToDeclaration.

SCWells72:
This is a fresh PR/branch for #853 which, while clean to merge against main, isn't easy to merge via rebase in its current state. It is the exact same contents as that other PR/branch.

### DIFF
--- a/docs/LSPApi.md
+++ b/docs/LSPApi.md
@@ -511,6 +511,115 @@ public class MyLSPSemanticTokensFeature extends LSPSemanticTokensFeature {
 }
 ```
 
+### Semantic Tokens File View Providers
+
+LSP4IJ can help incorporate a file's semantic tokens information so that it's readily available and can be used to
+implement specific behavior based on whether an element is a declaration or reference, string or numeric literal,
+comment, etc., based on the reported semantic tokens. This is implemented via a custom `FileViewProviderFactory` and
+`FileViewProvider`. LSP4IJ includes default implementations of these for TextMate files and plain text files associated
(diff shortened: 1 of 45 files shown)


angelozerr:
Thanks so much @SCWells72 !

Issue #7, opened 2023-10-03
Cache mapping between file and servers


angelozerr:
Today on each LSP operations, mapping (which uses DocumentMatcher) are processed, it could be better to cache them to have better performance (if we see that performance is degraded).

PR #883, opened 2025-03-11
feat: Add support for breadcrumbs and sticky lines (fixes #700)

### LINKED ISSUE #700: Implement IDE's `breadcrumbsInfoProvider` EP
I'm not 100% sure which LSP capability would best lend itself to this EP, but the information is definitely there. Basically it would just need to be able to discern the containing declarations up to the file itself exclusive. That would enable both more detailed breadcrumbs and sticky lines in 2024.1+.

SCWells72:
This PR implements support for the IDE's `breadcrumbsInfoProvider` -- including support for sticky lines ins 2024.1+ -- using LSP `textDocument/documentSymbol` information. Here's an example:

image

Overall the changes should be pretty straightforward aside from one aspect where, due to the sparsely-populated PSI element structure of plain text and TextMate files from the recent view provider work, we have to capture and retain -- via inheritable thread locals -- the last requested offset so that the correct location is used to enumerate breadcrumbs. I'll provide more details in the respective sections in the PR where it should make more sense.

This feature is _enabled by default_ for both custom and user-defined language server definitions, but it is _disabled by default_ in the templates for the XML and HTML language servers since those can produce such large document symbol responses and already have first-class language/breadcrumb support in all JetBrains IDEs. It is exposed as a feature flag and can be disabled if desired, either by overriding the respective method for a custom language server definition or via client config for a user-defined language server definition.


docs/LSPApi.md
```diff
@@ -653,13 +653,14 @@ or, if appropriate, the the child/descendant element that represents the declara
...
-| boolean isEnableSemanticTokensFileViewProvider(PsiFile file)   | Returns `true` if the semantic tokens-based file view provider is enabled and `false` otherwise.                                  | `true`, but a file view provider must be registered for non-user-defined language server definitions |
+| API                                                                  | Description                                                                                                                       | Default Behaviour                                                                                    |
+|----------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------|
```
SCWells72, line 657: Best to review this ignoring whitespace-only changes.

docs/LSPApi.md
```diff
@@ -653,13 +653,14 @@ or, if appropriate, the the child/descendant element that represents the declara
...
+| boolean isEnableTextMateNestedBracesImprovements(PsiFile file)       | Returns `true` if editor improvements for nested braces/brackets/parentheses in TextMate files are enabled and `false` otherwise. | `true` for user-defined language server definitions; otherwise `false`                               |
+| boolean isEnableSemanticTokensFileViewProvider(PsiFile file)         | Returns `true` if the semantic tokens-based file view provider is enabled and `false` otherwise.                                  | `true`, but a file view provider must be registered for non-user-defined language server definitions |
+| boolean isEnableDocumentSymbolsBreadcrumbsInfoProvider(PsiFile file) | Returns `true` if the the document symbols-based breadcrumbs info provider is enabled and `false` otherwise.                      | `true` for user-defined language server definitions; otherwise `false`                               |
```
SCWells72, line 663: Added documentation for the new feature flag by which this can be disabled.

src/main/java/com/redhat/devtools/lsp4ij/client/features/EditorBehaviorFeature.java
```diff
@@ -110,6 +110,17 @@ public boolean isEnableSemanticTokensFileViewProvider(@NotNull PsiFile file) {
         return true;
     }
 
+    /**
+     * Whether or not the document symbols-based breadcrumbs info provider is enabled.
+     *
+     * @param file the file
+     * @return true if the document symbols-based breadcrumbs info provider is enabled; otherwise false
+     */
+    public boolean isEnableDocumentSymbolsBreadcrumbsInfoProvider(@NotNull PsiFile file) {
```
SCWells72, line 119: Feature flag defaulting to _enabled_.

src/main/java/com/redhat/devtools/lsp4ij/client/features/EditorBehaviorFeature.java
```diff
@@ -181,4 +192,17 @@ public static boolean enableSemanticTokensFileViewProvider(@NotNull PsiFile file
...
         );
     }
+
+    /**
+     * Whether or not the document symbols-based breadcrumbs info provider is enabled.
+     *
+     * @param file the PSI file
+     * @return true if the document symbols-based breadcrumbs info provider is enabled; otherwise false
+     */
+    public static boolean enableDocumentSymbolsBreadcrumbsInfoProvider(@NotNull PsiFile file) {
```
SCWells72, line 202: Convenience access for feature flag.

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/DocumentSymbolData.java
```diff
@@ -37,7 +38,7 @@ public class DocumentSymbolData extends FakePsiElement {
     private final @NotNull PsiFile psiFile;
     private final DocumentSymbolData parent;
     private final @NotNull LanguageServerItem languageServer;
-    private TextRange textRange;
+    private volatile TextRange textRange = null;
```
SCWells72, line 41: This could be reverted if desired, but I noticed that `textRange` was being lazy-initialized in a manner that's not guaranteed atomic, so I converted it to use DCL.

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPDocumentSymbolBreadcrumbsInfoProvider.java
```diff
@@ -0,0 +1,152 @@
...
+import javax.swing.Icon;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Document symbol-based breadcrumbs info provider.
+ */
+public class LSPDocumentSymbolBreadcrumbsInfoProvider implements BreadcrumbsProvider {
```
SCWells72, line 43: This is the core of the feature, an implementation of `breadcrumbsInfoProvider` for LSP4IJ files.

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPDocumentSymbolBreadcrumbsInfoProvider.java
```diff
@@ -0,0 +1,152 @@
...
+
+    @Override
+    public Language[] getLanguages() {
+        // Register for all languages and filter in isSupported()
+        List<Language> languages = LanguageUtil.getLanguages(Conditions.alwaysTrue());
```
SCWells72, line 48: Instead of registering for distinct languages in `plugin.xml`, this EP returns the supported languages. Since we don't know the full set at packaging time, it just returns all languages and then filtering occurs in the respective method implementations. This allows it the feature to work for all LSP4IJ files.

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPDocumentSymbolBreadcrumbsInfoProvider.java
```diff
@@ -0,0 +1,152 @@
...
+                        // The document symbols feature must be available for the file's language server
+                        ls.getClientFeatures().getDocumentSymbolFeature().isEnabled(file) &&
+                                ls.getClientFeatures().getDocumentSymbolFeature().isSupported(file) &&
+                                // And this feature must be enabled
+                                EditorBehaviorFeature.enableDocumentSymbolsBreadcrumbsInfoProvider(file)
+        );
+    }
+
+    @Override
+    public boolean acceptElement(@NotNull PsiElement element) {
```
SCWells72, line 68: This method is called to determine which elements to include in sticky lines in 2024.1+. We only want to include those that are wrapped as `DocumentSymbolData`.

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPDocumentSymbolBreadcrumbsInfoProvider.java
```diff
@@ -0,0 +1,152 @@
...
+        // NOTE: This is only used to determine which already-vetted elements should be included in the breadcrumb trail
+        return element instanceof DocumentSymbolData;
+    }
+
+    @Override
+    @Nullable
+    public PsiElement getParent(@NotNull PsiElement element) {
```
SCWells72, line 75: The following two methods are how the caller determines what to show based on the offset. I wish it passed the offset itself, but instead it passes the element at that offset. In a plain text/TextMate LSP4IJ file, we do now have PSI elements, but only for semantic tokens. The intermediate space does not produce a distinct element aside from the file itself. That's why we're going to need to use inheritable thread locals with the original requested offset. More on that later, but just setting the context.

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPDocumentSymbolStructureViewModel.java
```diff
@@ -72,14 +82,22 @@ public LSPFileStructureViewElement(@NotNull PsiFile psiFile) {
 
         @Override
         public @NotNull Collection<StructureViewTreeElement> getChildrenBase() {
-            return collectElements(getElement());
+            PsiFile file = getElement();
```
SCWells72, line 85: Just making the behavior around `null` more explicit here.

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPDocumentSymbolStructureViewModel.java
```diff
@@ -72,14 +82,22 @@ public LSPFileStructureViewElement(@NotNull PsiFile psiFile) {
...
+            return file != null ? collectElements(file) : Collections.emptyList();
         }
 
         private @NotNull Collection<StructureViewTreeElement> collectElements(@NotNull PsiFile psiFile) {
-            if(ProjectIndexingManager.getInstance(psiFile.getProject()).isIndexingAll()) {
+            if (ProjectIndexingManager.isIndexingAll()) {
```
SCWells72, line 90: Just changing this based on the code inspection/intention.

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPDocumentSymbolStructureViewModel.java
```diff
@@ -101,26 +119,58 @@ public LSPFileStructureViewElement(@NotNull PsiFile psiFile) {
                     return Collections.emptyList();
                 }
                 return documentSymbols.stream()
-                        .map(documentSymbol -> getStructureViewTreeElement(documentSymbol))
+                        .map(LSPDocumentSymbolStructureViewModel::getStructureViewTreeElement)
```
SCWells72, line 122: Again, just applying a recommend code change.

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPDocumentSymbolStructureViewModel.java
```diff
@@ -101,26 +119,58 @@ public LSPFileStructureViewElement(@NotNull PsiFile psiFile) {
...
+            }
+
+            List<LanguageServerItem> languageServers = languageServersFuture.getNow(Collections.emptyList());
+            return !ContainerUtil.isEmpty(languageServers);
+        }
+
         @Override
         public @Nullable String getPresentableText() {
-            return getElement().getName();
+            PsiFile file = getElement();
```
SCWells72, line 160: More explicit handling of `null` values to be defensive.

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPDocumentSymbolStructureViewModel.java
```diff
@@ -136,19 +186,20 @@ public LSPDocumentSymbolViewElement(DocumentSymbolData documentSymbolData) {
         }
 
         private @NotNull Collection<StructureViewTreeElement> collectElements(@Nullable DocumentSymbolData documentSymbolData) {
-            var children = documentSymbolData.getChildren();
-            if (children.length == 0) {
+            var children = documentSymbolData != null ? documentSymbolData.getChildren() : null;
```
SCWells72, line 189: And again, a combination of being defensive against `null` values and applying recommended code changes.

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPDocumentSymbolUtils.java
```diff
@@ -0,0 +1,127 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.LinkedList;
+import java.util.List;
+
+/**
+ * Utility class for working with document symbols.
+ */
+final class LSPDocumentSymbolUtils {
```
SCWells72, line 39: Simple package-protected utility class that includes common behavior around document symbol information.

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPDocumentSymbolUtils.java
```diff
@@ -0,0 +1,127 @@
...
+    /**
+     * Returns the closest containing document symbol data for the element.
+     *
+     * @param element the element
+     * @return the closest containing document symbol data, or null if none was found
+     */
+    @Nullable
+    static DocumentSymbolData getDocumentSymbolData(@NotNull PsiElement element) {
+        LSPSemanticTokensFileViewProvider semanticTokensFileViewProvider = LSPSemanticTokensFileViewProvider.getInstance(element);
+        int effectiveOffset = semanticTokensFileViewProvider != null ? semanticTokensFileViewProvider.getEffectiveOffset(element) : -1;
```
SCWells72, line 71: `effectiveOffset` here is the offset that was requested if not on an actual element or the offset of the requested actual element. Again, this is part of how we can handle interpolation between real elements and show the correct breadcrumbs when the caret is on an offset that doesn't correspond to a (non-file) PSI element and sticky lines as the user scrolls the editor.

src/main/java/com/redhat/devtools/lsp4ij/features/semanticTokens/viewProvider/LSPSemanticToken.java
```diff
@@ -134,6 +139,7 @@ class LSPSemanticToken {
     private final boolean isFileLevel;
     private final LSPSemanticTokenElementType elementType;
     private volatile LSPSemanticTokenPsiElement element = null;
+    private final ThreadLocal<Integer> lastRequestedOffsetPtr = new InheritableThreadLocal<>();
```
SCWells72, line 142: This is one of the two inheritable thread locals that could contain the last requested offset into the file that didn't correspond to a semantic token element. It's automatically cleared when a semantic token element is requested so that it stays "fresh". This effectively allows us to simulate having a fully-populated PSI element structure for the file while only having PSI elements for semantic tokens.

src/main/java/com/redhat/devtools/lsp4ij/features/semanticTokens/viewProvider/LSPSemanticToken.java
```diff
@@ -254,6 +260,27 @@ LSPSemanticTokenPsiElement getElement() {
         return element;
     }
 
+    /**
+     * Stores the last requested offset for which a file-level semantic token was returned as a thread local.
+     *
+     * @param offset the last requested offset
+     */
+    void setLastRequestedOffset(int offset) {
+        if (isFileLevel) {
```
SCWells72, line 269: Again, only for file-level, so this wouldn't be populated for an offset that corresponds to a semantic token PSI element.

angelozerr:
@SCWells72 could you add doc at https://github.com/redhat-developer/lsp4ij/blob/main/docs/LSPSupport.md#document-symbol

 * add a new section `Structure View` before `Here is an example with [TypeScript Language Server](https://github.com/redhat-developer/lsp4ij/blob/main/docs/user-defined-ls/typescript-language-server.md) which ope...`
 * add a new Section `Breadcrumb` to explains that document symbol is also used to fill breadcrumb and add a screenshot.

In https://github.com/redhat-developer/lsp4ij/blob/main/docs/DeveloperGuide.md#document-symbol add some information about breadcrumb enabled like you have explained with XML, HTML and explains how to activate for custom plugin

src/main/java/com/redhat/devtools/lsp4ij/features/semanticTokens/viewProvider/LSPSemanticTokensFileViewProviderHelper.java
```diff
@@ -34,6 +35,7 @@
 public class LSPSemanticTokensFileViewProviderHelper implements LSPSemanticTokensContainer {
 
     private final LSPSemanticTokensFileViewProvider fileViewProvider;
+    private final ThreadLocal<Integer> effectiveOffsetPtr = new InheritableThreadLocal<>();
```
SCWells72, line 38: And this is the second inheritable thread local, and it's the one that's used for sticky lines as the user scrolls the editor viewport. Together, these allows us to have a full effective offset regardless of whether or not it corresponds to a real PSI element (whether for semantic tokens or a real PSI structure) or not, even for language servers that don't support semantic tokens, e.g. CSS.

src/main/java/com/redhat/devtools/lsp4ij/features/semanticTokens/viewProvider/LSPSemanticTokensFileViewProviderHelper.java
```diff
@@ -189,19 +191,62 @@ LSPSemanticToken getSemanticToken(int offset) {
         // If this file has semantic tokens, use them
         Map<Integer, LSPSemanticToken> semanticTokensByOffset = getSemanticTokensByOffset();
         if (!ContainerUtil.isEmpty(semanticTokensByOffset)) {
-            return semanticTokensByOffset.get(offset);
+            LSPSemanticToken semanticToken = semanticTokensByOffset.get(offset);
+            // Update the view provider's effective offset as appropriate
+            setEffectiveOffset(semanticToken == null ? offset : -1);
```
SCWells72, line 196: If the requested offset corresponds to a semantic token PSI element, store that information.

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPDocumentSymbolStructureViewModel.java
```diff
@@ -72,14 +82,22 @@ public LSPFileStructureViewElement(@NotNull PsiFile psiFile) {
...
         }
 
         private @NotNull Collection<StructureViewTreeElement> collectElements(@NotNull PsiFile psiFile) {
-            if(ProjectIndexingManager.getInstance(psiFile.getProject()).isIndexingAll()) {
+            if (ProjectIndexingManager.isIndexingAll()) {
+                return Collections.emptyList();
+            }
+
+            // Don't start a language server for this on the EDT or it can freeze the app
+            if (ApplicationManager.getApplication().isDispatchThread() && !hasStartedLanguageServers(psiFile)) {
```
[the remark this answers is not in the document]
angelozerr: I think you should not do that here because this code is also used by Structured view. My PR which fixes freeze should avoid freezing IJ.

src/main/java/com/redhat/devtools/lsp4ij/features/semanticTokens/viewProvider/LSPSemanticTokensFileViewProviderHelper.java
```diff
@@ -189,19 +191,62 @@ LSPSemanticToken getSemanticToken(int offset) {
...
+     * Stores the effective offset as a thread local.
+     *
+     * @param offset the effective offset
+     */
+    private void setEffectiveOffset(int offset) {
+        effectiveOffsetPtr.set(offset);
+    }
+
+    @Override
+    public int getEffectiveOffset(@NotNull PsiElement element) {
```
SCWells72, line 226: And this is how we compute it. If it's for a real element, return the effective offset of that element. If it's not, return the last requested offset from the view provider. If neither, just return -1 denoting that there's no effective offset.

src/main/java/com/redhat/devtools/lsp4ij/server/definition/launching/ClientConfigurationSettings.java
```diff
@@ -67,6 +67,11 @@ public static class ClientConfigurationEditorSettings {
          * Whether or not the semantic tokens-based file view provider is enabled. Defaults to true.
          */
         public boolean enableSemanticTokensFileViewProvider = true;
+
+        /**
+         * Whether or not the document symbols-based breadcrumbs info provider is enabled. Defaults to true.
+         */
+        public boolean enableDocumentSymbolsBreadcrumbsInfoProvider = true;
```
SCWells72, line 74: Feature flag in client config.

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPDocumentSymbolBreadcrumbsInfoProvider.java
angelozerr: Is this BreadcrumbsProvider API is available for old IJ supported by LSP4IJ?

src/main/java/com/redhat/devtools/lsp4ij/server/definition/launching/UserDefinedEditorBehaviorFeature.java
```diff
@@ -69,4 +69,11 @@ public boolean isEnableSemanticTokensFileViewProvider(@NotNull PsiFile file) {
         // Note that this defaults to enabled for user-defined language server definitions
         return (editorSettings == null) || editorSettings.enableSemanticTokensFileViewProvider;
     }
+
+    @Override
+    public boolean isEnableDocumentSymbolsBreadcrumbsInfoProvider(@NotNull PsiFile file) {
```
SCWells72, line 74: Feature flag accessor for user-defined language server definitions based on client config.

src/main/resources/META-INF/plugin.xml
```diff
@@ -313,6 +313,12 @@
                 language="textmate"
                 implementationClass="com.redhat.devtools.lsp4ij.features.documentSymbol.LSPDocumentSymbolStructureViewFactory"/>
 
+        <!-- Breadcrumbs/sticky lines support using document symbols -->
```
SCWells72, line 316: EP registration for the `breadcrumbsInfoProvider`.

src/main/resources/META-INF/plugin.xml
```diff
@@ -440,7 +446,8 @@
         <!-- LSP textDocument/semanticTokens request support -->
         <highlightVisitor
                 id="LSPSemanticTokensHighlightVisitor"
-                implementation="com.redhat.devtools.lsp4ij.features.semanticTokens.LSPSemanticTokensHighlightVisitor"/>
+                implementation="com.redhat.devtools.lsp4ij.features.semanticTokens.LSPSemanticTokensHighlightVisitor"
```
SCWells72, line 449: Ordered this first.

src/main/resources/jsonSchema/clientSettings.schema.json
```diff
@@ -88,6 +88,12 @@
           "title": "Enable semantic tokens-based file view provider",
           "description": "Whether or not the semantic tokens-based file view provider is enabled.",
           "default": true
+        },
+        "enableDocumentSymbolsBreadcrumbsInfoProvider": {
```
SCWells72, line 92: Client config JSON schema update for the new feature flag.

src/main/resources/templates/lemminx/clientSettings.json
```diff
@@ -6,6 +6,9 @@
     "useContextAwareSorting": true,
     "useTemplateForInvocationOnlySnippet": false
   },
+  "editor": {
```
SCWells72, line 9: Disabled by default for XML. Note that all JetBrains IDEs already have breadcrumbs support for XML anyway.

src/main/resources/templates/vscode-html-language-server/clientSettings.json
```diff
@@ -6,6 +6,9 @@
     "useContextAwareSorting": true,
     "useTemplateForInvocationOnlySnippet": false
   },
+  "editor": {
```
SCWells72, line 9: Same for HTML.

src/test/java/com/redhat/devtools/lsp4ij/features/documentSymbol/TypeScriptDocumentSymbolBreadcrumbsInfoProviderTest.java
```diff
@@ -0,0 +1,179 @@
...
+
+import com.redhat.devtools.lsp4ij.fixtures.LSPDocumentSymbolBreadcrumbsInfoProviderFixtureTestCase;
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Tests the document symbol-based breadcrumbs info provider for TypeScript.
+ */
+public class TypeScriptDocumentSymbolBreadcrumbsInfoProviderTest extends LSPDocumentSymbolBreadcrumbsInfoProviderFixtureTestCase {
```
SCWells72, line 22: Unit test to confirm breadcrumbs for TypeScript.

src/test/java/com/redhat/devtools/lsp4ij/features/documentSymbol/TypeScriptDocumentSymbolBreadcrumbsInfoProviderTest.java
```diff
@@ -0,0 +1,179 @@
...
+                Map.ofEntries(
+                        Map.entry("Foo", List.of("Foo")),
+                        Map.entry("bar", List.of("Foo", "bar")),
+                        Map.entry("console.log", List.of("Foo", "bar", "then() callback")),
+                        Map.entry("console.error", List.of("Foo", "bar", "catch() callback"))
+                )
+        );
+    }
+
+    public void testDisabled() {
```
SCWells72, line 172: And when disabled.

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPDocumentSymbolBreadcrumbsInfoProvider.java
angelozerr: LanguageServerregistry maintains a list of language where LS are associated in distinctLanguages. if you define a getter to this distinctLanguages, is it working for you?

src/test/java/com/redhat/devtools/lsp4ij/fixtures/LSPDocumentSymbolBreadcrumbsInfoProviderFixtureTestCase.java
```diff
@@ -0,0 +1,171 @@
...
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Base test fixture for the document symbol-based breadcrumbs info provider.
+ */
+@SuppressWarnings("unused")
+public abstract class LSPDocumentSymbolBreadcrumbsInfoProviderFixtureTestCase extends LSPCodeInsightFixtureTestCase {
```
SCWells72, line 43: And the common base class upon which tests for other languages could be easily created.

SCWells72:
@angelozerr this is ready for review now.

src/main/java/com/redhat/devtools/lsp4ij/client/features/EditorBehaviorFeature.java
angelozerr: What about having a new LSPBreadcrumbFeatures?

src/main/java/com/redhat/devtools/lsp4ij/client/features/EditorBehaviorFeature.java
```diff
@@ -181,4 +192,17 @@ public static boolean enableSemanticTokensFileViewProvider(@NotNull PsiFile file
...
+        return LanguageServiceAccessor.getInstance(file.getProject()).hasAny(
+                file.getVirtualFile(),
+                ls -> ls.getClientFeatures().getEditorBehaviorFeature().isEnableDocumentSymbolsBreadcrumbsInfoProvider(file)
```
angelozerr, line 205: If we have LSPBreadcrumbfeature, we could write:

` -> ls.getClientFeatures().getBreadcrumbFeature().isEnabled(file)`

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPDocumentSymbolBreadcrumbsInfoProvider.java
SCWells72: I'm having to reach back in my memory a few weeks, but the `getLanguages()` method is only called once for the `breadcrumbsInfoProvider`, so if language server definitions are configured at runtime, the IDE would have to be restarted to pick up those changes. This ensures that it's active for all languages, but I've registered it last in `plugin.xml` so that any more specific one for a given language would be used instead.

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPDocumentSymbolBreadcrumbsInfoProvider.java
angelozerr, line 42: To have a short name, I will rename to LSPBreadcrumbsProvider

docs/LSPApi.md
```diff
@@ -649,9 +649,18 @@ or, if appropriate, the the child/descendant element that represents the declara
 | boolean isSupported()       | Returns `true` if the LSP feature is supported and `false` otherwise. <br/>This supported state is called after starting the language server, which matches the LSP server capabilities. | Check the server capability            |
 | boolean supportsGotoClass() | Returns `true` if the LSP feature is efficient enough to support the IDE's Go To Class action which may be invoked frequently and `false` otherwise.                                     | `false`                                |
 
+## LSP Breadcrumbs Feature
```
SCWells72, line 652: New feature and feature flag doc.

docs/LSPSupport.md
```diff
@@ -690,6 +691,16 @@ You can also open the `Structure` view using the `View / Tool Windows / Structur
 
 ![textDocument/documentSymbol](./images/lsp-support/textDocument_documentSymbol.gif)
 
+#### Breadcrumbs and Sticky Lines
```
SCWells72, line 694: Subtopic for breadcrumbs/sticky lines.

src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPBreadcrumbsFeature.java
```diff
@@ -0,0 +1,65 @@
...
+import com.intellij.psi.PsiFile;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Breadcrumbs/sticky lines feature. This does not correspond to an actual LSP feature but is built upon
+ * {@link LSPDocumentSymbolFeature}.
+ */
+@ApiStatus.Experimental
+public class LSPBreadcrumbsFeature {
```
SCWells72, line 23: New feature for breadcrumbs.

src/main/java/com/redhat/devtools/lsp4ij/server/definition/launching/UserDefinedBreadcrumbsFeature.java
```diff
@@ -0,0 +1,44 @@
...
+import com.redhat.devtools.lsp4ij.client.features.LSPClientFeatures;
+import com.redhat.devtools.lsp4ij.server.definition.ClientConfigurableLanguageServerDefinition;
+import com.redhat.devtools.lsp4ij.server.definition.launching.ClientConfigurationSettings.ClientConfigurationBreadcrumbsSettings;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * User-defined breadcrumbs feature.
+ */
+public class UserDefinedBreadcrumbsFeature extends LSPBreadcrumbsFeature {
```
SCWells72, line 25: User-defined breadcrumbs feature based on client config.

src/main/resources/jsonSchema/clientSettings.schema.json
```diff
@@ -54,6 +54,19 @@
         }
       }
     },
+    "breadcrumbs": {
```
SCWells72, line 57: JSON schema update for client config feature and flag.

src/main/resources/templates/lemminx/clientSettings.json
```diff
@@ -2,6 +2,9 @@
   "caseSensitive": true,
   "blockCommentPrefix": "<!--",
   "blockCommentSuffix": "-->",
+  "breadcrumbs": {
```
SCWells72, line 5: Disabled by default for XML and HTML (below). Those languages already have first-class breadcrumbs support in all JetBrains IDEs.

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPDocumentSymbolStructureViewModel.java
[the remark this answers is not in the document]
SCWells72: I've now removed that. It's going to be critical that your fix go in to avoid IDE deadlocks.

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPDocumentSymbolBreadcrumbsInfoProvider.java
```diff
@@ -0,0 +1,151 @@
...
+    private boolean isSupported(@NotNull PsiElement element) {
+        Project project = element.getProject();
+        PsiFile file = element.getContainingFile();
+        VirtualFile virtualFile = file.getVirtualFile();
+        return LanguageServiceAccessor.getInstance(project).hasAny(
+                virtualFile,
+                ls ->
+                        // The document symbols feature must be available for the file's language server
+                        ls.getClientFeatures().getDocumentSymbolFeature().isEnabled(file) &&
+                                ls.getClientFeatures().getDocumentSymbolFeature().isSupported(file) &&
```
angelozerr, line 60: I will not use document symbol feature but LSPBreadcrumbsFeature implemented by default like this;

 * isEnabled returns true
 * isSupported implemented by calling getClientFeatures().getDocumentSymbolFeature().isSupported(file)

angelozerr (changes requested):

src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPBreadcrumbsFeature.java
angelozerr, line 23: Please extends abstraction class like other lsp feature

angelozerr (changes requested):

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPDocumentSymbolStructureViewModel.java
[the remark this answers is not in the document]
angelozerr: Thanks, lets see this problem in my pr.

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPBreadcrumbsProvider.java
```diff
@@ -0,0 +1,151 @@
...
+
+    private boolean isSupported(@NotNull PsiElement element) {
+        Project project = element.getProject();
+        PsiFile file = element.getContainingFile();
+        VirtualFile virtualFile = file.getVirtualFile();
+        return LanguageServiceAccessor.getInstance(project).hasAny(
+                virtualFile,
+                ls ->
+                        // The document symbols feature must be available for the file's language server
+                        ls.getClientFeatures().getDocumentSymbolFeature().isEnabled(file) &&
```
angelozerr, line 59: Please use only breadcrumb feature

angelozerr (changes requested):

src/test/java/com/redhat/devtools/lsp4ij/fixtures/LSPDocumentSymbolBreadcrumbsInfoProviderFixtureTestCase.java
```diff
@@ -0,0 +1,171 @@
...
```
angelozerr, line 43: Rename the test to match the new class short name

angelozerr (changes requested):

src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPBreadcrumbsFeature.java
```diff
@@ -0,0 +1,44 @@
...
+ * Breadcrumbs/sticky lines feature. This does not correspond directly to an LSP feature but is built upon
+ * {@link LSPDocumentSymbolFeature}.
+ */
+@ApiStatus.Experimental
+public class LSPBreadcrumbsFeature extends AbstractLSPDocumentFeature {
+
+    private ServerCapabilities serverCapabilities;
+
+    @Override
+    public void setServerCapabilities(@Nullable ServerCapabilities serverCapabilities) {
```
angelozerr, line 30: Implement this method as empty

src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPBreadcrumbsFeature.java
```diff
@@ -0,0 +1,44 @@
...
+        this.serverCapabilities = serverCapabilities;
+    }
+
+    public boolean isEnabled(@NotNull PsiFile file) {
```
angelozerr, line 34: Add Oveeride annotation

angelozerr (changes requested):

src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPBreadcrumbsFeature.java
angelozerr, line 34: Remove this method which is already defined in the extended class

angelozerr (changes requested):

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPBreadcrumbsProvider.java
```diff
@@ -0,0 +1,148 @@
...
+    @Override
+    public boolean acceptElement(@NotNull PsiElement element) {
+        // NOTE: This is only used to determine which already-vetted elements should be included in the breadcrumb trail
+        return element instanceof DocumentSymbolData;
+    }
+
+    @Override
+    @Nullable
+    public PsiElement getParent(@NotNull PsiElement element) {
+        if (isSupported(element)) {
```
angelozerr, line 72: I wonder if we should do that for each element, since document symbol belongs to the same file?

SCWells72: Assuming you mean using `CachedValuesManager` for caching, we'd need to add a modification tracker for language server configuration that causes a read-through if/when language server configs are changed. Otherwise cached values could become stale.

I'd probably lump this under the "potential premature optimization" category and would consider it if we observe performance issues around this check. I've not seen anything that concerns me even in large files.

angelozerr: I agree, it was just a question. I think we have no performance problem with this isSupported call

SCWells72: It does raise an interesting point, though. There might be opportunities (and need) for caching where cached elements should be evicted automatically when the associated language server definition config changes. If/when that's needed for sure, it would make sense to add a `ModificationTracker` implementation to `LanguageServerDefinition` that can be used as a cache dependency and is automatically incremented when that language server definition changes in any manner that should result in cache eviction.

Something to consider if/when it's needed...

angelozerr: I think more and more that we need to cache for all lsp features the isSupported in the psifile for each associated ls. See https://github.com/redhat-developer/lsp4ij/issues/7

docs/LSPApi.md
angelozerr, line 652: Teference this new feature on the top of this lsp api markdown file

angelozerr (changes requested):

angelozerr:
Thanks so much @SCWells72 !

Issue #871, opened 2025-02-28
Add support for custom semantic token types and modifiers


SCWells72:
We've discussed this a few times, but some language servers return semantic token types that aren't present in [`SemanticTokenTypes`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#semanticTokenTypes) and/or semantic token modifiers that aren't present in [`SemanticTokenModifiers`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#semanticTokenModifiers). Sometimes those custom token types/modifiers are actually quite important, e.g., `member` from the TypeScript language server, and if not understood explicitly by LSP4IJ, important functionality can be missing ranging from semantic syntax highlighting to the way that a given token should be translated into an element the PSI hierarchy.

It should be possible for language server definitions to augment the known semantic token types and modifiers via both code and configuration. One way to do this would be to add the following methods to `LSPSemanticTokensFeature`:

```java
/**
 * Returns custom semantic token types for the file.
 *
 * @param file the file
 * @return the file's custom semantic token types
 */
@NotNull
public Set<LSPSemanticTokenType> getCustomSemanticTokenTypes(@NotNull PsiFile file) {
    return Collections.emptySet();
}

/**
 * Returns custom semantic token modifiers for the file.
 *
 * @param file the file
 * @return the file's custom semantic token modifiers
 */
@NotNull
public Set<LSPSemanticTokenModifier> getCustomSemanticTokenModifiers(@NotNull PsiFile file) {
    return Collections.emptySet();
}
```

where `LSPSemanticTokenType` is defined as:

```java
/**
 * A custom semantic token type.
 */
public class LSPSemanticTokenType {
    /** The semantic token type name */
    public String name;
    /** Whether or not the semantic token should be interpreted as being an identifier */
    public boolean identifier = false;
    /** Whether or not the semantic token should be interpreted as being a type */
    public boolean type = false;
    /** Whether or not the semantic token should be interpreted as being a keyword */
    public boolean keyword = false;
    /** The name of the existing semantic token type from which this one should inherit its behavior */
    public String inheritFrom = null;
    /** The text attributes key that should be used for syntax highlighting of this semantic token */
    public String textAttributesKey = null;
}
```

and `LSPSemanticTokenModifier` is defined as:

```java
/**
 * A custom semantic token modifier.
 */
public class LSPSemanticTokenModifier {
    /** The semantic token modifier name */
    public String name = null;
[...]

PR #887, opened 2025-03-12
fix: Added (hard-coded) support for SourceKit LSP's `identifier` semantic token type which should be interpreted as being a declaration.

SCWells72:
This will need to be made configurable as part of #871.


angelozerr:
LGTM, thanks @SCWells72 !

PR #894, opened 2025-03-14
fix: check endPosition on window/showDocument

### LINKED ISSUE #892: window/showDocument doesn't respect selection end range
I think [here](https://github.com/redhat-developer/lsp4ij/blob/5d6fecff92d8adf3565cfa34f98ba16a636f06b9/src/main/java/com/redhat/devtools/lsp4ij/ServerMessageHandler.java#L181) should handle the end range, if start range is different than end range, it should use `SelectionModel` to make the range selected.

This affects clojure-lsp because I want to implement a command that uses a paredit feature which moves the cursor and select the text between parens.

ericdallo:
Fixes #892 

@angelozerr just tested this with clojure-lsp and it worked

### DIFF
--- a/src/main/java/com/redhat/devtools/lsp4ij/LSPIJUtils.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/LSPIJUtils.java
@@ -424,6 +424,7 @@ public class LSPIJUtils {
                 if (editor != null && endPosition != null && startPosition != endPosition) {
                     var endOffset = LSPIJUtils.toOffset(endPosition, document);
                     editor.getSelectionModel().setSelection(startOffset, endOffset);
+                    editor.getCaretModel().moveToOffset(endOffset);
                 }
                 return editor != null;
             }
--- a/src/main/java/com/redhat/devtools/lsp4ij/LSPIJUtils.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/LSPIJUtils.java
@@ -240,7 +240,8 @@ public class LSPIJUtils {
      * </p>
      *
      * @param fileUri            the file Uri to open.
-     * @param position           the position.
+     * @param startPosition      the start position.
+     * @param endPosition        the end position to make it selectable.
      * @param focusEditor        true if editor will take the focus and false otherwise.
      * @param createFileIfNeeded true if file must be created if doesn't exist and false otherwise.
      * @param fileUriSupport     the file Uri support.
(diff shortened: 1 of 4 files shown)


angelozerr:
Thanks so much @ericdallo, I see now how to that.

The main problem with your PR that it breaks API.

Let me create a PR by following your idea.

ericdallo:
@angelozerr I created a new openInEditor function, how it breaks the API? 🤔

angelozerr:
You have changed the LSPIJUtils openInEditor methods (by adding a new parameter endPosition). We should keep the existing method.

The idea is to keep teh existing method signature which must call your new method.

It would be nice also to have method with Range and use it directly.

I mean write:

`if (LSPIJUtils.openInEditor(uri, params.getSelection(), focusEditor, false, fileUriSupport, project)) {`

and check other LSP requst which have Range.

If you could do that it would be very nice, otherwise I will do it.

ericdallo:
@angelozerr we can't have a method with that args because it would be ambiguous with [this one](https://github.com/ericdallo/lsp4ij/blob/05796390a937540a42339d4faa3598706f9627a2/src/main/java/com/redhat/devtools/lsp4ij/LSPIJUtils.java#L251), so a function with startPosition and endPosition is the only thing I can see working

ericdallo:
I fixed the API break @angelozerr , LMK if that is enough

src/main/java/com/redhat/devtools/lsp4ij/LSPIJUtils.java
```diff
@@ -358,42 +380,68 @@ public static boolean openInEditor(@Nullable VirtualFile file,
...
+        if (startPosition == null) {
             return FileEditorManager.getInstance(project).openFile(file, true).length > 0;
         } else {
             if (document != null) {
-                OpenFileDescriptor desc = new OpenFileDescriptor(project, file, LSPIJUtils.toOffset(position, document));
-                return FileEditorManager.getInstance(project).openTextEditor(desc, focusEditor) != null;
+                OpenFileDescriptor desc = new OpenFileDescriptor(project, file, LSPIJUtils.toOffset(startPosition, document));
+                var startOffset = LSPIJUtils.toOffset(startPosition, document);
+                var editor = FileEditorManager.getInstance(project).openTextEditor(desc, focusEditor);
+                if (editor != null && endPosition != null && startPosition != endPosition) {
```
angelozerr, line 439: if ( editor != null && endPosition != null && !startPosition.equals(endPosition)) {

angelozerr:
> I fixed the API break @angelozerr , LMK if that is enough

Thanks!

angelozerr:
@ericdallo I wonder if this selection is standard with IJ. For instance when I do a ctrl+click on Java method, it doesn't select the method name, but just set the cursor.

ericdallo:
@angelozerr yeah I don't think the default should be select in every feature, but for sure the showDocument should handle it as it's what the LSP spec intends with `selection`

angelozerr:
Do you know how vscode manage this LSP request?

ericdallo:
No, but it's what emacs does, I'd expect the same since it's a Range and not a Position that this request supports

angelozerr:
> No, but it's what emacs does, I'd expect the same since it's a Range and not a Position that this request supports

LocationLink uses also Range.

angelozerr:
Definition can return a locationlink which defines a range. I wonder if we should select the target in the 3ditor in this case.

But it will not follow IJ behaviour. I we dont do that, is it consistent to select the target only in showDocument case?

If no perhaps we should control this behiour in some lsp features api.

angelozerr:
@ericdallo your PR looks good. Let merge it and we will see how to manage other LSP request like definition.

Thanks @ericdallo !

PR #908, opened 2025-03-19
fix: Ensure that breadcrumbs/sticky lines are queried for files in open editor tabs after a language server starts

SCWells72:
This addresses an issue with the new LSP-based sticky lines feature where the IDE would query the `breadcrumbsInfoProvider` for sticky line info before the language server has started and therefore before breadcrumbs are available.

It adds a `postStartupActivity` that is invoked when a project is first opened that registers a `LanguageServerLifecycleListener` so that, when a language server's status is changed to `started`, sticky lines for open editor tabs for that language server can be refreshed. This has to happen in a bit of an indirect manner, though. Rather than talk through it here, it's probably best to defer to the code itself.

### DIFF
--- a/src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPBreadcrumbsRefreshListener.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPBreadcrumbsRefreshListener.java
@@ -29,6 +29,7 @@ import com.intellij.xml.breadcrumbs.BreadcrumbsPanel;
 import com.redhat.devtools.lsp4ij.LSPIJUtils;
 import com.redhat.devtools.lsp4ij.LanguageServerWrapper;
 import com.redhat.devtools.lsp4ij.ServerStatus;
+import com.redhat.devtools.lsp4ij.client.features.LSPBreadcrumbsFeature;
 import com.redhat.devtools.lsp4ij.lifecycle.LanguageServerLifecycleListener;
 import com.redhat.devtools.lsp4ij.lifecycle.LanguageServerLifecycleManager;
 import kotlin.Unit;
@@ -56,28 +57,30 @@ public class LSPBreadcrumbsRefreshListener implements ProjectActivity, LanguageS
     @Override
     public void handleStatusChanged(LanguageServerWrapper languageServer) {
         if (languageServer.getServerStatus() == ServerStatus.started) {
-            Project project = languageServer.getProject();
-            for (VirtualFile virtualFile : FileEditorManager.getInstance(project).getOpenFiles()) {
-                PsiFile file = LSPIJUtils.getPsiFile(virtualFile, project);
-                if ((file != null) &&
-                    languageServer.getClientFeatures().getBreadcrumbsFeature().isSupported(file) &&
-                    languageServer.getClientFeatures().getBreadcrumbsFeature().isEnabled(file)) {
-                    for (Editor editor : LSPIJUtils.editorsForFile(virtualFile, project)) {
-                        BreadcrumbsPanel breadcrumbsComponent = BreadcrumbsPanel.getBreadcrumbsComponent(editor);
-                        if (breadcrumbsComponent != null) {
-                            // Only add a listener to the editor once
-                            if (editor.getUserData(ADDED_BREADCRUMB_LISTENER) == null) {
-                                breadcrumbsComponent.addBreadcrumbListener(new LSPBreadcrumbListener(file, editor), breadcrumbsComponent);
-                                editor.putUserData(ADDED_BREADCRUMB_LISTENER, true);
-                            }
+            ApplicationManager.getApplication().executeOnPooledThread(() -> {
+                Project project = languageServer.getProject();
+                LSPBreadcrumbsFeature breadcrumbsFeature = languageServer.getClientFeatures().getBreadcrumbsFeature();
 
-                            // Denote the file as needing restart and queue an update to its breadcrumbs
-                            file.putUserData(NEEDS_RESTART, true);
-                            ApplicationManager.getApplication().invokeLater(breadcrumbsComponent::queueUpdate);
+                for (VirtualFile virtualFile : FileEditorManager.getInstance(project).getOpenFiles()) {
+                    PsiFile file = LSPIJUtils.getPsiFile(virtualFile, project);
+                    if ((file != null) && breadcrumbsFeature.isSupported(file) && breadcrumbsFeature.isEnabled(file)) {
+                        for (Editor editor : LSPIJUtils.editorsForFile(virtualFile, project)) {
+                            BreadcrumbsPanel breadcrumbsComponent = BreadcrumbsPanel.getBreadcrumbsComponent(editor);
+                            if (breadcrumbsComponent != null) {
+                                // Only add a listener to the editor once
+                                if (editor.getUserData(ADDED_BREADCRUMB_LISTENER) == null) {
+                                    breadcrumbsComponent.addBreadcrumbListener(new LSPBreadcrumbListener(file, editor), breadcrumbsComponent);
+                                    editor.putUserData(ADDED_BREADCRUMB_LISTENER, true);
+                                }
+
+                                // Denote the file as needing restart and queue an update to its breadcrumbs
+                                file.putUserData(NEEDS_RESTART, true);
+                                ApplicationManager.getApplication().invokeLater(breadcrumbsComponent::queueUpdate);
+                            }
                         }
                     }
                 }
-            }
+            });
         }
     }
 
@@ -104,6 +107,8 @@ public class LSPBreadcrumbsRefreshListener implements ProjectActivity, LanguageS
             // If the file is still valid and needs restart, clear that flag and schedule restart
             if (file.isValid() && (file.getUserData(NEEDS_RESTART) == Boolean.TRUE)) {
                 file.putUserData(NEEDS_RESTART, false);
+                // We must force the modification stamp to increment or sticky lines won't be recomputed
+                file.clearCaches();
                 ApplicationManager.getApplication().invokeLater(() -> DaemonCodeAnalyzer.getInstance(file.getProject()).restart(file));
             }
         }
--- a/src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPBreadcrumbsRefreshListener.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPBreadcrumbsRefreshListener.java
@@ -1,0 +1,111 @@
+/*******************************************************************************
+ * Copyright (c) 2025 Red Hat Inc. and others.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+ * which is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ *
+ * Contributors:
+ *     Red Hat Inc. - initial API and implementation
+ *******************************************************************************/
+
+package com.redhat.devtools.lsp4ij.features.documentSymbol;
+
+import com.intellij.codeInsight.daemon.DaemonCodeAnalyzer;
+import com.intellij.openapi.application.ApplicationManager;
+import com.intellij.openapi.editor.Editor;
+import com.intellij.openapi.fileEditor.FileEditorManager;
+import com.intellij.openapi.project.Project;
+import com.intellij.openapi.startup.ProjectActivity;
+import com.intellij.openapi.util.Key;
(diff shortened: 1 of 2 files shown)


src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPBreadcrumbsRefreshListener.java
```diff
@@ -0,0 +1,111 @@
...
+import org.eclipse.lsp4j.jsonrpc.messages.Message;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Refreshes the breadcrumbs and sticky lines for open editor tabs when a language server has started.
+ */
+public class LSPBreadcrumbsRefreshListener implements ProjectActivity, LanguageServerLifecycleListener {
+
+    private static final Key<Boolean> ADDED_BREADCRUMB_LISTENER = Key.create(LSPBreadcrumbsRefreshListener.class + ".ADDED_BREADCRUMB_LISTENER");
```
SCWells72, line 46: Editor-level user data key for whether or not we've already added a `BreadcrumbListener` to the editor.

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPBreadcrumbsRefreshListener.java
```diff
@@ -0,0 +1,111 @@
...
+    private static final Key<Boolean> NEEDS_RESTART = Key.create(LSPBreadcrumbsRefreshListener.class + ".NEEDS_RESTART");
```
SCWells72, line 47: File-level user data key for whether or not the next `BreadcrumbListener#breadcrumbsChanged` event should result in restarting analysis on the file.

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPBreadcrumbsRefreshListener.java
```diff
@@ -0,0 +1,111 @@
...
+
+    @Override
+    @Nullable
+    public Object execute(@NotNull Project project, @NotNull Continuation<? super Unit> continuation) {
+        LanguageServerLifecycleManager.getInstance(project).addLanguageServerLifecycleListener(this);
```
SCWells72, line 52: When the project is opened, register this class as a `LanguageServerLifecycleListener` so that we can known when language servers have moved into a `started` status.

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPBreadcrumbsRefreshListener.java
```diff
@@ -0,0 +1,111 @@
...
+        return null;
+    }
+
+    @Override
+    public void handleStatusChanged(LanguageServerWrapper languageServer) {
```
SCWells72, line 57: Hopefully the majority of this is pretty straightforward, but basically when a language server has been started, find all open files for which it supports breadcrumbs and force their editor to re-query breadcrumbs -- which only helps for breadcrumbs and doesn't automatically update sticky lines -- and denote the associated file as needing analysis restart -- which does update sticky lines based on the latest breadcrumbs.

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPBreadcrumbsRefreshListener.java
```diff
@@ -0,0 +1,111 @@
...
+            for (VirtualFile virtualFile : FileEditorManager.getInstance(project).getOpenFiles()) {
+                PsiFile file = LSPIJUtils.getPsiFile(virtualFile, project);
+                if ((file != null) &&
+                    languageServer.getClientFeatures().getBreadcrumbsFeature().isSupported(file) &&
+                    languageServer.getClientFeatures().getBreadcrumbsFeature().isEnabled(file)) {
+                    for (Editor editor : LSPIJUtils.editorsForFile(virtualFile, project)) {
+                        BreadcrumbsPanel breadcrumbsComponent = BreadcrumbsPanel.getBreadcrumbsComponent(editor);
+                        if (breadcrumbsComponent != null) {
+                            // Only add a listener to the editor once
+                            if (editor.getUserData(ADDED_BREADCRUMB_LISTENER) == null) {
```
SCWells72, line 69: This ensures that we only register a listener on the editor once.

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPBreadcrumbsRefreshListener.java
```diff
@@ -0,0 +1,111 @@
...
+                                breadcrumbsComponent.addBreadcrumbListener(new LSPBreadcrumbListener(file, editor), breadcrumbsComponent);
+                                editor.putUserData(ADDED_BREADCRUMB_LISTENER, true);
+                            }
+
+                            // Denote the file as needing restart and queue an update to its breadcrumbs
+                            file.putUserData(NEEDS_RESTART, true);
```
SCWells72, line 75: And this tells the event listener that the next `#breadcrumbsChanged` event for the file should result in restarting analysis on the file.

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPBreadcrumbsRefreshListener.java
```diff
@@ -0,0 +1,111 @@
...
+                            ApplicationManager.getApplication().invokeLater(breadcrumbsComponent::queueUpdate);
```
SCWells72, line 76: It's important that we defer processing here and below via the EDT for proper ordering of events.

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPBreadcrumbsRefreshListener.java
```diff
@@ -0,0 +1,111 @@
...
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    @Override
+    public void handleLSPMessage(Message message, MessageConsumer consumer, LanguageServerWrapper languageServer) {
```
SCWells72, line 85: No-ops for the rest of the `LanguageServerLifecycleListener` interface.

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPBreadcrumbsRefreshListener.java
```diff
@@ -0,0 +1,111 @@
...
+    private record LSPBreadcrumbListener(PsiFile file, Editor editor) implements BreadcrumbListener {
+        private LSPBreadcrumbListener(@NotNull PsiFile file, @NotNull Editor editor) {
+            this.file = file;
+            this.editor = editor;
+        }
+
+        @Override
+        public void breadcrumbsChanged(@NotNull Iterable<? extends Crumb> crumbs) {
+            // If the file is still valid and needs restart, clear that flag and schedule restart
+            if (file.isValid() && (file.getUserData(NEEDS_RESTART) == Boolean.TRUE)) {
```
SCWells72, line 105: We know that we have new breadcrumbs, so if the file is still valid and is marked as needing analysis restarted, clear the "needs restart" flag and schedule analysis restart. That will result in `StickyLinesPass` being run and new sticky lines being collected based on the new breadcrumbs.

src/main/resources/META-INF/plugin.xml
```diff
@@ -318,6 +318,9 @@
                 id="LSPBreadcrumbsProvider"
                 implementation="com.redhat.devtools.lsp4ij.features.documentSymbol.LSPBreadcrumbsProvider"
                 order="last"/>
+        <postStartupActivity
```
SCWells72, line 321: Register the project post-startup activity for refreshing breadcrumbs.

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPBreadcrumbsRefreshListener.java
```diff
@@ -0,0 +1,111 @@
...
+    @Nullable
+    public Object execute(@NotNull Project project, @NotNull Continuation<? super Unit> continuation) {
+        LanguageServerLifecycleManager.getInstance(project).addLanguageServerLifecycleListener(this);
+        return null;
+    }
+
+    @Override
+    public void handleStatusChanged(LanguageServerWrapper languageServer) {
+        if (languageServer.getServerStatus() == ServerStatus.started) {
+            Project project = languageServer.getProject();
```
angelozerr, line 59: Please do that asynchronously (ex: in an NonBlockingReadAction). The server start must be fast and we must not block the start with some ReadAction.

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPBreadcrumbsRefreshListener.java
angelozerr: Instead of having 2 keys, I will have one key with a class which maintains info.

angelozerr:
After reading your code, I have the impression that it is the same topic than CodeVision,InlayHint, etc  which must be refreshed.

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPBreadcrumbsRefreshListener.java
SCWells72: No, these are purposefully stored at the correct levels. File and editors do not have a 1:1 relationship. Consider a split editor. The listener must be registered on the editor and therefore that flag must be stored there. Analysis happens on a per-file basis, so that flag must be stored there.

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPBreadcrumbsRefreshListener.java
SCWells72: Sure, no problem, though I would question why the event dispatcher isn't moving onto a background thread. Otherwise it's deferring a potential issue to any subscribing event listener.

SCWells72:
> After reading your code, I have the impression that it is the same topic than CodeVision,InlayHint, etc which must be refreshed.
> 
> It would be super nice if we could use the EditorFeatureManager

That's probably true, but those can likely be refreshed (approximately) synchronously with the language server being denoted as `started` whereas here we have to queue a refresh of the breadcrumb trail info for the editor explicitly and then queue an analysis refresh of the editor's file once we know that has completed to yield the correct results. The more general version of this is likely another language server lifecycle listener.

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPBreadcrumbsRefreshListener.java
```diff
@@ -0,0 +1,116 @@
...
+    @Nullable
+    public Object execute(@NotNull Project project, @NotNull Continuation<? super Unit> continuation) {
+        LanguageServerLifecycleManager.getInstance(project).addLanguageServerLifecycleListener(this);
+        return null;
+    }
+
+    @Override
+    public void handleStatusChanged(LanguageServerWrapper languageServer) {
+        if (languageServer.getServerStatus() == ServerStatus.started) {
+            ApplicationManager.getApplication().executeOnPooledThread(() -> {
```
SCWells72, line 60: Immediately move to a background thread to ensure that further event dispatch isn't slowed by this processing.

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPBreadcrumbsRefreshListener.java
```diff
@@ -0,0 +1,116 @@
...
+            this.editor = editor;
+        }
+
+        @Override
+        public void breadcrumbsChanged(@NotNull Iterable<? extends Crumb> crumbs) {
+            // If the file is still valid and needs restart, clear that flag and schedule restart
+            if (file.isValid() && (file.getUserData(NEEDS_RESTART) == Boolean.TRUE)) {
+                file.putUserData(NEEDS_RESTART, false);
+                // We must force the modification stamp to increment or sticky lines won't be recomputed
+                file.clearCaches();
```
SCWells72, line 111: This **must** happen because `StickyLinesPassFactory#createHighlightingPass()` checks that the file's and/or document's modification stamps has been updated before running a new pass.

src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPBreadcrumbsRefreshListener.java
SCWells72: Included in the latest pushed commit.

angelozerr:
LGTM, thanks @SCWells72 !

PR #909, opened 2025-03-20
fix: Ensured that the LSP4IJ usage element description provider doesn't return any value for non-LSP4IJ elements

SCWells72:


src/main/java/com/redhat/devtools/lsp4ij/usages/LSPUsageElementDescriptionProvider.java
```diff
@@ -53,6 +54,11 @@ public String getElementDescription(@NotNull PsiElement element, @NotNull Elemen
             }
         }
 
+        // If it's not one of ours, don't provide a description
+        if (!(element instanceof LSPPsiElement)) {
```
SCWells72, line 58: Should be pretty self-explanatory, but if we get an element that's not ours, don't return anything here.

angelozerr:
Thanks @SCWells72 !

PR #911, opened 2025-03-21
perf: optimize LSPDocumentSymbolUtils

FalsePattern:
This PR optimizes the LSPDocumentSymbolUtils lookup logic.
It's split up into 2 commits

1df4d698db35b76742843a45a223a6e7f04d6807:
- Extracts LSPIJUtils.getDocument(file) outside of the recursive walk inside getDocumentSymbolData (FileDocumentManager calls are very expensive)
- Adds a "prefetched document length" option, because `document.getTextLength()` is an expensive operation (walking the PSI tree in some cases)
These two changes decreased the profiler footprint about 10-15%

e6588cb0eba07a4eb1f381f584665791709a5fed:
- Replaces PSIElement.getChildren() with getChildrenBase() inside the recursive search inside `getContainingDocumentSymbolDatas`.
This skips a lot of PSI element filtering logic and decreased the profiler footprint by about 30-40% on my machine, and the stickly lines/breadcrumbs seem correct to me, but it needs more testing and can be dropped from this PR if deemed too aggressive


src/main/java/com/redhat/devtools/lsp4ij/LSPIJUtils.java
```diff
@@ -731,7 +731,11 @@ public static Range toRange(TextRange range, Document document) {
     }
 
     public static @Nullable TextRange toTextRange(@NotNull Range range, @NotNull Document document) {
-        return toTextRange(range, document, null, false);
+        return toTextRange(range, document, null, false, -1);
+    }
+
+    public static @Nullable TextRange toTextRange(@NotNull Range range, @NotNull Document document, int docLengthPrefetched) {
```
SCWells72, line 737: Please annotate this as `@ApiStatus.Internal`.

src/main/java/com/redhat/devtools/lsp4ij/LSPIJUtils.java
```diff
@@ -745,7 +749,13 @@ public static Range toRange(TextRange range, Document document) {
     public static @Nullable TextRange toTextRange(@NotNull Range range,
                                                   @NotNull Document document,
                                                   boolean adjust) {
-        return toTextRange(range, document, null, adjust);
+        return toTextRange(range, document, null, adjust, -1);
+    }
+    public static @Nullable TextRange toTextRange(@NotNull Range range,
```
SCWells72, line 754: Please restore the JavaDoc comment for this original signature since it's used externally.

src/main/java/com/redhat/devtools/lsp4ij/LSPIJUtils.java
```diff
@@ -759,11 +769,12 @@ public static Range toRange(TextRange range, Document document) {
     public static @Nullable TextRange toTextRange(@NotNull Range range,
```
SCWells72, line 769: And this signature can/should be `private` since it's only used from `LSPIJUtils`.

SCWells72 (changes requested):
Just a few requested changes around retaining the same `LSPIJUtils` API and not "leaking" the new signature that includes the precomputed length but is used externally.

SCWells72:
Also confirmed that breadcrumbs/sticky lines look fine with these changes.

SCWells72 (approved):
Looks great! Thanks for both the optimizations and the review feedback changes.

angelozerr:
Thanks @FalsePattern @SCWells72 !!!

PR #925, opened 2025-03-24
perf: Simplify access to client config information (fixes #912)

### LINKED ISSUE #912: Futures should not be required when gathering local-only information, e.g., client configuration
Right now to get client configuration -- and perhaps other local-only -- information, one must often use `CompleteableFuture`. This is both inefficient and can lead to potential locking issues, particularly if that information is needed from the EDT or under a read/write lock.

It's possible to avoid this for checking flags, e.g.:

```java
boolean rangeFormattingSupported = LanguageServiceAccessor.getInstance(project).hasAny(
    file.getVirtualFile(),
    languageServerWrapper -> languageServerWrapper.getClientFeatures().getFormattingFeature().isRangeFormattingSupported(file)
);
```

but if multiple pieces of information are required and/or non-flag information is required, that's at best inefficient.

It would be nice to be able to do something more like:

```java
boolean rangeFormattingSupported = false;
String formatOnCloseBrace = false;
String formatOnCloseBraceCharacters = "";
LanguageServiceAccessor.getInstance(project).processLanguageServers(
    file.getVirtualFile(),
    ls -> {
        LSPFormattingFeature formattingFeature = ls.getClientFeature().getFormattingFeature();
        rangeFormattingSupported |= formattingFeature.isRangeFormattingSupported(file);
        formatOnCloseBrace |= formattingFeature.isFormatOnCloseBrace(file);
        formatOnCloseBraceCharacters += formattingFeature.getFormatOnCloseBraceCharacters(file);
        // ...
    }
);
```

That's pseudocode because the values being set aren't mutable in the lambda, but it should convey the idea, and that issue can be solved by using `Ref` or similar.

I'm happy to look at implementing something like this and updating current usages that could benefit from such a change, ideally eliminating the need for `CompleteableFuture` for what should be simple client config usages like these.

SCWells72:
This implements the optimization suggested in #912 and updates client-side on-type formatting to use it.


src/main/java/com/redhat/devtools/lsp4ij/BooleanRef.java
```diff
@@ -0,0 +1,60 @@
...
+package com.redhat.devtools.lsp4ij;
+
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Reference to a boolean value that simplifies mutation of that value in lambdas/inner classes.
+ */
+@ApiStatus.Internal
+public class BooleanRef extends MutableRef<Boolean> {
```
SCWells72, line 21: Several mutable reference types to provide indirect access to a mutable value of the respective simple types in lambdas. For `boolean` values we want to be able to perform logical ANDs and ORs; for `string` values we want to be able to append; for enumerated constant values we want to be able to retain the highest/lowest-ordered value.

src/main/java/com/redhat/devtools/lsp4ij/MutableRef.java
```diff
@@ -0,0 +1,36 @@
...
+
+import com.intellij.openapi.util.Ref;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Reference to a value that simplifies mutation of that value in lambdas/inner classes.
+ *
+ * @param <T> the value type
+ */
+class MutableRef<T> extends Ref<T> {
```
SCWells72, line 22: Base class for the above mutable ref types.

src/main/java/com/redhat/devtools/lsp4ij/MutableRef.java
```diff
@@ -0,0 +1,36 @@
...
+        super(value);
+    }
+
+    /**
+     * Returns the value.
+     *
+     * @return the value
+     */
+    @Nullable
+    public T getValue() {
```
SCWells72, line 33: We have this so that subclasses can change from `@Nullable` to `@NotNull` as appropriate. Unfortunately `get()` is `final`.

src/main/java/com/redhat/devtools/lsp4ij/features/formatting/LSPClientSideOnTypeFormattingTypedHandler.java
```diff
@@ -50,120 +52,119 @@ public Result charTyped(char charTyped,
...
+        BooleanRef formatOnStatementTerminatorRef = BooleanRef.create();
+        ComparableRef<FormattingScope> formatOnStatementTerminatorScopeRef = ComparableRef.create(FormattingScope.STATEMENT);
+        StringRef formatOnStatementTerminatorCharactersRef = StringRef.create();
+        BooleanRef formatOnCompletionTriggerRef = BooleanRef.create();
+        StringRef formatOnCompletionTriggerCharactersRef = StringRef.create();
+        LanguageServiceAccessor.getInstance(project).processLanguageServers(
+                file,
+                ls -> {
+                    if (ls.getServerStatus() == ServerStatus.started) {
+                        LSPFormattingFeature formattingFeature = ls.getClientFeatures().getFormattingFeature();
```
SCWells72, line 69: Collect all of that info across all matching, started language servers.

src/main/java/com/redhat/devtools/lsp4ij/features/formatting/LSPClientSideOnTypeFormattingTypedHandler.java
```diff
@@ -50,120 +52,119 @@ public Result charTyped(char charTyped,
...
+                Character closeBraceChar = bracePair.getValue();
+                if (StringUtil.isEmpty(formatOnCloseBraceCharacters) ||
+                        (formatOnCloseBraceCharacters.indexOf(closeBraceChar) > -1)) {
+                    return handleCloseBraceTyped(
                             project,
                             editor,
                             file,
-                            formattingFeature,
-                            charTyped
+                            formatOnCloseBraceScope,
```
SCWells72, line 113: We were only passing `formattingFeature` to get to `formatOnCloseBraceScope`, so just pass that scope.

src/main/java/com/redhat/devtools/lsp4ij/features/formatting/LSPClientSideOnTypeFormattingTypedHandler.java
```diff
@@ -50,120 +52,119 @@ public Result charTyped(char charTyped,
...
+                // Make sure the formatter supports formatting of the configured scope
+                (formatOnStatementTerminatorScope != null) &&
+                ((formatOnStatementTerminatorScope == FormattingScope.FILE) || rangeFormattingSupported)) {
+            if (StringUtil.isNotEmpty(formatOnStatementTerminatorCharacters) &&
+                    (formatOnStatementTerminatorCharacters.indexOf(charTyped) > -1)) {
+                return handleStatementTerminatorTyped(
+                        project,
+                        editor,
+                        file,
+                        formatOnStatementTerminatorScope,
```
SCWells72, line 132: We were only passing `formattingFeature` to get to `formatOnStatementTerminatorScope`, so just pass that scope.

src/main/java/com/redhat/devtools/lsp4ij/features/formatting/LSPClientSideOnTypeFormattingTypedHandler.java
```diff
@@ -50,120 +52,119 @@ public Result charTyped(char charTyped,
...
+                        editor,
+                        file
+                );
             }
         }
 
         return super.charTyped(charTyped, project, editor, file);
     }
 
-    @Nullable
```
SCWells72, line 122: Now we can remove this.

src/main/java/com/redhat/devtools/lsp4ij/features/formatting/LSPClientSideOnTypeFormattingTypedHandler.java
```diff
@@ -50,120 +51,108 @@ public Result charTyped(char charTyped,
...
                     }
                 }
-            }
+        );
+        boolean rangeFormattingSupported = rangeFormattingSupportedRef.get();
 
-            // Statement terminators
-            if (formattingFeature.isFormatOnStatementTerminator(file) &&
+        // Close braces
+        if (onTypeFormattingSettings.formatOnCloseBrace &&
```
SCWells72, line 86: And everything below here is just updating to use that collected info. I recommend hiding whitespace-only changes for the best diff.

src/main/java/com/redhat/devtools/lsp4ij/features/formatting/LSPClientSideOnTypeFormattingTypedHandler.java
```diff
@@ -50,120 +51,108 @@ public Result charTyped(char charTyped,
...
+                Character closeBraceChar = bracePair.getValue();
+                if (StringUtil.isEmpty(onTypeFormattingSettings.formatOnCloseBraceCharacters) ||
+                        (onTypeFormattingSettings.formatOnCloseBraceCharacters.indexOf(closeBraceChar) > -1)) {
+                    return handleCloseBraceTyped(
                             project,
                             editor,
                             file,
-                            formattingFeature,
-                            charTyped
+                            onTypeFormattingSettings.formatOnCloseBraceScope,
```
SCWells72, line 102: We were previously passing `formattingFeature` here just to get the formatting scope, so now we just pass the formatting scope directly.

src/main/java/com/redhat/devtools/lsp4ij/features/formatting/LSPClientSideOnTypeFormattingTypedHandler.java
```diff
@@ -50,120 +51,108 @@ public Result charTyped(char charTyped,
...
+        if (onTypeFormattingSettings.formatOnStatementTerminator &&
+                // Make sure the formatter supports formatting of the configured scope
+                ((onTypeFormattingSettings.formatOnStatementTerminatorScope == FormattingScope.FILE) || rangeFormattingSupported)) {
+            if (StringUtil.isNotEmpty(onTypeFormattingSettings.formatOnStatementTerminatorCharacters) &&
+                    (onTypeFormattingSettings.formatOnStatementTerminatorCharacters.indexOf(charTyped) > -1)) {
+                return handleStatementTerminatorTyped(
+                        project,
+                        editor,
+                        file,
+                        onTypeFormattingSettings.formatOnStatementTerminatorScope,
```
SCWells72, line 120: Same as the above change to pass the scope instead of the formatting feature.

src/main/java/com/redhat/devtools/lsp4ij/server/definition/launching/ClientConfigurationSettings.java
```diff
@@ -103,7 +103,7 @@ public static class ClientSideOnTypeFormattingSettings {
          * The specific close brace characters that should trigger client-side on-type formatting. Defaults to the
          * language's close brace characters.
          */
-        public String formatOnCloseBraceCharacters = null;
+        public String formatOnCloseBraceCharacters = "";
```
SCWells72, line 106: Changed these to default to the empty string instead of null. Downstream consumers are just checking for an empty or null string, so it's a no-op other than that we don't have to check for null when collecting these from language server definitions before appending.

src/main/java/com/redhat/devtools/lsp4ij/features/formatting/LSPClientSideOnTypeFormattingTypedHandler.java
```diff
@@ -50,120 +51,108 @@ public Result charTyped(char charTyped,
...
-        LSPFormattingFeature formattingFeature = getFormattingFeature(file);
-        if (formattingFeature != null) {
-            boolean rangeFormattingSupported = formattingFeature.isRangeFormattingSupported(file);
+        // Gather all of the relevant client configuration
+        Ref<Boolean> rangeFormattingSupportedRef = Ref.create(false);
+        ClientSideOnTypeFormattingSettings onTypeFormattingSettings = new ClientSideOnTypeFormattingSettings();
+        LanguageServiceAccessor.getInstance(project).processLanguageServers(
+                file,
+                ls -> {
+                    if (ls.getServerStatus() == ServerStatus.started) {
```
angelozerr, line 60: I think this test should be done in the processLanguageServers to be compliant with hasAny.

src/main/java/com/redhat/devtools/lsp4ij/features/formatting/LSPClientSideOnTypeFormattingTypedHandler.java
angelozerr: You need to check that ls enable/support formatting features (and not on type formatting). See old hasLanguageServerSupportingOnlyFormatting

src/main/java/com/redhat/devtools/lsp4ij/features/formatting/LSPClientSideOnTypeFormattingTypedHandler.java
SCWells72: Done.

src/main/java/com/redhat/devtools/lsp4ij/features/formatting/LSPClientSideOnTypeFormattingTypedHandler.java
SCWells72: Done.

angelozerr: IMHO we should use hasAny and processLanguageServers both because you evaluate mappings twice. I wonder if your processLanguageServers should have a new parameter with filter.

SCWells72: 100% agreed. That was a sloppy oversight on my part. Running tests on a change to consolidate that check into the existing loop.

SCWells72: Done.

src/main/java/com/redhat/devtools/lsp4ij/features/formatting/LSPClientSideOnTypeFormattingTypedHandler.java
```diff
@@ -50,120 +52,112 @@ public Result charTyped(char charTyped,
...
+        Ref<Boolean> rangeFormattingSupportedRef = Ref.create(false);
+        ClientSideOnTypeFormattingSettings onTypeFormattingSettings = new ClientSideOnTypeFormattingSettings();
+        LanguageServiceAccessor.getInstance(project).processLanguageServers(
+                file,
+                ls -> {
+                    // Only include servers that support formatting and don't support server-side on-type formatting
+                    LSPClientFeatures clientFeatures = ls.getClientFeatures();
+                    LSPFormattingFeature formattingFeature = clientFeatures.getFormattingFeature();
+                    LSPOnTypeFormattingFeature onTypeFormattingFeature = clientFeatures.getOnTypeFormattingFeature();
+                    if (formattingFeature.isEnabled(file) && formattingFeature.isSupported(file) &&
```
SCWells72, line 65: All of these feature flags default to `false`/disabled, so if we don't find a language server that supports formatting but not server-side on-type formatting, everything below will be an effective no-op.

src/main/java/com/redhat/devtools/lsp4ij/features/formatting/LSPClientSideOnTypeFormattingTypedHandler.java
```diff
@@ -50,120 +52,112 @@ public Result charTyped(char charTyped,
                             @NotNull Project project,
                             @NotNull Editor editor,
                             @NotNull PsiFile file) {
-        LSPFormattingFeature formattingFeature = getFormattingFeature(file);
-        if (formattingFeature != null) {
-            boolean rangeFormattingSupported = formattingFeature.isRangeFormattingSupported(file);
+        // Gather all of the relevant client configuration
```
angelozerr, line 55: Ig yiu see the old getFormattingFeatures method it checks if project is indexing. Please add this check at first

angelozerr (changes requested):

src/main/java/com/redhat/devtools/lsp4ij/features/formatting/LSPClientSideOnTypeFormattingTypedHandler.java
SCWells72: Done.

angelozerr:
Thanks @SCWells72 !

PR #935, opened 2025-03-27
fix: Added caching of the file view provider's one file.

SCWells72:


src/main/java/com/redhat/devtools/lsp4ij/features/semanticTokens/viewProvider/LSPSemanticTokensFileViewProviderHelper.java
```diff
@@ -47,10 +51,39 @@ public LSPSemanticTokensFileViewProviderHelper(@NotNull LSPSemanticTokensFileVie
 
     @Nullable
     private PsiFile getFile() {
-        // There should only be one PSI file
-        List<PsiFile> allFiles = fileViewProvider.getAllFiles();
-        PsiFile file = allFiles.size() == 1 ? ContainerUtil.getFirstItem(allFiles) : null;
-        return (file != null) && file.isValid() && EditorBehaviorFeature.enableSemanticTokensFileViewProvider(file) ? file : null;
+        // Lazy-initialize the one PSI file for this view provider
```
SCWells72, line 54: There are two aspects to this cache. First, the one file is cached on the file view provider helper instance via DCL lazy-initialization with automatic read-through when the cached file becomes invalid.

src/main/java/com/redhat/devtools/lsp4ij/features/semanticTokens/viewProvider/LSPSemanticTokensFileViewProviderHelper.java
```diff
@@ -33,7 +33,33 @@
  */
 public class LSPSemanticTokensFileViewProviderHelper implements LSPSemanticTokensContainer {
 
+    /**
+     * Simple short-duration thread-local cache for the file view provider's PSI file to minimize read-through during
+     * burst reads of information from the file view provider.
+     */
+    private static final class CachedPsiFile {
```
SCWells72, line 40: This basically implements the recommendation [here](https://github.com/redhat-developer/lsp4ij/discussions/899#discussioncomment-12640287). The main differences are that an invalid cached PSI file will result in a cache read-through, and the start/duration times are measured in ms directly instead of ns which are then converted to ms. The cache is retained for a 1 sec duration, but that can be tuned if desired. Generally that should result in a good cache hit ratio for burst reads while being imperceptible if/when config changes.

src/main/java/com/redhat/devtools/lsp4ij/features/semanticTokens/viewProvider/LSPSemanticTokensFileViewProviderHelper.java
```diff
@@ -33,7 +33,33 @@
...
+            return isValid() ? file : null;
+        }
+
+        private boolean isValid() {
+            return file.isValid() && ((System.currentTimeMillis() - created) < MAX_DURATION);
+        }
+    }
+
     private final LSPSemanticTokensFileViewProvider fileViewProvider;
+    private final ThreadLocal<CachedPsiFile> cachedFilePtr = new InheritableThreadLocal<>();
```
SCWells72, line 62: Oh, and this is an inheritable thread local so that the cache can be leveraged if logic moves to another thread.

src/main/java/com/redhat/devtools/lsp4ij/features/semanticTokens/viewProvider/LSPSemanticTokensFileViewProviderHelper.java
```diff
@@ -169,8 +169,8 @@ public void addSemanticToken(@NotNull TextRange textRange,
         if (semanticTokensByOffset != null) {
             LSPSemanticToken semanticToken = new LSPSemanticToken(file, textRange, tokenType, tokenModifiers);
 
-            // Index the token for its text range
-            for (int offset = textRange.getStartOffset(); offset <= textRange.getEndOffset(); offset++) {
+            // Index the token for its text range up to but not including the end offset
+            for (int offset = textRange.getStartOffset(); offset < textRange.getEndOffset(); offset++) {
```
SCWells72, line 173: The real fix here is `<` instead of `<=`. Basically for a file that uses the semantic tokens file view provider, `file.getElementAt(offset)` should return an element for `[startOffset, endOffset)` where those are the start/end offsets from `element.getTextRange()`. Without this change, it was `[startOffset, endOffset]`, and that was causing an infinite loop in `XDebuggerUtilImpl` which was using `endOffset` as the new offset when searching for the next element.

I'm amazed this wasn't apparent before now!

angelozerr:
Thanks so much @SCWells72 !!!

PR #942, opened 2025-03-31
feat: Create a Docker Language Server template

rcjsuen:
Adding a new template to help users quickly configure the Docker Language Server in a JetBrains IDE.


angelozerr:
Thanks so much @rcjsuen !

Is your language server can be configured with settings.json? If yes do you have a JSON Schema?

I noticed that you have a DAP server in vscode-docker, have you tested the DAP support from LSP4IJ. 

If yes, I think it can be a good idea to add it also as template.

I merge your PR, if you need to improve it, don't hesitate to contribute again.

Thanks!

rcjsuen:
> Is your language server can be configured with settings.json? If yes do you have a JSON Schema?

Yeah, I did see that schema file in the other templates but wasn't totally sure what it was exactly configuring. I need to compare the JSON-RPC traffic with VS Code and JetBrains to make sure I'm implementing things by the specification correctly.

Is `settings.json` for `workspace/configuration` and `workspace/didChangeConfiguration`?

> I noticed that you have a DAP server in vscode-docker, have you tested the DAP support from LSP4IJ.
> 
> If yes, I think it can be a good idea to add it also as template.

Interesting. I didn't realize Microsoft was using the DAP there. I will do some digging, thanks for the pointer!

> I merge your PR, if you need to improve it, don't hesitate to contribute again.

Thanks, Angelo!

angelozerr:
For JSON Schema yo should see in https://github.com/microsoft/vscode-docker/blob/main/package.json which should define some helpfull JSON Schema property to configure the docker language server.

If you write a JSON Schema, you should benefir with completion, validation hover in the  Configuration JSON text editor

PR #952, opened 2025-04-03
fix: Docker documentation typos and bullet ordering

rcjsuen:
1. I did some copy/paste and the numbers incorrectly started at 3 instead of 1. :P
2. The language server is called "Docker Language Server" and not "Dockerfile Language Server".


angelozerr:
Thanks @rcjsuen !

PR #953, opened 2025-04-03
feat: Restoring external references search as an opt-in feature that's disabled by default.

SCWells72:
Hopefully this one should be quite straightforward, but I'll add a few comments in the PR itself.

Here's an example of **Find Usages** with the feature flag disabled per its default setting:

image

and here it is explicitly enabled via client config picking up an external reference to TypeScript in an HTML file:

image

### DIFF
--- a/docs/LSPSupport.md
+++ b/docs/LSPSupport.md
@@ -135,6 +135,31 @@ This menu action either opens the reference in a popup or navigates to the refer
 
 [textDocument/references](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_references) is used also via [Find Usages](./UserGuide.md#find-usages) to show references.
 
+#### External References
+
+LSP4IJ supports integration of externally-added references to language server-derived symbols when appropriate. External
+references are those that are added client-side by either the same plugin or by other installed plugins, e.g., in a
+polyglot environment where symbols in one language can be referenced in other languages.
+
+When enabled, this ensures that external references are included in **Find Usages** results and during symbol rename
+operations. This feature is _disabled by default_ but can be enabled for a custom language server implementation by
+overriding `LSPReferencesFeature#processExternalReferences(PsiFile)` to return `true`, or for a user-defined language
+server definition, by setting `references.processExternalReferences` to `true` in client configuration, e.g.:
+
+```json
+{
+  "references": {
+    "processExternalReferences": true
+  }
+}
+```
+
+This feature should **only** be enabled if you are **explicitly** aware of plugins that might be contributing 
+client-side external references to a language server's symbols. Note that enabling this feature will result in a
+textual search, albeit indexed, of the search scope for the symbol name for which references have been requested.
+This can result in longer overall references search times in larger projects and/or where the language server is
+supplementary to first-class IDE language support.  
+
 ### Implementation
 
 The [textDocument/implementation](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_implementation) is consumed with:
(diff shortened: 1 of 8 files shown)


docs/LSPSupport.md
```diff
@@ -135,6 +135,31 @@ This menu action either opens the reference in a popup or navigates to the refer
 
 [textDocument/references](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_references) is used also via [Find Usages](./UserGuide.md#find-usages) to show references.
 
+#### External References
```
SCWells72, line 138: Added documentation for this opt-in feature including how/when/why it can be enabled.

src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPReferencesFeature.java
```diff
@@ -62,5 +62,16 @@ public void setServerCapabilities(@Nullable ServerCapabilities serverCapabilitie
             referencesCapabilityRegistry.setServerCapabilities(serverCapabilities);
         }
     }
-    
+
+    /**
```
SCWells72, line 66: It's exposed as a client-side references feature that's _disabled by default_.

src/main/java/com/redhat/devtools/lsp4ij/server/definition/launching/ClientConfigurationSettings.java
```diff
@@ -167,6 +167,16 @@ public static class ClientConfigurationFormatSettings {
         public @NotNull OnTypeFormattingSettings onTypeFormatting = new OnTypeFormattingSettings();
     }
 
+    /**
+     * Client-side references settings.
+     */
+    public static class ClientConfigurationReferencesSettings {
```
SCWells72, line 173: And can be configured via client config for user-defined language server definitions. It's _disabled by default_ here as well.

src/main/java/com/redhat/devtools/lsp4ij/usages/LSPExternalReferencesFinder.java
```diff
@@ -70,22 +70,29 @@ public static void processExternalReferences(@NotNull PsiFile file,
...
-                if (StringUtil.isNotEmpty(wordText)) {
-                    processExternalReferences(
-                            file,
-                            wordText,
-                            wordTextRange,
-                            searchScope,
-                            ProgressManager.getInstance().getProgressIndicator(),
-                            processor
-                    );
+        // Only if enabled by at least one of the file's language servers
```
SCWells72, line 73: If you look at this with whitespace-only changes hidden, you'll see that this logic is now gated by a feature flag check. This ensures that **all** callers must pass this gate including `LSPUsageSearcher` and `LSPRenameRefactoringDialog` which wasn't disabled for 0.12.0.

src/main/java/com/redhat/devtools/lsp4ij/usages/LSPUsageSearcher.java
```diff
@@ -140,16 +140,13 @@ public void processElementUsages(@NotNull PsiElement element, @NotNull Processor
                 LOGGER.error("Error while collection LSP Usages", e);
             }
 
-            // TODO : as this external reference process can be very slow (for our IJ Quarkus);
-            // the call of processExternalReferences is commented.
-            //  Reactivate it just for some language servers (by using an lsp client featurees)
             // For completeness' sake, also collect external usages to LSP (pseudo-)elements
-            /*LSPExternalReferencesFinder.processExternalReferences(
+            LSPExternalReferencesFinder.processExternalReferences(
```
SCWells72, line 144: Restored this to an enabled state since the invoked method encapsulates the feature flag check.

src/main/resources/jsonSchema/clientSettings.schema.json
```diff
@@ -192,6 +192,19 @@
         }
       }
     },
+    "references": {
```
SCWells72, line 195: And finally, updated the client config JSON schema for this feature flag.

Note that I have not automatically enabled this feature _anywhere_ including the TypeScript language server definition template. Unless explicitly enabled somewhere, this feature is inactive. Since I know that my own plugin can contribute external references to JavaScript, TypeScript, and CSS, I will enable it in the integration managed by my plugin. That will limit the potential risk around restoring this while making it available to integrations where it's known to be valuable.

angelozerr:
LGTM thanks @SCWells72 !

PR #963, opened 2025-04-14
docs: Add Huly Code to the 'Using' section of LSP4IJ

kosz78:


angelozerr:
Thanks so much @kosz78 !

Could you please add a review in https://plugins.jetbrains.com/plugin/23257-lsp4ij/versions to help us to promote LSP4IJ. Thanks!

I have noticed that you have forked our LSP4IJ, and I would like to really to have a commons LSP4IJ repository. 

I'm integrating your nice fixes/improvements (but not the whole, because we need to discuss about that).

Let's continue this discussion at https://github.com/hcengineering/huly-code/issues/17

PR #967, opened 2025-04-15
Fix order in DAPStackFrame presentation

n1kolasM:
Fix order in the presentation of a stack frame, currently it is `<source_position> at <frame_name>`, should be `<frame_name> at <source_position>`


angelozerr:
Here a screenshot of theJav adebugger:

image

Here the LSP4IJDAP debugger (without your PR)

image

IMHO I think we should do like Java debugger:

Instead of `<frame_name> at <source_position>:<line>`, do 

`<frame_name>:<line>, <source_position>`

src/main/java/com/redhat/devtools/lsp4ij/dap/client/DAPStackFrame.java
```diff
@@ -54,19 +54,19 @@ public DAPStackFrame(@NotNull DAPClient client,
 
     @Override
     public void customizePresentation(@NotNull ColoredTextContainer component) {
+        component.append(stackFrame.getName(), SimpleTextAttributes.REGULAR_BOLD_ATTRIBUTES);
+        int line = stackFrame.getLine();
+        if (line > 0) {
+            component.append(":" + line, SimpleTextAttributes.REGULAR_ATTRIBUTES);
+        }
+        component.append(", ", SimpleTextAttributes.REGULAR_ATTRIBUTES);
```
angelozerr, line 62: Comma must be add only if sourceName exists.

angelozerr (changes requested):

src/main/java/com/redhat/devtools/lsp4ij/dap/client/DAPStackFrame.java
angelozerr: Oh sorry ! You are right.

PR #970, opened 2025-04-16
fix: Default TextMate enter-between-braces behavior test expectation fixes for 2024.1+

SCWells72:
The default TextMate enter-between-braces behavior differs between 2023.3 and 2024.1+. In 2023.3, typing enter between braces/brackets/parens such as:

```
values: [<typeEnterHere>]
```

would result in:

```
values: [
]
```
wheres in 2024.1-2025.1 (and presumably going forward), it results in:

```
values: [
    ]
```

That was causing a few tests that confirm the default TextMate enter-between-braces behavior when the enhanced enter-between-braces handler is disabled to fail.

I've updated these test expectations so that it's based on the host IDE version. Once 2023.3 support is dropped from LSP4IJ, this branching logic based on host IDE version can be removed and the 2024.1+ test expectations can be used.

### DIFF
--- a/src/main/java/com/redhat/devtools/lsp4ij/LSPIJUtils.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/LSPIJUtils.java
@@ -16,6 +16,7 @@ import com.intellij.lang.LanguageUtil;
 import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.application.ReadAction;
 import com.intellij.openapi.application.WriteAction;
+import com.intellij.openapi.application.ex.ApplicationInfoEx;
 import com.intellij.openapi.editor.Document;
 import com.intellij.openapi.editor.Editor;
 import com.intellij.openapi.editor.EditorFactory;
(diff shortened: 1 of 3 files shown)


src/main/java/com/redhat/devtools/lsp4ij/LSPIJUtils.java
```diff
@@ -1434,4 +1437,82 @@ public static String getProjectUri(Project project) {
...
+    // IDE version check utilities to support branching logic based on the host IDE version
+
+    /**
+     * Determines whether or not the host IDE is at least the specified major/minor version.
+     *
+     * @param minimumMajorVersion the major version number
+     * @param minimumMinorVersion the minor version number
+     * @return true if the host IDE is at least the specified major/minor version; otherwise false
+     */
+    public static boolean isVersionAtLeast(
```
SCWells72, line 1449: I ported the utility methods from my plugin to check for minimum host IDE version. These are useful for when logic has to branch -- or only occur -- in a minimum IDE version when the plugin supports versions older than that as well.

angelozerr:
Thanks so much @SCWells72 !

PR #1003, opened 2025-04-28
docs: Add telemetry configuration and documentation for the Docker Language Server

rcjsuen:
Document that the telemetry in the Docker Language Server is opt-out by default but can be configured as the user wishes during the initial setup of the language server.

Did I setup the `initializationOptions.schema.json` schema file correctly...? 🤔


angelozerr:
If I remember initializationOptions Json editor cannot be managed with Json Schema

angelozerr:
Let's merge it even if Json Schema is not implemented

angelozerr:
Thanks!

PR #1006, opened 2025-04-29
fix: Include compose.yaml and compose.yml to the Docker Language Server

rcjsuen:
The default path for Compose is `compose.yaml` and `compose.yml` so we should add those to the Docker Language Server as well.


angelozerr:
Thanks

PR #1026, opened 2025-05-09
docs(readme): Add bloc linter plugin on 'Who is using LSP4IJ'

vmichalak:
I'm here to resolve our issue : https://github.com/felangel/bloc/issues/4478 🫣
Thanks a lot for your work on LSP4IJ.


vmichalak:
CI seems to fail but that does not seem to be linked to my commit 🫣

angelozerr:
> CI seems to fail but that does not seem to be linked to my commit 🫣

Indeed I have this problem since afternoon. Let's merge your PR.

PR #1039, opened 2025-05-13
chore: Update build.gradle.kts to support latest version

sbouchet:


angelozerr:
I think with 2024.? there is the same problem (we need to add this json bundle)

sbouchet:
it's only for 2024.3 and newer versions. before that i don't remember the json module exists, thus it will fail the build.

angelozerr:
> it's only for 2024.3 and newer versions. before that i don't remember the json module exists, thus it will fail the build. 

My idea is to avoid adding future version of IJ like 2026 each year

sbouchet:
> > it's only for 2024.3 and newer versions. before that i don't remember the json module exists, thus it will fail the build.
> 
> My idea is to avoid adding future version of IJ like 2026 each year

yes got it, but as soon target version on [gradle.properties](https://github.com/redhat-developer/lsp4ij/blob/d85679f5f787d73d9ba0f4e0b70df0f220994cbc/gradle.properties#L16) is at least 2024.3, we can get rid of this clause and put json module in dependencies.

For now, as we support (target) version from 2023.3 to latest (2025.1 for now) adding 2026 to this clause is necessary ONLY for the latest IJ workflow build to pass and check source code can compile and passes tests.

angelozerr:
Thanks @sbouchet !

PR #1045, opened 2025-05-17
feat: Add Chinese localization for Language Server and DAP properties

lin-qingying:
feat: Add Chinese language support

Added Simplified Chinese language support for the LSP4IJ plugin, including:

- Added DAPBundle_zh_CN.properties file with Chinese translations for Debug Adapter Protocol related content
- Added LanguageServerBundle_zh_CN.properties file with Chinese translations for Language Server related content


angelozerr:
I cannot read Chinese, I trust you.

Thanks @lin-qingying !

PR #1060, opened 2025-05-26
build: implement a version check using Integer

sbouchet:


sbouchet:
yes as soon GitHub will stop hanging :D

PR #1073, opened 2025-05-29
test: address FileSystemWatcherManagerTest lint warnings

jomof:
It's a minor, test-only change, just something that was bugging me.


jomof:
I'm trying to get lsp4ij building as part of a larger bazel build which has lint standards enforced. I could set up a baseline with the warnings but I wondered whether I could just give you guys a PR to fix them? In IntelliJ, these show also show up as yellow squiggled.

PR #1087, opened 2025-06-03
fix: Correct the name of the Docker Language Server

rcjsuen:
1. There is another language server named the [Dockerfile Language Server](https://github.com/rcjsuen/dockerfile-language-server). This is _not_ that.
2. This language server provides features for Compose files and Bake files as well so this language server is _not_ only for Dockerfiles.


angelozerr:
Thanks @rcjsuen !

rcjsuen:
> https://github.com/redhat-developer/lsp4ij/tree/main/src/main/resources/templates/lsp/dockerfile-language-server should be also renamed, no?

Oh yeah, sorry. I guess I created it incorrectly in the beginning, sorry about that.

PR #1099, opened 2025-06-06
feat: added separation of application output consoles and DAP debugger consoles

Ant-ON:
It can be quite difficult to find the application output in the DAP debugger output. Added separation to two switchable consoles.

Checked on LLDB-DAP.

### DIFF
--- a/src/main/java/com/redhat/devtools/lsp4ij/dap/DAPDebugProcess.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/dap/DAPDebugProcess.java
@@ -33,11 +33,11 @@ import com.intellij.xdebugger.frame.XStackFrame;
 import com.intellij.xdebugger.frame.XSuspendContext;
 import com.intellij.xdebugger.ui.XDebugTabLayouter;
 import com.redhat.devtools.lsp4ij.dap.breakpoints.DAPBreakpointHandler;
-import com.redhat.devtools.lsp4ij.dap.breakpoints.DAPExceptionBreakpointsPanel;
 import com.redhat.devtools.lsp4ij.dap.client.DAPClient;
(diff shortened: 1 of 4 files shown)


angelozerr:
Thanks so much for your contribution!

Could you add a screenshot please to see what is the result of your PR.

angelozerr:
@Ant-ON I am sorry I don't understand your PR. It seems you have add a new toolbar `Show Toolbar Log` on the left:

image

when I click on it I see that:

image

We loose for instance the console.log used in JavaScrprit. (ex: if you see my first screenshot, you see foo in the lsp console), we should see that in your second console.

angelozerr:
> Checked on LLDB-DAP.

Oh I think understand the problem, you see horrible DAP traces in the console, is that?

If it that it is because LLDB-DAP uses stdio and I have not found a solution to avoid displaying content from process in the console (we need to write it for socket).

Ant-ON:
@angelozerr Yes, you are right. LLDB-DAP and GDB in DAP mode output a lot of information to stdout. My fix shows only output from the [Output Events](https://microsoft.github.io/debug-adapter-protocol//specification.html#Events_Output) to the primary console. Everything that is getting from stdin is displayed in the second console.

I can send screenshots tomorrow of what it looks like with LLDB-DAP.

angelozerr:
@Ant-ON I think the proper fix is to find a solution to not display stdout on the console. 

When we will find a solution for that, how to distinguish between DAP stdout and program stdout (via a print method)

Ant-ON:
@angelozerr In order to remove stdout from the console output, it is enough to override the `attachToProcess` method in the `DAPConsoleView` class.

ps I thought stdout was a need and transferred it to a separate console. Because there may be server startup messages / errors that it only outputs to the console.

PR #1108, opened 2025-06-10
feat: remove stdout messages from DAP console

Ant-ON:
Continue #1099. This changes completely remove stdout output from the DAP console.

Tested on LLDB-DAP and VSCode JS Debug

изображение

### DIFF
--- a/src/main/java/com/redhat/devtools/lsp4ij/dap/console/DAPConsoleView.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/dap/console/DAPConsoleView.java
@@ -12,6 +12,7 @@ package com.redhat.devtools.lsp4ij.dap.console;
 
 import com.intellij.execution.impl.ConsoleViewImpl;
 import com.intellij.execution.impl.EditorHyperlinkSupport;
+import com.intellij.execution.process.ProcessHandler;
 import com.intellij.openapi.actionSystem.ActionManager;
 import com.intellij.openapi.actionSystem.AnAction;
 import com.intellij.openapi.editor.actions.ScrollToTheEndToolbarAction;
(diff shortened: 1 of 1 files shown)


angelozerr:
One question, if you do a print in your script, do you see the result in the console? It is my fear that we loose this feature.

Ant-ON:
@angelozerr Sorry, I made a typo. Yes, I tested it on LLDP-DAP and everything works there. CLANGD is an LSP server.
Also tested with GDB in the DAP mode. There are some problems with DAP protocol (launching works, debugging does not). I don't know on which side the problem.

Okay, I add new DAP template.

If I print to stdout in my C application, I see it in the console. If I do `console.log("message")` in JS script, I see it in console.

Ant-ON:
@angelozerr  #1099 is more tolerant, in that it transfers the log from stdout to another console.

angelozerr:
> @angelozerr Sorry, I made a typo. Yes, I tested it on LLDP-DAP and everything works there. CLANGD is an LSP server. Also tested with GDB in the DAP mode. There are some problems with DAP protocol (launching works, debugging does not). I don't know on which side the problem.

I suggest that you create an issue about that.

> 
> Okay, I add new DAP template.

Thanks!

> 
> If I print to stdout in my C application, I see it in the console. If I do `console.log("message")` in JS script, I see it in console.

JS script uses socket, so it doesn't use process stout I think. I would like to know if with stdio (like LLDB) it is working.

angelozerr:
> @angelozerr #1099 is more tolerant, in that it transfers the log from stdout to another console.

I would like to avoid having 2 consoles. It is hard for user to know that there are 2 consoles.

angelozerr:
@Ant-ON don't loose your time, I am working on a proper fix.

angelozerr:
Replaced with https://github.com/redhat-developer/lsp4ij/pull/1115

Ant-ON:
@angelozerr yes, it work fine. Thank you!

angelozerr:
Thanks for your feedback.

Please note that now you can implement a server installer. I need to write doc

PR #1138, opened 2025-06-20
chore: revert using fixed version of pluginVerifier

sbouchet:


angelozerr:
Thanks so much @sbouchet !

PR #1164, opened 2025-06-29
feat: add support for integer IDs in JSON-RPC messages

Coobaha:
This pull request introduces support for integer IDs in JSON-RPC messages. This change enhances the flexibility and compliance of JSON-RPC message handling by allowing integers to be used as valid identifiers, in addition to any existing formats.

Specifically, it was failing with Elixir's `next-ls`, which only expected integer IDs.

```
Untitled
org.eclipse.lsp4j.jsonrpc.ResponseErrorException: Invalid request from the client

Received: %{"id" => "1", "jsonrpc" => "2.0", "method" => "initialize"....}
Errors: %{"id" => "expected an integer"}
```

image

### DIFF
--- a/src/main/java/com/redhat/devtools/lsp4ij/LanguageServerWrapper.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/LanguageServerWrapper.java
@@ -45,6 +45,7 @@ import com.redhat.devtools.lsp4ij.lifecycle.LanguageServerLifecycleManager;
 import com.redhat.devtools.lsp4ij.lifecycle.NullLanguageServerLifecycleManager;
 import com.redhat.devtools.lsp4ij.server.*;
 import com.redhat.devtools.lsp4ij.server.capabilities.TextDocumentServerCapabilityRegistry;
+import com.redhat.devtools.lsp4ij.server.definition.ClientConfigurableLanguageServerDefinition;
 import com.redhat.devtools.lsp4ij.server.definition.LanguageServerDefinition;
 import com.redhat.devtools.lsp4ij.settings.ProjectLanguageServerSettings;
 import com.redhat.devtools.lsp4ij.settings.ServerTrace;
@@ -371,10 +372,11 @@ public class LanguageServerWrapper implements Disposable {
 
                         // Get the useIntegerIds setting for this language server
                         boolean useIntegerIds = false;
-                        var settings = ProjectLanguageServerSettings.getInstance(initialProject)
-                                .getLanguageServerSettings(serverDefinition.getId());
-                        if (settings != null) {
-                            useIntegerIds = settings.isUseIntegerIds();
+                        if (serverDefinition instanceof ClientConfigurableLanguageServerDefinition clientConfigurableDefinition) {
+                            var clientConfiguration = clientConfigurableDefinition.getLanguageServerClientConfiguration();
+                            if (clientConfiguration != null) {
+                                useIntegerIds = clientConfiguration.jsonRpc.useIntegerIds;
+                            }
                         }
 
 
--- a/src/main/java/com/redhat/devtools/lsp4ij/LanguageServerWrapper.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/LanguageServerWrapper.java
@@ -46,13 +46,16 @@ import com.redhat.devtools.lsp4ij.lifecycle.NullLanguageServerLifecycleManager;
 import com.redhat.devtools.lsp4ij.server.*;
 import com.redhat.devtools.lsp4ij.server.capabilities.TextDocumentServerCapabilityRegistry;
 import com.redhat.devtools.lsp4ij.server.definition.LanguageServerDefinition;
-import com.redhat.devtools.lsp4ij.settings.ServerTrace;
 import com.redhat.devtools.lsp4ij.settings.ProjectLanguageServerSettings;
+import com.redhat.devtools.lsp4ij.settings.ServerTrace;
 import org.eclipse.lsp4j.*;
 import org.eclipse.lsp4j.jsonrpc.Launcher;
 import org.eclipse.lsp4j.jsonrpc.MessageConsumer;
 import org.eclipse.lsp4j.jsonrpc.messages.Either;
+import org.eclipse.lsp4j.jsonrpc.messages.IdentifiableMessage;
 import org.eclipse.lsp4j.jsonrpc.messages.Message;
+import org.eclipse.lsp4j.jsonrpc.messages.RequestMessage;
+import org.eclipse.lsp4j.jsonrpc.messages.ResponseMessage;
 import org.eclipse.lsp4j.services.LanguageServer;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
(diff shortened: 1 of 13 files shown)


angelozerr:
Thanks @Coobaha for your contribution.

First your new settings should be hosted in client configuration settings like https://github.com/redhat-developer/lsp4ij/blob/main/src/main/resources/templates/lsp/typescript-language-server/clientSettings.json

You will need to change https://github.com/redhat-developer/lsp4ij/blob/main/src/main/resources/jsonSchema/clientSettings.schema.json

Instead of copying the message to update the id, I wonder if LSP4J himself can manage this configuration?

Have the intention to contribute to user defined ls with next-ls? If yes please read https://github.com/redhat-developer/lsp4ij/blob/main/docs/UserDefinedLanguageServerTemplate.md#how-to-contribute-a-user-defined-language-server-template and it would be nice to have also an installer.json

It seems that Helixir ls provide also a DAP server https://github.com/elixir-lsp/elixir-ls?tab=readme-ov-file#elixirls---elixir-language-server-and-debug-adapter it would be very nice if you could contribute with user defined dap, see sample at https://github.com/redhat-developer/lsp4ij/tree/main/src/main/resources/templates/dap

src/main/java/com/redhat/devtools/lsp4ij/LanguageServerWrapper.java
```diff
@@ -366,16 +370,32 @@ public synchronized void start() throws LanguageServerException {
                             initParams.setRootPath(rootURI.getPath());
                         }
 
+                        // Get the useIntegerIds setting for this language server
+                        boolean useIntegerIds = false;
+                        if (serverDefinition instanceof ClientConfigurableLanguageServerDefinition clientConfigurableDefinition) {
```
angelozerr, line 375: Thanks for having switched to client configuration. The next step now id to manage any language server definition: user-defined and from extension plugin (ls  embedded in an IJ plugin).

The main idea is to not use ClientConfigurableLanguageServerDefinition which is specific for user defined ls, but use LSPClieantFeatures. In other words you should write a code like this (perhaps the method name is not very good)

`boolean useIntegerIds  =  getClientFeatures().isUseJsonRpcIntegerIds()`

To implement this method isUseJsonRpcIntegerIds, I suggets that you see LSPCLientFeatures#getBlock#getCommentPrefix and impl in UserDefinedClientFeatures#getCommentPrefix

angelozerr:
@Coobaha I think we should avoid copying messages. Here a suggestion (I have not tested) to manage this id with a clean solution.

First the problem comes from https://github.com/eclipse-lsp4j/lsp4j/blob/783b6e788c1e374b6de63f69804c7d0f96be4da2/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/RemoteEndpoint.java#L182 where id is hard coded as String.

The idea of my suggestion is to configure this code.

Here my suggestion:

1. Defines LSPClientFeatures#isUseIntAsJsonRpcId() which returns false by default and override it in UserDefinedClientFeatures to use your CientConfigSettings.

1. Create the class `RemoteEndpointWithIdAsInt`  like this:

```java
package com.redhat.devtools.lsp4ij.server.definition;

import org.eclipse.lsp4j.jsonrpc.Endpoint;
import org.eclipse.lsp4j.jsonrpc.MessageConsumer;
import org.eclipse.lsp4j.jsonrpc.RemoteEndpoint;
import org.eclipse.lsp4j.jsonrpc.messages.RequestMessage;
import org.eclipse.lsp4j.jsonrpc.messages.ResponseError;

import java.util.concurrent.atomic.AtomicInteger;
import java.util.function.Function;

class RemoteEndpointWithIdAsInt extends RemoteEndpoint {

    private final AtomicInteger nextRequestId = new AtomicInteger();

    public RemoteEndpointWithIdAsInt(MessageConsumer out,
                                     Endpoint localEndpoint,
                                     Function<Throwable, ResponseError> exceptionHandler) {
        super(out, localEndpoint, exceptionHandler);
    }

    public RemoteEndpointWithIdAsInt(MessageConsumer out,
                                     Endpoint localEndpoint) {
        super(out, localEndpoint);
    }

    @Override
    protected RequestMessage createRequestMessage(String method, Object parameter) {
        RequestMessage requestMessage = new RequestMessage();
        // Use int as JSON-RPC id instead of String (by default from LSP4J)
        requestMessage.setId(nextRequestId.incrementAndGet());
        requestMessage.setMethod(method);
        requestMessage.setParams(parameter);
        return requestMessage;
    }
}

```

Replace [LanguageServerDefinition#createLauncherBuilder](https://github.com/redhat-developer/lsp4ij/blob/05c06d3e156530e2c45d58aacb3fd03358f513df/src/main/java/com/redhat/devtools/lsp4ij/server/definition/LanguageServerDefinition.java#L205) with:

```java
public <S extends LanguageServer> Launcher.Builder<S> createLauncherBuilder() {

        return new Launcher.Builder<S>() {
            @java.lang.Override
            protected RemoteEndpoint createRemoteEndpoint(MessageJsonHandler jsonHandler) {

                boolean useIntAsId = createClientFeatures().isUseIntAsJsonRpcId();
                if (!useIntAsId) {
                    // Use JSON-RPC as String (default behavior of LSP4J)
                    return super.createRemoteEndpoint(jsonHandler);
                }

                // Override the remote endpoint to use JSON-RPC id as int
                MessageConsumer outgoingMessageStream = new StreamMessageConsumer(output, jsonHandler);
                outgoingMessageStream = wrapMessageConsumer(outgoingMessageStream);
                Endpoint localEndpoint = ServiceEndpoints.toEndpoint(localServices);
                RemoteEndpoint remoteEndpoint;
                if (exceptionHandler == null)
                    remoteEndpoint = new RemoteEndpointWithIdAsInt(outgoingMessageStream, localEndpoint);
                else
                    remoteEndpoint = new RemoteEndpointWithIdAsInt(outgoingMessageStream, localEndpoint, exceptionHandler);
                jsonHandler.setMethodProvider(remoteEndpoint);
                return remoteEndpoint;
            }
        }.configureGson(builder -> {
                    // Add a custom CodeLensOptionsAdapter to support old language server
                    // which declares codeLenProvider with a boolean instead of Json object.
                    builder.registerTypeAdapter(CodeLensOptions.class, new CodeLensOptionsAdapter());
                });
    }
```

And that's it. You don't need to update LanguageServerWrapper java class.

Please try that.

angelozerr:
> > @Coobaha I think we should avoid copying messages. Here a suggestion (I have not tested) to manage this id with a clean solution.
> > Please try that.
> 
> Thank you so much for the valuable insights and feedback! I will get back to this pr later this week and will change to the version you proposed 🙏

You are welcome!

And do you think you would like to contribute to user defined ls with Helixir LSP and DAP  ?

Coobaha:
> > > @Coobaha I think we should avoid copying messages. Here a suggestion (I have not tested) to manage this id with a clean solution.
> > > Please try that.
> > 
> > 
> > Thank you so much for the valuable insights and feedback! I will get back to this pr later this week and will change to the version you proposed 🙏
> 
> You are welcome!
> 
> And do you think you would like to contribute to user defined ls with Helixir LSP and DAP ?

I think you got confused between Elixir and Helix 😅. 

We don't need to set them up now, as three Elixir LSP servers are merging into a single official version. I'll be happy to contribute and create a configuration once the new official server is released, but they don't have any ETA. There's also a native IntelliJ plugin for Elixir, but it lacks some LSP features, which led me to improve support for lsp4ij for one of the Elixir LSP servers.

src/main/java/com/redhat/devtools/lsp4ij/server/definition/LanguageServerDefinition.java
```diff
@@ -211,6 +250,37 @@ public <S extends LanguageServer> Launcher.Builder<S> createLauncherBuilder() {
                 });
     }
 
+    public <S extends LanguageServer> Launcher.Builder<S> createLauncherBuilder(LSPClientFeatures clientFeatures) {
```
angelozerr, line 253: * Please remove the old `createLauncherBuilder()`
* use NotNull annotation: `createLauncherBuilder(NotNull LSPClientFeatures clientFeatures) `

Coobaha: Done 🙇

angelozerr: Thanks!

angelozerr:
Could you add your new method doc at https://github.com/redhat-developer/lsp4ij/blob/main/docs/LSPApi.md in the list.

angelozerr:
LGTM, thanks so much @Coobaha !

angelozerr:
@Coobaha you can install nighty https://github.com/redhat-developer/lsp4ij?tab=readme-ov-file#testing-nightly-builds to enjoy with your contribution

angelozerr:
> > > > @Coobaha I think we should avoid copying messages. Here a suggestion (I have not tested) to manage this id with a clean solution.
> > > > Please try that.
> > > 
> > > 
> > > Thank you so much for the valuable insights and feedback! I will get back to this pr later this week and will change to the version you proposed 🙏
> > 
> > 
> > You are welcome!
> > And do you think you would like to contribute to user defined ls with Helixir LSP and DAP ?
> 
> I think you got confused between Elixir and Helix 😅.

My bad sorry -(

> 
> We don't need to set them up now, as three Elixir LSP servers are merging into a single official version. I'll be happy to contribute and create a configuration once the new official server is released, but they don't have any ETA. There's also a native IntelliJ plugin for Elixir, but it lacks some LSP features, which led me to improve support for lsp4ij for one of the Elixir LSP servers.

It is a good new for LSP4IJ. DO you want to create an IJ plugin or just contribute to user defined ls?

And if you like LSP4IJ please add a review in JetBrains Marketplace to help us to promote LSP4IJ.

Issue #584, opened 2024-10-25
Inconsistent diagnostics behaviour due to range


rcjsuen:
1. If a language server specifies `1<<32 - 1` (4294967295) for `diagnostic.range.end.character` then it does not appear to work. I guess JetBrains still does not support this per [this other claim from 2019](https://github.com/microsoft/language-server-protocol/issues/257#issuecomment-496148251). I understand that no consensus has been reached there but could this "known issue" be documented somewhere in this repository?

2. If my diagnostics array has a) one valid diagnostic with a "non-crazy" range and b) a second diagnostic with said crazy range, neither of them will show up _and_ I do not see a `Received notification 'textDocument/publishDiagnostics'` message in the "Traces" section. Why would the traces be filtered out? 🤔

angelozerr:
I have tried to manage invalid range like vscode does.

Any contribution are welcome to improve it.

rcjsuen:
I think I understand the issue now. The `character` must be a `uinteger` which the specification defines as `2^31 - 1` (2147483647) which is lower than 4294967295. So the conversion code is probably raising an exception or becoming a negative number somewhere. That's an issue with the language server which can be corrected on that end this but it does not quite explain why the "Traces" section is not showing the JSON-RPC notification at all. 🤔

angelozerr:
Is not a problem coming from LSP himselft. Perhaps your issue is similar to https://github.com/redhat-developer/lsp4ij/issues/579 ?

rcjsuen:
Thanks Angelo. Yeah, my guess is your comment [here](https://github.com/redhat-developer/lsp4ij/issues/579#issuecomment-2419446384) which I guess explains why the "Traces" section is empty.

Could we at least get a "Failed to deserialize JSON-RPC message" in the "Traces" section or something? That would have helped with debugging.

PR #1169, opened 2025-07-03
fix: using integer max value as line end indicator

mpeltonen:
Some LSP servers use Integer.MAX_VALUE as a special character position value to indicate "end of line" in position coordinates. LSPIJUtils.toOffset was not correctly handling this, which lead to for example received diagnostics notifications to be silently ignored.

### DIFF
--- a/src/main/java/com/redhat/devtools/lsp4ij/LSPIJUtils.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/LSPIJUtils.java
@@ -658,6 +658,10 @@ public class LSPIJUtils {
         int lineOffset = document.getLineStartOffset(line);
         int nextLineOffset = document.getLineEndOffset(line);
         // If the character value is greater than the line length it defaults back to the line length
+        // Handle the common LSP pattern where Integer.MAX_VALUE indicates "end of line" as a special case.
+        if (character == Integer.MAX_VALUE) {
+            return nextLineOffset;
+        }
         return Math.max(Math.min(lineOffset + character, nextLineOffset), lineOffset);
     }
 
--- a/src/test/java/com/redhat/devtools/lsp4ij/LSPIJUtils_toOffsetTest.java
+++ b/src/test/java/com/redhat/devtools/lsp4ij/LSPIJUtils_toOffsetTest.java
@@ -74,6 +74,11 @@ public class LSPIJUtils_toOffsetTest extends BasePlatformTestCase {
         assertOffset("foo\nbar", 999999, 999999, 7, null);
     }
 
+    public void testIntegerMaxValueAsLineEnding() {
+        assertOffset("foo\nbar", 0, Integer.MAX_VALUE, 3, null);
+        assertOffset("foo\nbar", 1, Integer.MAX_VALUE, 7, null);
+    }
+    
     public void test_vscode_EmptyContent() {
         // See https://github.com/microsoft/vscode-languageserver-node/blob/8e625564b531da607859b8cb982abb7cdb2fbe2e/textDocument/src/test/textdocument.test.ts#L18
         String str = "";


angelozerr:
@mpeltonen thanks for your contribution.

Your PR looks good but could you give some links about language server which uses that?

rcjsuen:
This sounds like https://github.com/redhat-developer/lsp4ij/issues/584 and https://github.com/microsoft/vscode-languageserver-node/issues/1574...? 🤔

mpeltonen:
I came across this when implementing IntelliJ plugin for Garmin's MonkeyC language server which is included in their ConnectIQ SDK and this is the only one I know for sure: https://developer.garmin.com/connect-iq/sdk/

Started wondering why some of the diagnostic notifications were not visible in IntelliJ's "Problems" view, and further investigation led to this.

The "common pattern" is probably an overstatement and came from my discussions with ChatGPT about the issue. 

From ChatGPT:

> 1.	Eclipse JDT Language Server (Java)
> 	•	Uses Integer.MAX_VALUE as a sentinel value to indicate end-of-document in various LSP-related operations, particularly when mapping positions beyond the last line.
> 	•	Source example: org.eclipse.lsp4j.Position objects being constructed with line = Integer.MAX_VALUE for EOF operations.
> 2.	VSCode TypeScript Language Server
> 	•	Not in the core typescript-language-server, but in extensions or middleware handling, Number.MAX_SAFE_INTEGER (JS equivalent) is used to indicate an EOF-style position, which gets translated to Integer.MAX_VALUE in some language-bridging layers (e.g., Theia, Java-based LSP clients).
> 3.	Theia (Eclipse project)
> 	•	Some Java backends using LSP for Theia encode Integer.MAX_VALUE when calculating document ranges that are meant to represent an open-ended range, especially for formatting or text replacement.
> 4.	Google Java Format Language Server
> 	•	When applying formatting edits, uses Integer.MAX_VALUE to indicate edits that extend to the end of the file, especially in its TextEdit or Range wrappers for LSP compliance.
> 5.	SonarLint Language Server
> 	•	Uses Integer.MAX_VALUE as part of range construction for whole-file diagnostics or fixes, especially when the file length is unknown at the time of range creation.
> 
> All cases exploit Integer.MAX_VALUE to bypass or denote special behavior in line/character-based document models of LSP. The claim is a synthesis from training data encompassing community discussions, partial implementations, and tooling integration layers rather than direct canonical usage in primary language server codebases.

mpeltonen:
Hmm, this might very well fix #584, but cannot be 100% sure. 

The crux of the problem/fix is that if the end position is Integer.MAX_VALUE, then in `LSPIJUtils.toTextRange` this can cause null return (because of an overflow, start > end):

```
int start = LSPIJUtils.toOffset(range.getStart(), document);
int end = LSPIJUtils.toOffset(range.getEnd(), document);
if (start > end || end > docLength) {
    // Language server reports invalid range, ignore it.
    return null;
}
```

And there are more than one places like this in code:

```
TextRange range = LSPIJUtils.toTextRange(...)
if (range == null) {
    // Language server reports invalid diagnostic, ignore it.
    return;
}
```

mpeltonen:
Thanks for a great plugin! 

Will do those things after my plugin is in releasable state (and released)! 😎

rcjsuen:
Just to close the loop here, I updated from 0.14.0 to 0.14.3-20250705-112934 but I don't think it helped with #584. The server is sending this notification but I don't see the diagnostics in LSP4IJ.

And to be clear again as aforementioned in https://github.com/redhat-developer/lsp4ij/issues/584#issuecomment-2437641078,  I recognize that servers should _not_ be sending this value (as it is greater than the defined `uinteger`).
```JSON
{
    "uri": "file:///tmp/compose.yaml",
    "version": 2,
    "diagnostics": [
        {
            "range": {
                "start": {
                    "line": 3,
                    "character": 6
                },
                "end": {
                    "line": 3,
                    "character": 4294967295
                }
            },
            "severity": 1,
            "source": "docker-language-server",
            "message": "unexpected key name"
        }
    ]
}
```

PR #1176, opened 2025-07-09
fix: Changed "StyleLint" to the correct case "Stylelint"

SCWells72:
The correct tool name is "Stylelint" without a capital "L". This corrects the name in all places where it was wrong.


angelozerr:
Thanks @SCWells72 !

PR #1235, opened 2025-08-25
fix: adds immediate document saving after applying text edits to trig…

### LINKED ISSUE #1225: didChangeWatchedFiles is not sent immediately to LS after applying edits from rename response
### Environment

- IDE: IntelliJ IDEA Community Edition 2025.2
- Plugin: lsp4ij 0.14.2
- OS: macOS 

### Description

When performing a rename that affects multiple files (many of which are not opened in the IDEA), the subsequent _find usages_ returns only usages from the files that are currently opened. After a significant delay (or after opening the git commit window), the missing usages start to appear.

### Preconditions

The symbol to be renamed has usages in several files, including files that are not opened in IDEA.

**Steps to Reproduce**

- Open any file in IDEA containing the symbol to be renamed.
- Run _find usages_  on that symbol and note the total count.
- Perform _rename_ for that symbol. The language server returns edits that modify multiple files (including unopened files).
- Immediately run _find usages_ again.

### Actual Result

The second _find usages_ lists usages only in opened files (or a subset), so the total count is smaller than before the rename.

After ~60 seconds (in my tests), or earlier if I open the Git Commit window to view pending changes, additional usages appear and the count increases.

### Expected Result

The second Find Usages should match the first Find Usages count (pre-rename), reflecting all updated files right away, even if those files are not opened in the editor.

### Technical Notes

The edits returned by the language server are applied to files on disk.
However, didChangeWatchedFiles request are not sent to LS immediately.

Also it reproduces with clojure via: https://github.com/clojure-lsp/clojure-lsp-intellij

sergeysavchuk:
Ensures immediate VFS events when LSPIJUtils.applyEdits is used on unopened files

Added document saving after applying text edits to trigger VFS events right away.

Tested on Clojure LSP plugin:

Before fix:

https://github.com/user-attachments/assets/41194789-72a9-45fd-a24e-c61484033d7d

After fix:

https://github.com/user-attachments/assets/ccbba016-c69c-4e22-acfa-fbd97ce0581f

Fixes #1225

### DIFF
--- a/src/main/java/com/redhat/devtools/lsp4ij/LSPIJUtils.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/LSPIJUtils.java
@@ -1212,6 +1212,8 @@ public class LSPIJUtils {
         if (newCaretOffset > -1 && oldCaretOffset != newCaretOffset) {
             editor.getCaretModel().moveToOffset(newCaretOffset);
         }
+
+        FileDocumentManager.getInstance().saveDocument(document);
     }
 
     /**
--- a/src/main/java/com/redhat/devtools/lsp4ij/LSPIJUtils.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/LSPIJUtils.java
@@ -1213,6 +1213,9 @@ public class LSPIJUtils {
             editor.getCaretModel().moveToOffset(newCaretOffset);
         }
 
+        // Explicit document save is required to trigger LSPFileListener#contentsChanged immediately for files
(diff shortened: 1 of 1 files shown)


angelozerr:
Let me try it.

Could you add please a comment which explains your line code

sergeysavchuk:
@angelozerr I've added a comment, please review, e.g. before to trigger it, you can exec File -> "Save all" (cmd+S)

angelozerr:
LGTM, thanks!

PR #1278, opened 2025-09-18
fix: override getLanguage in LSPUsagePsiElement to return file language

### LINKED ISSUE #1277: Allow overriding of FindUsagesProvider for specific languages
At the moment, `com.redhat.devtools.lsp4ij.usages.LSPUsagePsiElement` inherits the default implementation of `com.intellij.psi.impl.FakePsiElement#getLanguage`, which always returns `Language.ANY`.  
This prevents IntelliJ from correctly resolving language-specific `FindUsagesProvider` implementations.  

#### Actual behavior  
- `LSPUsagePsiElement#getLanguage()` returns `Language.ANY`  
- As a result, the default `FindUsagesProvider` is used, ignoring the actual language of the containing file.  
- - Because of this, it is currently impossible to properly support binary files in `com.intellij.usages.UsageInfo2UsageAdapter#getPlainText`.  
  Example:  

  ```java
  if (element != null && isNullOrBinary) {
      return clsType(element) + " " + clsName(element);
  }
  ```

#### Expected behavior  
- `LSPUsagePsiElement#getLanguage()` should return the language of the containing file.  
- This will enable IntelliJ to use the appropriate `FindUsagesProvider` for that language.  

#### Proposed fix  
Override the method in `LSPUsagePsiElement`:  

```java
@Override
public @NotNull Language getLanguage() {
    return getContainingFile().getLanguage();
}
```

sergeysavchuk:
Allow overriding of FindUsagesProvider for specific languages

Fixes #1277

### DIFF
--- a/src/main/java/com/redhat/devtools/lsp4ij/usages/LSPUsagePsiElement.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/usages/LSPUsagePsiElement.java
@@ -10,6 +10,7 @@
  ******************************************************************************/
 package com.redhat.devtools.lsp4ij.usages;
 
+import com.intellij.lang.Language;
 import com.intellij.openapi.util.TextRange;
 import com.intellij.psi.PsiFile;
 import com.redhat.devtools.lsp4ij.features.LSPPsiElement;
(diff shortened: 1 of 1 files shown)


angelozerr:
I wonder if the fix should be done in LSPPsiElement ?

sergeysavchuk:
@angelozerr possibly, but each subclass should be reviewed separately to ensure correct `getLanguage()` behavior and tested.  At the moment I only faced this specific case with `LSPUsagePsiElement`, so the fix is focused on it.

PR #1289, opened 2025-09-21
fix: wait to start server during language server restart

IntegerLimit:
This PR makes the following change to the restart function in LSP Wrapper:
- Waits for server to stop before starting

This fixes an edge case discovered whilst working on [GroovyScript Plugin](https://github.com/IntegerLimit/GroovyScriptPlugin): due to the server taking long to process the `shutdown` request, stopping the server takes until the timeout runs out (5 seconds), resulting in the following sequence of events if the server was restarted:
- Server `stop` called
- Server `start` called, server starts
- Connection Provider `stop` called, server stops again

Furthermore, even if the server is in the `stopped` state, starting the server via the console would always result in these events. The only method of recovering from this (prior to this fix) is to disable and re-enable the language server through external means (IDE restart, settings change).

This PR is confirmed to fix the above issue.

### DIFF
--- a/src/main/java/com/redhat/devtools/lsp4ij/LanguageServerWrapper.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/LanguageServerWrapper.java
@@ -253,7 +253,7 @@ public class LanguageServerWrapper implements Disposable {
         numberOfRestartAttempts = 0;
         serverError = null;
         setEnabled(true);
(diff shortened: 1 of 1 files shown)


angelozerr:
Please wait again.... It is not finished.

IntegerLimit:
Yes, that seems to have fixed the issue. Thanks!

IntegerLimit:
Closed as superseeded by https://github.com/redhat-developer/lsp4ij/commit/9ff4d64117995dfd4337984fe768c71cce865bcd.

PR #1296, opened 2025-09-26
fix: fix warnings from JDT (mostly removed unused stuff)

fbricon:


angelozerr:
Thanks so much @fbricon !

PR #1297, opened 2025-09-26
fix: Rename template id to match the Markdown file

rcjsuen:
LSP4IJ uses the id in `template.json` to map to the Markdown file when using the "Need help with install, config, or usage? Click here" link so we need the id to be `buildx-dockerfile` so it opens the `buildx-dockefile.md` file.


angelozerr:
Thanks @rcjsuen

Issue #1298, opened 2025-09-26
Environment variables of a DAP launch config are ignored


rcjsuen:
Our debug adapter relies on a build environment to be set. If I set it on my system and try to override it with the wrong value it still works.

<img width="774" height="316" alt="Image" src="https://github.com/user-attachments/assets/4e28ce29-015f-4622-8b54-8c6d0158774a" />

rcjsuen:
I'm not sure if things are being merged properly. I have 0.17.0-20251003-113003 installed and I set `BUILDX_EXPERIMENTAL=0` to try to _disable_ the feature but the debugger still runs when it fails in the command line because `dap` is unrecognized.

<img width="226" height="77" alt="Image" src="https://github.com/user-attachments/assets/c9c10dcb-dff6-4b00-90ba-4040ed1eb9ac" />

```
$ BUILDX_EXPERIMENTAL=0 docker buildx dap
Usage:  docker buildx [OPTIONS] COMMAND

Extended build capabilities with BuildKit

Options:
      --builder string   Override the configured builder instance
  -D, --debug            Enable debug logging

Management Commands:
  history     Commands to work on build records
  imagetools  Commands to work on images in registry

Commands:
  bake        Build from a file
  build       Start a build
  create      Create a new builder instance
  dial-stdio  Proxy current stdio streams to builder instance
  du          Disk usage
  inspect     Inspect current builder instance
  ls          List builder instances
  prune       Remove build cache
  rm          Remove one or more builder instances
  stop        Stop builder instance
  use         Set the current builder instance
  version     Show buildx version information

Run 'docker buildx COMMAND --help' for more information on a command.

Experimental commands and flags are hidden. Set BUILDX_EXPERIMENTAL=1 to show them.
ERROR: unknown command: "dap"
```

angelozerr:
> I'm not sure if things are being merged properly. I have 0.17.0-20251003-113003 installed and I set `BUILDX_EXPERIMENTAL=0` to try to _disable_ the feature but the debugger still runs when it fails in the command line because `dap` is unrecognized.
> 
[...]

PR #1315, opened 2025-10-03
docs: Ask the user to explicitly set the BUILDX_EXPERIMENTAL variable

rcjsuen:
Until the environment variable issues have been resolved, it will be safer to ask the user to set the variable explicitly system-wide so that it will be picked up by the IDE internally as well.


rcjsuen:
> Is not enough to define this env var in the run config?

It didn't work and is why I opened #1298.

> I dont know if you have seen but now you can define env in the template.json

Yes, I did notice you made that change but wasn't quite sure of its implications.

> I did that for your dap server. Please verify by recreating the dap server that value is correct and please adjust your doc by saying that this value is set by the template

It _does_ seem to be working now. I'll create a separate pull request after doing some more testing. Sorry for the noise.

PR #1346, opened 2025-10-24
feat: in scala/metals, isHttpEnabled should be true by default. Also fixed an inconsistency between template & installer

tribbloid:


angelozerr:
Double quoted support path which have spaces

src/main/resources/templates/lsp/metals/template.json
```diff
@@ -2,8 +2,8 @@
   "id": "metals",
   "name": "Scala Language Server (Metals)",
   "programArgs": {
-    "windows": "\"${BASE_DIR}/metals.bat\"",
-    "default": "\"${BASE_DIR}/metals\""
+    "windows": "${workingDir}/metals.bat",
```
angelozerr: Why you did this change in this PR? What it is workingDir?

I suppose you want to reference metals file from the project dir, right?

If it that it should be $PROJECT_DIR$

angelozerr:
Thanks @tribbloid !

PR #1356, opened 2025-11-01
feat: Update Chinese localization for DAP and Language Server properties

lin-qingying:


angelozerr:
Thanks @lin-qingying !

PR #1378, opened 2025-11-27
fix: use Path.toUri() to handle WSL paths correctly

tim-sh:
The previous implementation used `File.toURI()`, which incorrectly encodes the server name of a UNC path (including WSL paths like `\\wsl$\...`) into the URI's path component. This resulted in invalid URIs with four leading slashes (e.g., `file:////wsl$/...`).

This change updates the logic to use the modern `java.nio.file.Path.toUri()` method. This correctly parses the server name (e.g., `wsl$`) into the URI's authority component, producing a valid and standards-compliant URI (e.g., `file://wsl$/...`).

Also:
- add a new test case to verify the correct handling of WSL paths.
- add the JUnit Jupiter engine dependency to the build to enable the execution of the test.

### DIFF
--- a/src/test/java/com/redhat/devtools/lsp4ij/LSPIJUtils_toUriTest.java
+++ b/src/test/java/com/redhat/devtools/lsp4ij/LSPIJUtils_toUriTest.java
@@ -19,8 +19,6 @@ import java.io.File;
 import java.net.URI;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertNotNull;
-import static org.junit.jupiter.api.Assertions.fail;
 
 /**
  * Tests for {@link LSPIJUtils#toUri(File)} method.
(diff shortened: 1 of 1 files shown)


angelozerr:
Windows CI build is failling https://github.com/redhat-developer/lsp4ij/actions/runs/19733661567/job/56541191426?pr=1378#step:5:74 but it is perhaps normal (you have not finished your work).

angelozerr:
Do you need this PR for the next release?

tim-sh:
> The PR seems very nice. My only fear is that it doesnt cover some cases and introduce some bugs

Thanks for the feedback! I can understand that fear. 
Is there any way I could test the version of this PR from a nightly-like build? This way I could do a first smoke test for both Linux and Windows.

tim-sh:
I will also add a test comparing the old and the new result and asserting that it only changes for WSL or other paths starting with a host name. What do you think, would that provide sufficient safety?

angelozerr:
@tim-sh please note that test fails https://github.com/redhat-developer/lsp4ij/actions/runs/19746312974/job/56630539800?pr=1378#step:5:79

angelozerr:
@tim-sh your PR looks very good but there is no test with wsl, is it expected? It is because of `The core problem seems to be Java disallowing $ in URL authorities`., right?

angelozerr:
Ok let's merge the PR. Thanks so much @tim-sh !

tim-sh:
@angelozerr You’re welcome and thanks for merging! ❤️

PR #1402, opened 2025-12-24
fix: "goto > Declaration or Usages" to show references when ⌘B on definitions

aridclown:
IntelliJ's "Go to > Declaration or Usages" (⌘B) shortcut is a commonly used shortcut that, in general, should display usages when triggered from the declaration itself.

In the LSP handler, we only call `textDocument/definition`, so pressing `⌘B` on a declaration has no effect (tested with LuaLS).

A few things to note:
- As discussed in #531, this change tentatively aims to instruct `LSPGotoDeclarationHandler` to fall back to `textDocument/references`: if the definition we receive overlaps the caret, we reuse the same reference list and popup that the "Go to → LSP References" command uses. 
- Ctrl-hover still just underlines the symbol, but a click now opens the rich usages panel instead of silently failing.
- To avoid duplicating code, we introduced a tiny `LSPReferenceCollector` helper and pointed both the standalone reference action and the new fallback at it.

> PS: I'm not an expert in LSP4IJ, so this is a tentative implementation. I tested it, and it works, but my main goal here is to restart the conversation on this issue and work out a solution, as this is a much-desired feature.

### DIFF
--- a/src/main/java/com/redhat/devtools/lsp4ij/features/navigation/LSPGotoDeclarationHandler.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/features/navigation/LSPGotoDeclarationHandler.java
@@ -64,21 +64,13 @@ public class LSPGotoDeclarationHandler implements GotoDeclarationHandler {
     public PsiElement[] getGotoDeclarationTargets(@Nullable PsiElement sourceElement,
                                                   int offset,
                                                   Editor editor) {
-        Project project = editor.getProject();
-        if (project == null || project.isDisposed()) {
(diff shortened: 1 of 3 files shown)


angelozerr:
Thanks so much for your contribution.

Let me some times to test and study your PR.

Please fix your commit message by adding first fix:

angelozerr:
@aridclown I did a  (different) fix in https://github.com/redhat-developer/lsp4ij/pull/1497/changes

angelozerr:
Replaced with https://github.com/redhat-developer/lsp4ij/pull/1497

PR #1403, opened 2025-12-30
docs: add rovo to projects that use lsp4ij

Arthurdw:
Closes https://github.com/Arthurdw/rovo/issues/13

Rovo uses lsp4ij to have the extension communicate with the Rovo LSP.


angelozerr:
Thanks @Arthurdw !

PR #1471, opened 2026-04-11
fix: add no-op for workspace/inlineValue/refresh to prevent UnsupportedOperationException

### LINKED ISSUE #1470: workspace/inlineValue/refresh throws UnsupportedOperationException
## Description

When a language server sends a `workspace/inlineValue/refresh` request, the LSP4IJ client throws an `UnsupportedOperationException` instead of handling it.

## Stack trace

```
Caused by: java.lang.UnsupportedOperationException
    at org.eclipse.lsp4j.services.LanguageClient.refreshInlineValues(LanguageClient.java:238)
```

## Expected behavior

Per the [LSP 3.17 spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_inlineValue_refresh), the server sends `workspace/inlineValue/refresh` to ask the client to re-request inline values for all open documents. The client should either:

1. Properly implement the refresh — trigger re-evaluation of inline values for all open editors, or
2. Return a no-op resolved future so the server doesn't crash (same approach as other unimplemented refresh notifications)

## Suggested fix

The pattern is already established in LSP4IJ for similar refresh requests. For example, `workspace/codeLens/refresh` and `workspace/semanticTokens/refresh` are handled in the `LanguageClientImpl` (or equivalent proxy class). Adding a handler for `workspace/inlineValue/refresh` should follow the same pattern.

A minimal no-op fix would be:

```java
@Override
public CompletableFuture<Void> refreshInlineValues() {
    // TODO: trigger inline value re-evaluation for open editors
    return CompletableFuture.completedFuture(null);
}
```

A full implementation would iterate over open editors, find those associated with the language server, and re-request inline values — mirroring the `workspace/semanticTokens/refresh` implementation.

## Steps to reproduce

1. Use any LSP server that sends `workspace/inlineValue/refresh` (e.g. php-lsp)
2. Open a file handled by that server in IntelliJ/PHPStorm with LSP4IJ
3. Observe the `UnsupportedOperationException` in the log

jorgsowa:
Fixes https://github.com/redhat-developer/lsp4ij/issues/1470

### DIFF
--- a/src/main/java/com/redhat/devtools/lsp4ij/client/LanguageClientImpl.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/client/LanguageClientImpl.java
@@ -209,6 +209,11 @@ public class LanguageClientImpl implements LanguageClient, Disposable {
     }
 
     @Override
+    public CompletableFuture<Void> refreshInlineValues() {
+        return CompletableFuture.completedFuture(null);
+    }
+
+    @Override
     public CompletableFuture<Void> refreshDiagnostics() {
         return CompletableFuture.runAsync(this::refreshDiagnosticsForAllOpenedFiles);
     }


angelozerr:
Thanks @jorgsowa !

PR #1485, opened 2026-05-09
fix: report definition request errors correctly

bodiam:
## What

Use the `textDocument/definition` request constant when handling definition request failures.

## Why

LSPDefinitionSupport calls TextDocumentService.definition(...), but passed the `textDocument/declaration` label to CancellationSupport. As a result, definition failures were logged or reported as declaration failures.

## Verification

- ./gradlew compileJava


angelozerr:
Good catch! Thanks for your contribution.

PR #1487, opened 2026-05-10
docs(README): add Xonsh plugin

nahoj:
Thank you for making this plugin!

Absent any obvious order, I'll be bold and insert Xonsh plugin near the top :upside_down_face:


nahoj:
Okay, there is an order.

angelozerr:
Thanks @nahoj !

PR #1489, opened 2026-05-10
feat: support dynamic plugin unloading via disposable listeners and cleanup logic

CommanderTvis:
Allows LSP4IJ and plugins contributing LSP servers to load/unload at runtime without an IDE restart.

Notes:

  - Removing a `semanticTokensColorsProvider` resets the definition to the default provider so the unloading plugin's class is dropped.
  - `LanguageServerWrapper` executors now produce daemon threads with the context classloader pinned to the platform parent loader - required so worker threads don't pin the plugin classloader.
  - `dispose()` waits up to 2s per executor and warns on timeout; the warning is the signal of a classloader leak on unload.
  - `ServerInstallerManager`: factory instantiation failures are now logged instead of silently swallowed (separate from the unload work, but in the same touched code).

### DIFF
--- a/src/main/java/com/redhat/devtools/lsp4ij/LanguageServerWrapper.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/LanguageServerWrapper.java
@@ -159,16 +159,33 @@ public class LanguageServerWrapper implements Disposable {
         VirtualFileManager.getInstance().addAsyncFileListener(fileListener, this);
 
         String projectName = sanitize(!serverDefinition.isSingleton() ? ("@" + project.getName()) : "");  //$NON-NLS-1$//$NON-NLS-2$
+
+        // Worker threads must not inherit the plugin classloader, or they pin it and block dynamic unload.
+        var workerCtxClassLoader = ApplicationManager.class.getClassLoader();
         String dispatcherThreadNameFormat = "LS-" + serverDefinition.getId() + projectName + "#dispatcher"; //$NON-NLS-1$ //$NON-NLS-2$
-        this.dispatcher = Executors
-                .newSingleThreadExecutor(new ThreadFactoryBuilder().setNameFormat(dispatcherThreadNameFormat).build());
+        this.dispatcher = Executors.newSingleThreadExecutor(new ThreadFactoryBuilder()
+                .setNameFormat(dispatcherThreadNameFormat)
+                .setDaemon(true)
+                .setThreadFactory(r -> {
+                    var t = new Thread(r);
+                    t.setContextClassLoader(workerCtxClassLoader);
+                    return t;
+                })
+                .build());
 
         // Executor service passed through to the LSP4j layer when we attempt to start the LS. It will be used
         // to create a listener that sits on the input stream and processes inbound messages (responses, or server-initiated
         // requests).
         String listenerThreadNameFormat = "LS-" + serverDefinition.getId() + projectName + "#listener-%d"; //$NON-NLS-1$ //$NON-NLS-2$
-        this.listener = Executors
-                .newCachedThreadPool(new ThreadFactoryBuilder().setNameFormat(listenerThreadNameFormat).build());
+        this.listener = Executors.newCachedThreadPool(new ThreadFactoryBuilder()
+                .setNameFormat(listenerThreadNameFormat)
+                .setDaemon(true)
+                .setThreadFactory(r -> {
+                    var t = new Thread(r);
+                    t.setContextClassLoader(workerCtxClassLoader);
+                    return t;
+                })
+                .build());
         updateStatus(ServerStatus.none);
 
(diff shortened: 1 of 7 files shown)


angelozerr:
Your PR looks promising. Have you an IJ plugin which is based on LSP4IJ?

src/main/java/com/redhat/devtools/lsp4ij/LanguageServerWrapper.java
```diff
@@ -159,16 +159,32 @@ public LanguageServerWrapper(@NotNull Project project,
...
 
         String projectName = sanitize(!serverDefinition.isSingleton() ? ("@" + project.getName()) : "");  //$NON-NLS-1$//$NON-NLS-2$
+
+        var workerCtxClassLoader = LanguageServerWrapper.class.getClassLoader().getParent();
         String dispatcherThreadNameFormat = "LS-" + serverDefinition.getId() + projectName + "#dispatcher"; //$NON-NLS-1$ //$NON-NLS-2$
-        this.dispatcher = Executors
-                .newSingleThreadExecutor(new ThreadFactoryBuilder().setNameFormat(dispatcherThreadNameFormat).build());
+        this.dispatcher = Executors.newSingleThreadExecutor(new ThreadFactoryBuilder()
+                .setNameFormat(dispatcherThreadNameFormat)
+                .setDaemon(true)
```
angelozerr: I don't undetstand very well why you need daemon Thread?

src/main/java/com/redhat/devtools/lsp4ij/LanguageServerWrapper.java
```diff
@@ -567,7 +597,7 @@ public void updateStatus(@NotNull ServerStatus serverStatus) {
     }
 
     private void startStopTimer() {
-        timer = new Timer("Stop Language Server Timer"); //$NON-NLS-1$
+        timer = new Timer("Stop Language Server Timer", true); //$NON-NLS-1$
```
angelozerr: why you need daemon timer? This old code comes from LSP4E, I think it should be better to use IJ Alarm.

CommanderTvis:
> Your PR looks promising. Have you an IJ plugin which is based on LSP4IJ?

Yes, https://bitbucket.org/chromawallet/rell-jetbrains/

angelozerr:
> > Your PR looks promising. Have you an IJ plugin which is based on LSP4IJ?
> 
> Yes, https://bitbucket.org/chromawallet/rell-jetbrains/

Nice!

Please add your IJ plugin in https://github.com/redhat-developer/lsp4ij#who-is-using-lsp4ij and if you like LSP4IJ please add a review in JetBrains Marketplace to help us to promote LSP4IJ. Thanks.

src/main/java/com/redhat/devtools/lsp4ij/LanguageServerWrapper.java
CommanderTvis: `setDaemon` is a precaution: if the listener thread is ever stuck and the `awaitTermination(2s)` below times out, a non-daemon thread would probably keep the JVM alive at shutdown.

src/main/java/com/redhat/devtools/lsp4ij/LanguageServerWrapper.java
CommanderTvis: Same precaution as above. But you're right, `Alarm` is the idiomatic thing.

src/main/java/com/redhat/devtools/lsp4ij/LanguageServerWrapper.java
angelozerr: Ok have you checked that when IJ is closed, there are none dameon thread which are alive.

src/main/java/com/redhat/devtools/lsp4ij/LanguageServerWrapper.java
```diff
@@ -121,7 +121,7 @@ public class LanguageServerWrapper implements Disposable {
     private LanguageServer languageServer;
     private LanguageClientImpl languageClient;
     private ServerCapabilities serverCapabilities;
-    private Timer timer;
+    private final Alarm stopAlarm = new Alarm(Alarm.ThreadToUse.POOLED_THREAD, this);
```
angelozerr: This alarm is required when server is stopped, so it is better to create it only if needed. See sample at UnresolvedCodeLensViewportContext

src/main/java/com/redhat/devtools/lsp4ij/LanguageServerWrapper.java
CommanderTvis: Checked. Indeed that was not necessary, because IDEA kills everything with `System.exit`

angelozerr:
Thanks @CommanderTvis ! Please install nighty build.

PR #1520, opened 2026-05-19
Let `LSPFindUsagesHandler` handle only its own elements

### LINKED ISSUE #1512: Infinite Recursion in `LSPFindUsagesHandlerFactory`
## Summary

[LSPFindUsagesHandlerFactory.java](file:///usr/local/google/home/skill/ij/lsp4ij/src/main/java/com/redhat/devtools/lsp4ij/usages/LSPFindUsagesHandlerFactory.java) uses a decoration pattern that defers to other `FindUsagesHandlerFactory` instances. It has a latent infinite recursion bug: if two or more factories in the extension list use the same "check all other factories" pattern, they call each other's `canFindUsages()` in a cycle, causing a `StackOverflowError`.

---

### How It Works

```mermaid
sequenceDiagram
    participant IJ as IntelliJ Platform
    participant LSP as LSPFindUsagesHandlerFactory
    participant Others as Other FindUsagesHandlerFactory(s)

    IJ->>LSP: canFindUsages(element)
    LSP->>LSP: canExecuteLSPFeature() == NOW?
    LSP->>Others: canExternalFindUsages(element)
    loop For each factory != this
        LSP->>Others: factory.canFindUsages(element)
        Others-->>LSP: true/false
    end
    LSP-->>IJ: !canExternalFindUsages (negated)
```

### The Decorator Pattern

The intent is:

> "Handle this element **only if** no other `FindUsagesHandlerFactory` can handle it — I'm the fallback."

This is implemented in [canFindUsages()](file:///usr/local/google/home/skill/ij/lsp4ij/src/main/java/com/redhat/devtools/lsp4ij/usages/LSPFindUsagesHandlerFactory.java#L28-L38):

```java
public boolean canFindUsages(@NotNull PsiElement element) {
    if (ProjectIndexingManager.canExecuteLSPFeature(...) == ExecuteLSPFeatureStatus.NOW) {
        return !canExternalFindUsages(element);  // "I handle it only if nobody else does"
    }
    return false;
}
```

And the external check in [canExternalFindUsages()](file:///usr/local/google/home/skill/ij/lsp4ij/src/main/java/com/redhat/devtools/lsp4ij/usages/LSPFindUsagesHandlerFactory.java#L45-L60):

```java
private boolean canExternalFindUsages(@NotNull PsiElement element) {
    for (FindUsagesHandlerFactory factory : EP_NAME.getExtensions(element.getProject())) {
        if (!factory.equals(this) && factory.canFindUsages(element)) {
            return true;
        }
    }
    return false;
}
```

---

### Root Cause

The `canExternalFindUsages()` loop iterates **all** registered `FindUsagesHandlerFactory` extensions and calls `canFindUsages()` on each. The guard is `!factory.equals(this)`, which only prevents self-recursion for the **same instance**.

If **two different factory instances** both use this pattern:

```
FactoryA.canFindUsages() →
  FactoryA.canExternalFindUsages() →
    FactoryB.canFindUsages()  [FactoryB != FactoryA, so it passes the guard] →
      FactoryB.canExternalFindUsages() →
        FactoryA.canFindUsages()  [FactoryA != FactoryB, so it passes the guard] →
          FactoryA.canExternalFindUsages() →
            FactoryB.canFindUsages() →
              ... StackOverflowError
```

### When This Happens

| Scenario | Risk |
|----------|------|
| **Two LSP plugins** both bundle LSP4IJ (or a fork) | High — each bundles its own `LSPFindUsagesHandlerFactory` instance |
| **Copy-paste derivative** — another plugin copies this pattern | High — common in IntelliJ plugin development |
| **Subclass in a downstream plugin** that extends `LSPFindUsagesHandlerFactory` and doesn't override `canFindUsages()` | Medium — `equals()` returns false for different classes |
| **Multiple classloaders** loading the same class | Low — IntelliJ's plugin classloader isolation can cause `equals()` to fail |

### Why It Hasn't Blown Up Yet (Mostly)

Currently there's only **one** `LSPFindUsagesHandlerFactory` registered as `order="last, before default"`. As long as no other factory uses the same delegation pattern, the recursion terminates because all other factories' `canFindUsages()` implementations are non-recursive (e.g., `JavaFindUsagesHandlerFactory` just checks `instanceof PsiClass`).

---

## Reproduction Scenario

1. Create a second plugin with a `FindUsagesHandlerFactory` that uses the same "check all others" pattern:

```java
public class MyFindUsagesHandlerFactory extends FindUsagesHandlerFactory {
    @Override
    public boolean canFindUsages(@NotNull PsiElement element) {
        // Same pattern: "I'm the fallback"
        for (var factory : EP_NAME.getExtensions(element.getProject())) {
            if (!factory.equals(this) && factory.canFindUsages(element)) {
                return false; // someone else handles it
            }
        }
        return true;
    }
    ...
}
```

2. Register it in `plugin.xml` with `order="last"`.
3. Trigger "Find Usages" on any `PsiElement` → `StackOverflowError`.

---

### Option A: ThreadLocal Reentrancy Guard (Minimal Change)

Add a `ThreadLocal<Boolean>` flag to prevent reentrant calls:

```diff
 public class LSPFindUsagesHandlerFactory extends FindUsagesHandlerFactory {
+    private static final ThreadLocal<Boolean> IN_EXTERNAL_CHECK = ThreadLocal.withInitial(() -> false);
+
     @Override
     public boolean canFindUsages(@NotNull PsiElement element) {
+        if (IN_EXTERNAL_CHECK.get()) {
+            // Another factory is already checking us — answer without recursing.
+            // Return true here means "yes I can handle it" which tells the
+            // caller they are NOT the sole handler.
+            return ProjectIndexingManager.canExecuteLSPFeature(
+                element.getContainingFile()) == ExecuteLSPFeatureStatus.NOW;
+        }
         if (ProjectIndexingManager.canExecuteLSPFeature(element.getContainingFile()) == ExecuteLSPFeatureStatus.NOW) {
             return !canExternalFindUsages(element);
         }
         return false;
     }

     private boolean canExternalFindUsages(@NotNull PsiElement element) {
+        IN_EXTERNAL_CHECK.set(true);
+        try {
             for (FindUsagesHandlerFactory factory : EP_NAME.getExtensions(element.getProject())) {
                 try {
                     if (!factory.equals(this) && factory.canFindUsages(element)) {
                         return true;
                     }
                 }
                 catch (IndexNotReadyException | ProcessCanceledException e) {
                     throw e;
                 }
                 catch (Exception e) {
                 }
             }
             return false;
+        } finally {
+            IN_EXTERNAL_CHECK.set(false);
+        }
     }
 }
```

> [!TIP]
> When `IN_EXTERNAL_CHECK` is true and another factory with the same pattern calls us, we return a simple capability check (can we handle LSP for this file?) without recursing into `canExternalFindUsages()` again. This breaks the cycle.

### Option B: `instanceof` Guard (Broader)

Skip any factory that is also an `LSPFindUsagesHandlerFactory`:

```diff
     private boolean canExternalFindUsages(@NotNull PsiElement element) {
         for (FindUsagesHandlerFactory factory : EP_NAME.getExtensions(element.getProject())) {
             try {
-                if (!factory.equals(this) && factory.canFindUsages(element)) {
+                if (factory != this
+                    && !(factory instanceof LSPFindUsagesHandlerFactory)
+                    && factory.canFindUsages(element)) {
                     return true;
                 }
```

> [!WARNING]
> This only protects against other `LSPFindUsagesHandlerFactory` subclasses. It doesn't protect against unrelated factories using the same delegation pattern.

### Option C: Class-level Static Guard (Most Robust)

Use a static `Set<Class<?>>` to track which factory classes are currently in a delegation check:

```diff
+    private static final ThreadLocal<Set<Class<?>>> CHECKING_CLASSES =
+        ThreadLocal.withInitial(HashSet::new);
+
     private boolean canExternalFindUsages(@NotNull PsiElement element) {
+        Set<Class<?>> checking = CHECKING_CLASSES.get();
+        checking.add(getClass());
+        try {
             for (FindUsagesHandlerFactory factory : EP_NAME.getExtensions(element.getProject())) {
                 try {
-                    if (!factory.equals(this) && factory.canFindUsages(element)) {
+                    if (factory != this
+                        && !checking.contains(factory.getClass())
+                        && factory.canFindUsages(element)) {
                         return true;
                     }
                 ...
             }
             return false;
+        } finally {
+            checking.remove(getClass());
+        }
     }
```

> [!TIP]
> This is the most robust option. It prevents recursion between any combination of factories that use this delegation pattern, regardless of their class hierarchy.

## Recommendation

**Option C** (class-level static guard) is recommended. It's the most general solution and prevents the bug even when unrelated factory implementations copy the pattern.

SergejSalnikov:
fix: #1512

### DIFF
--- a/src/main/java/com/redhat/devtools/lsp4ij/usages/LSPFindUsagesHandlerFactory.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/usages/LSPFindUsagesHandlerFactory.java
@@ -12,8 +12,6 @@ package com.redhat.devtools.lsp4ij.usages;
 
 import com.intellij.find.findUsages.FindUsagesHandler;
 import com.intellij.find.findUsages.FindUsagesHandlerFactory;
-import com.intellij.openapi.progress.ProcessCanceledException;
-import com.intellij.openapi.project.IndexNotReadyException;
 import com.intellij.psi.PsiElement;
 import com.redhat.devtools.lsp4ij.client.ExecuteLSPFeatureStatus;
 import com.redhat.devtools.lsp4ij.client.indexing.ProjectIndexingManager;
@@ -26,36 +24,15 @@ import org.jetbrains.annotations.Nullable;
 public class LSPFindUsagesHandlerFactory extends FindUsagesHandlerFactory {
     @Override
     public boolean canFindUsages(@NotNull PsiElement element) {
-        // Execute the dummy LSP find usage handler to collect references, implementations
-        // with LSPUsageSearcher if file is associated to a language server.
-        if (ProjectIndexingManager.canExecuteLSPFeature(element.getContainingFile()) == ExecuteLSPFeatureStatus.NOW) {
-            // To avoid overriding existing FindUsagesHandlerFactory for the given PsiElement
-            // (ex: JavaFindUsagesHandlerFactory which is also defined as "last".)
-            // we check if it exists a FindUsagesHandlerFactory for the element.
-            return !canExternalFindUsages(element);
-        }
-        return false;
+        return element instanceof LSPUsageTriggeredPsiElement &&
+            // Execute the dummy LSP find usage handler to collect references, implementations
+            // with LSPUsageSearcher if file is associated to a language server.
+            ProjectIndexingManager.canExecuteLSPFeature(element.getContainingFile()) == ExecuteLSPFeatureStatus.NOW;
+
     }
 
     @Override
     public @Nullable FindUsagesHandler createFindUsagesHandler(@NotNull PsiElement element, boolean forHighlightUsages) {
         return new LSPFindUsagesHandler(element);
-    }
-
-    private boolean canExternalFindUsages(@NotNull PsiElement element) {
-        for (FindUsagesHandlerFactory factory : FindUsagesHandlerFactory.EP_NAME.getExtensions(element.getProject())) {
-            try {
-                if (!factory.equals(this) &&  factory.canFindUsages(element)) {
-                    return true;
-                }
-            }
-            catch (IndexNotReadyException | ProcessCanceledException e) {
-                throw e;
-            }
-            catch (Exception e) {
-
-            }
-        }
-        return false;
     }
 }


angelozerr:
Replaced with https://github.com/redhat-developer/lsp4ij/pull/1523

PR #1534, opened 2026-05-21
Reuse cached documentLinks when available.

### LINKED ISSUE #1533: `textDocument/documentLink` navigation fails for files outside project directories (e.g., external libraries)
### **Status / Impact**
- **Severity:** Medium (Breaks document link navigation for all files located outside the project roots, such as library sources, SDKs, or external dependencies).
- **Affects:** All IntelliJ Platform versions.
- **Scope:** Any LSP-backed file that is located outside the project's module content roots, even though the server supports it and the links are correctly decorated (underlined).

---

### **1. Summary of the Problem**
LSP document links (`textDocument/documentLink`) are correctly requested, resolved, and underlined by `LSPDocumentLinkAnnotator` in all files, including those located **outside** the project directory (such as external library sources or JDK files).

However, **Ctrl+Click (or Cmd+Click on macOS) has no effect** in these external files. Clicking a highlighted link does nothing, and no browser or editor is opened.

---

### **2. Root Cause Analysis**
When the user Ctrl-clicks a decorated link, `LSPDocumentLinkGotoDeclarationHandler.getGotoDeclarationTargets` is invoked to resolve the navigation target.

The handler attempts to resolve the active `Project` instance using the following logic:
```java
        Document document = editor.getDocument();
        VirtualFile file = LSPIJUtils.getFile(document);
        Module module = LSPIJUtils.getModule(file, sourceElement.getProject());
        Project project = module != null ? module.getProject() : null;
        if (project == null || project.isDisposed()) {
            return PsiElement.EMPTY_ARRAY; // <--- ABORTS HERE
        }
```

#### **Why it fails for external files:**
1. If the file is located outside the project's directories (e.g., a read-only library file, JDK source, or system file), it is not associated with any of the project's modules.
2. Therefore, `LSPIJUtils.getModule()` returns `null`.
3. Because `module` is `null`, the `project` variable is resolved as `null`.
4. The handler hits the early return condition and immediately returns `PsiElement.EMPTY_ARRAY` (aborting the resolution process).

This happens even though:
- `sourceElement.getProject()` is already a valid, non-null `Project` instance (representing the currently active project where the editor is open).
- The language server successfully processed the file and returned the links (which is why the annotator successfully underlined them).

SergejSalnikov:
Fix: #1533

### DIFF
--- a/src/main/java/com/redhat/devtools/lsp4ij/features/documentLink/LSPDocumentLinkGotoDeclarationHandler.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/features/documentLink/LSPDocumentLinkGotoDeclarationHandler.java
@@ -59,17 +59,24 @@ public class LSPDocumentLinkGotoDeclarationHandler implements GotoDeclarationHan
         if (!LanguageServersRegistry.getInstance().isFileSupported(psiFile)) {
             return PsiElement.EMPTY_ARRAY;
         }
+        Project project = sourceElement.getProject();
         Document document = editor.getDocument();
-        VirtualFile file = LSPIJUtils.getFile(document);
-        Module module = LSPIJUtils.getModule(file, sourceElement.getProject());
-        Project project = module != null ? module.getProject() : null;
-        if (project == null || project.isDisposed()) {
-            return PsiElement.EMPTY_ARRAY;
-        }
 
         LSPDocumentLinkSupport documentLinkSupport = LSPFileSupport.getSupport(psiFile).getDocumentLinkSupport();
-        var params = new DocumentLinkParams(new TextDocumentIdentifier());
-        CompletableFuture<List<DocumentLinkData>> documentLinkFuture = documentLinkSupport.getDocumentLinks(params);
+        CompletableFuture<List<DocumentLinkData>> documentLinkFuture = documentLinkSupport.getValidLSPFuture();
+
+        if (documentLinkFuture == null) {
+            VirtualFile file = LSPIJUtils.getFile(document);
+            Module module = LSPIJUtils.getModule(file, sourceElement.getProject());
+            // Query LS only for files inside the project.
+            if (module == null || module.isDisposed()) {
+                return PsiElement.EMPTY_ARRAY;
+            }
+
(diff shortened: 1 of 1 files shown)


src/main/java/com/redhat/devtools/lsp4ij/features/documentLink/LSPDocumentLinkGotoDeclarationHandler.java
```diff
@@ -59,17 +59,24 @@ public class LSPDocumentLinkGotoDeclarationHandler implements GotoDeclarationHan
...
-        Module module = LSPIJUtils.getModule(file, sourceElement.getProject());
-        Project project = module != null ? module.getProject() : null;
-        if (project == null || project.isDisposed()) {
-            return PsiElement.EMPTY_ARRAY;
-        }
 
         LSPDocumentLinkSupport documentLinkSupport = LSPFileSupport.getSupport(psiFile).getDocumentLinkSupport();
-        var params = new DocumentLinkParams(new TextDocumentIdentifier());
-        CompletableFuture<List<DocumentLinkData>> documentLinkFuture = documentLinkSupport.getDocumentLinks(params);
+        CompletableFuture<List<DocumentLinkData>> documentLinkFuture = documentLinkSupport.getValidLSPFuture();
```
angelozerr, line 66: When you consume `documentLinkSupport.getDocumentLinks(params);` it will return documentLinkSupport.getValidLSPFuture() if Psi File has not changed, so today it uses the cache.

angelozerr:
@SergejSalnikov to check what I mean I suggest that you set in Debug tab the trace as verbose to see lsp traces in the lsp console.

When you will do ctrl+click on hyperlink, you should not seen textDocument/documebtLink request.

angelozerr:
I close this PR since cache is already used.

SergejSalnikov:
This PR is not about caching, but tries to address #1533. The intent was not to try to send any extra requests, but if the document outside of the project was already annotated with links, to open those links.
Because currently the documents outside of the project have unclickable links.

angelozerr:
@SergejSalnikov indeed your PR fixes file outside project but please keep the existing cache.

Issue #1553, opened 2026-05-25
Server status shows "stopping" during `lastDocumentDisconnectedTimeout` grace period


SergejSalnikov:
When `lastDocumentDisconnectedTimeout` is set to a non-trivial value (e.g., 180 seconds), the LSP console UI displays **"stopping pid:12345"** with an animated spinner and an elapsed-time counter for the entire duration of the grace period. The server is still fully functional during this time — it accepts new connections and processes requests — but the UI incorrectly suggests it is in the process of shutting down.

## Steps to Reproduce

1. Register a language server with a `lastDocumentDisconnectedTimeout` > 5 seconds, e.g.:
   ```xml
   <server id="myserver" name="MyServer"
           lastDocumentDisconnectedTimeout="180"
           factoryClass="..."/>
   ```
2. Open a file that triggers the language server to start.
3. Close the file (or close all files associated with the server).
4. Observe the **Language Servers** tool window (LSP console tree).

**Expected:** The server shows as **"started pid:12345"** (or a new "idle" state) during the grace period, since it's still alive and accepting connections.

**Actual:** The server immediately shows **"stopping pid:12345"** with a spinning animation and an elapsed-time counter that ticks for the full 180 seconds.

## Root Cause

In [`LanguageServerWrapper.startStopTimer()`](https://github.com/redhat-developer/lsp4ij/blob/main/src/main/java/com/redhat/devtools/lsp4ij/LanguageServerWrapper.java) (line ~606):

```java
private void startStopTimer() {
    updateStatus(ServerStatus.stopping);  // ← BUG: status set immediately
    int delayMs = (int) TimeUnit.SECONDS.toMillis(
        serverDefinition.getLastDocumentDisconnectedTimeout());
    getStopAlarm().addRequest(() -> {
        try {
            stop();
        } catch (Throwable t) {
            LOGGER.error("...", t);
        }
    }, delayMs);
}
```

`updateStatus(ServerStatus.stopping)` is called **immediately** when the timer starts, not when the server actually begins its shutdown sequence. This triggers:

1. **`LanguageServerProcessTreeNode.setServerStatus(stopping)`** — sets `displayName` to `"stopping pid:..."` and starts the elapsed-time counter (`startTime = System.currentTimeMillis()`).
2. **`LanguageServerTreeRenderer`** — detects `ServerStatus.stopping` and renders the animated spinner icon + elapsed time text.

[...]

PR #1554, opened 2026-05-25
fix: defer ServerStatus.stopping until grace period expires

SergejSalnikov:
startStopTimer() was setting ServerStatus.stopping immediately when the lastDocumentDisconnectedTimeout timer started, not when the server actually began shutting down. This caused the LSP console to display "stopping pid:..." with a spinner for the entire grace period, even though the server was still fully functional and accepting connections. With the default 5-second timeout this was barely noticeable, but with longer timeouts (e.g. 180s for expensive-to-start servers) the UI was misleading — users could think the server was broken or unresponsive. Move updateStatus(ServerStatus.stopping) inside the alarm callback, just before stop(). The server now stays in 'started' status during the grace period and only transitions to 'stopping' when shutdown actually begins.

FIX=#1553

### DIFF
--- a/src/main/java/com/redhat/devtools/lsp4ij/LanguageServerWrapper.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/LanguageServerWrapper.java
@@ -633,9 +633,9 @@ public class LanguageServerWrapper implements Disposable {
     }
 
     private void startStopTimer() {
-        updateStatus(ServerStatus.stopping);
         int delayMs = (int) TimeUnit.SECONDS.toMillis(serverDefinition.getLastDocumentDisconnectedTimeout());
         getStopAlarm().addRequest(() -> {
+            updateStatus(ServerStatus.stopping);
             try {
                 stop();
             } catch (Throwable t) {


angelozerr:
If you do that you will never seen that server is stopping, right?

SergejSalnikov:
I did not look into "stopping" logic. I though that it send terminate command and waits until server shuts down (which I assumed could take time). 

Maybe the state should be renamed from "stopping" into "idling"?

SergejSalnikov:
"stopping" implies that there is some busy work is in progress.

angelozerr:
@SergejSalnikov IMHO I think stopping is the right name. The server is stopping....

I wonder if in your case you want never stop the server when all files are closed. If you want to do that you will need to create [custom LSPClientFeatures](https://github.com/redhat-developer/lsp4ij/blob/main/docs/LSPApi.md#lsp-client-features) and `override canStopServerByUser to return false`;

SergejSalnikov:
I want to have 3 min idling time before the server shuts itself down.

angelozerr:
Why not never stop the server?

SergejSalnikov:
Because remote language server holds many dependent services such as clangd, golangd, typescript ls and others.
stopping and starting those is an expensive operation.

angelozerr:
And why you don't do my suggestion?

>I wonder if in your case you want never stop the server when all files are closed. If you want to do that you will need to create [custom LSPClientFeatures](https://github.com/redhat-developer/lsp4ij/blob/main/docs/LSPApi.md#lsp-client-
> features) and override canStopServerByUser to return false;

SergejSalnikov:
because if user doesn't really use those, we don't need to waste memory and resources on remote host.
3 min without relevant docs open is a good compromise for us.

angelozerr:
I would like to keep stopping status, but if you don't want see this state, we could add in the LSPClientFeatures a boolean getter which does the behaviour that you wish to do

angelozerr:
Or we could provide a customization of those Label getter of the server status to display.

PR #1568, opened 2026-05-28
fix: diagnostics not displaying in Remote Development (Gateway)

### LINKED ISSUE #1567: Diagnostic prompts do not show up in Gateway/Remote development
When using Gateway Remote development Diagnostic messages in the LSP receive correctly but then do not display in the editor

Using GoLand 2024.2.3

Repro

1. install LSP that uses notifications
2. trigger a notification with a file opened on remote host
3. see no squiggly line opens

NicholasBlaskey:
Diagnostics not showing up in remote development, fixes https://github.com/redhat-developer/lsp4ij/issues/1567

Root cause seems to be `isFileOpen` always returns false. The doc comment isn't super clear to me but `isFileOpenWithRemotes` seems to indicate there might be a gap here.
```
  /**
   * @return {@code true} if {@code file} is opened in the host editor, {@code false} otherwise.
   */
  public abstract boolean isFileOpen(@NotNull VirtualFile file);

  /**
   * @return {@code true} if {@code file} is opened, {@code false} otherwise.
   * Unlike {@link #isFileOpen(VirtualFile)} includes files which were opened
   * by all guests during a collaborative development session.
   */
  @ApiStatus.Experimental
  public boolean isFileOpenWithRemotes(@NotNull VirtualFile file) {
      return isFileOpen(file);
  }
```

Add an additional check to see if we just called open on the file. 

Tested by installing and validating that the diagnostics do show correctly after this fix.

### DIFF
--- a/src/main/java/com/redhat/devtools/lsp4ij/LanguageServerWrapper.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/LanguageServerWrapper.java
@@ -1409,7 +1409,8 @@ public class LanguageServerWrapper implements Disposable {
         if (fileUri == null) {
             return;
         }
-        boolean isOpen = FileEditorManager.getInstance(getProject()).isFileOpen(file);
+        boolean isOpen = getOpenedDocument(fileUri, false) != null
+                || FileEditorManager.getInstance(getProject()).isFileOpen(file);
         final LSPDocumentBase openedOrClosedDocument = isOpen ? getOpenedDocument(fileUri, true) : getClosedDocument(fileUri, true);
         if (openedOrClosedDocument == null) {
             return;


src/main/java/com/redhat/devtools/lsp4ij/LanguageServerWrapper.java
```diff
@@ -1409,7 +1409,9 @@ public void updateDiagnostics(@NotNull VirtualFile file,
         if (fileUri == null) {
             return;
         }
-        boolean isOpen = FileEditorManager.getInstance(getProject()).isFileOpen(file);
+        // FileEditorManager.isFileOpen() always returns false for Gateway remote development.
+        boolean isOpen = getOpenedDocument(fileUri, false) != null
+                || FileEditorManager.getInstance(getProject()).isFileOpen(file);
```
angelozerr, line 1413: What about using this code?

```java
boolean isOpen = Arrays.stream(
    FileEditorManager.getInstance(getProject()).getOpenFiles()
).anyMatch(f -> f.equals(file));
```

angelozerr:
@NicholasBlaskey thanks for your fix, is it working for you?

src/main/java/com/redhat/devtools/lsp4ij/LanguageServerWrapper.java
NicholasBlaskey: Does not seem to work. Using `.getOpenFilesWithRemotes` probably will work but its experimental and not sure if it will have any compatibility issues

NicholasBlaskey:
> @NicholasBlaskey thanks for your fix, is it working for you?

no unfortunately, responded in the comment

PR #1590, opened 2026-06-09
fix: read goToLocations mode argument as JsonPrimitive (enable peek/gotoAndPeek)

### LINKED ISSUE #1588: editor.action.goToLocations ignores "peek"/"gotoAndPeek" mode — always navigates to the first location
## Summary

`GoToLocationsAction` (handler for `editor.action.goToLocations` / `editor.action.peekLocations`) reads the `mode` argument with `instanceof String`. But command arguments are Gson `JsonElement`s — the action itself reads the locations argument as `(JsonArray) command.getArgumentAt(2)`. A JSON string therefore arrives as a `JsonPrimitive`, the `instanceof String` check is `false`, `mode` stays at its `"goto"` default, and `handleGoto` runs — navigating to the **first** location. As a result `"peek"` and `"gotoAndPeek"` are unreachable.

**File:** `src/main/java/com/redhat/devtools/lsp4ij/commands/editor/GoToLocationsAction.java`

## Steps to reproduce

A language server returns a CodeLens/command:

```json
{
  "title": "injected: 9 candidates",
  "command": "editor.action.goToLocations",
  "arguments": [
    "file:///project/Consumer.os",
    { "line": 18, "character": 4 },
    [ { "uri": "file:///project/A.os", "range": { "start": {"line":0,"character":0}, "end": {"line":0,"character":1} } },
      { "uri": "file:///project/B.os", "range": { "start": {"line":0,"character":0}, "end": {"line":0,"character":1} } } ],
    "peek"
  ]
}
```

Click the lens.

- **Expected:** the locations popup opens (`handlePeek`), because `mode == "peek"`.
- **Actual:** the editor navigates to the first location (`A.os`); the popup never opens.

(`"gotoAndPeek"` is affected the same way.)

## Root cause

```java
// Get the fourth argument (mode: "goto" | "peek" | "gotoAndPeek")
String mode = MODE_GOTO; // default
Object modeArg = command.getArgumentAt(3);
if (modeArg instanceof String) {   // JsonPrimitive, not java.lang.String -> never matches
    mode = (String) modeArg;
}
```

`command.getArgumentAt(3)` returns a Gson `JsonElement` (a `JsonPrimitive` for the string), so `instanceof String` is always `false` and `mode` stays `"goto"`.

## Proposed fix

Read the mode from the `JsonPrimitive` (keeping a fallback for an already-decoded `String`):

```java
import com.google.gson.JsonPrimitive;
...
String mode = MODE_GOTO; // default
Object modeArg = command.getArgumentAt(3);
if (modeArg instanceof JsonPrimitive primitive && primitive.isString()) {
    mode = primitive.getAsString();
} else if (modeArg instanceof String stringMode) {
    mode = stringMode;
}
```

PR follows.

nixel2007:
Fixes #1588

## Problem

`GoToLocationsAction` reads the optional `mode` argument (`"goto"` | `"peek"` | `"gotoAndPeek"`) with `instanceof String`. Command arguments are Gson `JsonElement`s (the action reads the locations argument as `(JsonArray) command.getArgumentAt(2)`), so the mode string arrives as a `JsonPrimitive`. The check never matched, `mode` stayed at its `"goto"` default, and the action always navigated to the first location — `"peek"` and `"gotoAndPeek"` were unreachable.

## Fix

Read the mode from the `JsonPrimitive` via `getAsString()`, keeping a fallback for an already-decoded `String`:

```java
if (modeArg instanceof JsonPrimitive primitive && primitive.isString()) {
    mode = primitive.getAsString();
} else if (modeArg instanceof String stringMode) {
    mode = stringMode;
}
```

## Notes

- Single-file change; no behavior change for the `"goto"` default path.

### DIFF
--- a/src/main/java/com/redhat/devtools/lsp4ij/commands/editor/GoToLocationsAction.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/commands/editor/GoToLocationsAction.java
@@ -11,6 +11,7 @@
 package com.redhat.devtools.lsp4ij.commands.editor;
 
 import com.google.gson.JsonArray;
+import com.google.gson.JsonPrimitive;
 import com.intellij.openapi.actionSystem.ActionUpdateThread;
 import com.intellij.openapi.actionSystem.AnActionEvent;
 import com.intellij.openapi.actionSystem.DataContext;
@@ -89,11 +90,7 @@ public class GoToLocationsAction extends LSPCommandAction {
         }
 
         // Get the fourth argument (mode: "goto" | "peek" | "gotoAndPeek")
-        String mode = MODE_GOTO; // default
-        Object modeArg = command.getArgumentAt(3);
-        if (modeArg instanceof String) {
-            mode = (String) modeArg;
-        }
+        String mode = resolveMode(command.getArgumentAt(3));
 
         DataContext dataContext = e.getDataContext();
         LanguageServerItem languageServer = dataContext.getData(CommandExecutor.LSP_COMMAND_LANGUAGE_SERVER);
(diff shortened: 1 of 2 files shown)


nixel2007:
Update: I extracted the mode parsing into a package-private `GoToLocationsAction#resolveMode(Object)` and added `GoToLocationsActionTest` covering the `JsonPrimitive` cases (`peek`/`gotoAndPeek`/`goto`), an already-decoded `String`, `null`, and a non-string primitive.

I built and ran it locally against the IntelliJ platform (JDK 21):

```
GoToLocationsActionTest
  jsonPrimitivePeekIsResolved PASSED
  jsonPrimitiveGotoAndPeekIsResolved PASSED
  jsonPrimitiveGotoIsResolved PASSED
  alreadyDecodedStringIsResolved PASSED
  nullArgumentDefaultsToGoto PASSED
  nonStringPrimitiveDefaultsToGoto PASSED
SUCCESS: Executed 6 tests
```

The `jsonPrimitivePeekIsResolved` case is the regression guard — it fails with the previous `instanceof String` check.

angelozerr:
Good catch @nixel2007 !

angelozerr:
@nixel2007  please install https://github.com/redhat-developer/lsp4ij#testing-nightly-builds

nixel2007:
@angelozerr I've already verified the fix with locally built distribution, thanks

PR #1609, opened 2026-07-02
feat: add Candor Language Server template

tombaldwin:
Adds a template for the [candor](https://candor.poly.io) language server — a per-function **side-effect map** rendered in-editor from a precomputed report:

- **CodeLens** on each effectful function: `⚡ Db, Net · blast radius 12` (transitive effects + how many functions are affected by a change);
- **Diagnostics**: the repo's checked-in architecture policy (e.g. *the domain layer performs no I/O*) evaluated live, a violation squiggled at the offending function;
- **Hover**: effect provenance — the call chain through which an inherited effect reaches its source.

The server is engine-agnostic (one report format covers JVM bytecode, TypeScript/JavaScript, Rust and Swift) and ships as `candor-lsp` in the [candor-ts npm package](https://www.npmjs.com/package/candor-ts). The template + installer follow the `typescript-language-server` pattern (npm install into the lsp4ij working dir, then configure the `.bin` command); `template.json` maps the file types the report can cover.

Tested locally: both JSON files parse, and the server negotiates codeLens/hover/diagnostics over LSP stdio (its behavioral suite runs in the candor-ts repo).

### DIFF
--- a/src/main/resources/templates/lsp/candor-lsp/README.md
+++ b/src/main/resources/templates/lsp/candor-lsp/README.md
@@ -1,0 +1,25 @@
+You can use the [candor language server](https://candor.poly.io) to see each function's **side-effect
+map** where the code is:
+
+ * **CodeLens** on every effectful function: `⚡ Db, Net · blast radius 12` — the effects it (transitively)
+   performs and how many functions are affected if it changes;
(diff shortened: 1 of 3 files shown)


angelozerr:
LGTM, thank's @tombaldwin !

PR #1613, opened 2026-07-15
docs: Add ARO as downstream project

KrisSimon:
As mentioned by @angelozerr in https://github.com/arolang/aro/discussions/223


angelozerr:
Thanks @KrisSimon !

PR #1625, opened 2026-07-30
docs: remove Rell plugin from Who is using LSP4IJ

CommanderTvis:
The Rell IntelliJ Plugin no longer uses LSP4IJ. It switched to the internal IntelliJ LSP support, so it's removed from the "Who is using LSP4IJ?" list.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


angelozerr:
Thanks for the update.

PR #1634, opened 2026-08-15
fix: support IntelliJ IDEA 2026.2

### LINKED ISSUE #1633: LSP4IJ 0.20.1 does not work with IntelliJ IDEA 2026.2.1
> Preface: I'm filing this using my coding agent. I've manually reviewed this, and it _seems_ correct, but apologies if I've missed something.

## Description

LSP4IJ 0.20.1 does not work with IntelliJ IDEA 2026.2.1 (`IU-262.9437.185`).

IntelliJ Plugin Verifier reports two binary incompatibilities caused by Structure View classes that are no longer available in build 262:

- `com.intellij.ide.impl.StructureViewWrapperImpl`
- `com.intellij.ide.structureView.impl.common.PsiTreeElementBase`

These unresolved classes can cause `NoSuchClassError` at runtime. LSP4IJ 0.20.1 declares `since-build="242"` without an `until-build`, so it is currently treated as compatible with IntelliJ 2026.2.1.

BAML depends on LSP4IJ and cannot publish support for IntelliJ 2026.2 while this incompatibility exists. We are merging [BoundaryML/baml#4395](https://github.com/BoundaryML/baml/pull/4395) to cap BAML at `until-build="261.*"` until a compatible LSP4IJ release is available.

## How to reproduce

Download LSP4IJ 0.20.1 and IntelliJ Plugin Verifier 1.408:

```shell
curl -fL https://github.com/redhat-developer/lsp4ij/releases/download/0.20.1/lsp4ij-0.20.1.zip -o lsp4ij-0.20.1.zip
curl -fL https://github.com/JetBrains/intellij-plugin-verifier/releases/download/1.408/verifier-cli-1.408-all.jar -o verifier-cli-1.408-all.jar
```

Install or extract IntelliJ IDEA Ultimate 2026.2.1, then run:

```shell
java -jar verifier-cli-1.408-all.jar check-plugin lsp4ij-0.20.1.zip /absolute/path/to/ideaIU-2026.2.1
```

Plugin Verifier schedules:

```text
com.redhat.devtools.lsp4ij:0.20.1 against IU-262.9437.185
```

and reports:

```text
2 compatibility problems.

Access to unresolved class com.intellij.ide.impl.StructureViewWrapperImpl

Method com.redhat.devtools.lsp4ij.internal.editor.StructureViewEditorFeature.lambda$collectUiRunnable$0() references an unresolved class com.intellij.ide.impl.StructureViewWrapperImpl. This can lead to NoSuchClassError at runtime.

Package 'com.intellij.ide.structureView.impl.common' is not found along with its class com.intellij.ide.structureView.impl.common.PsiTreeElementBase.

PsiTreeElementBase is referenced by:
- LSPDocumentSymbolStructureViewModel.LSPDocumentSymbolViewElement
- LSPDocumentSymbolStructureViewModel.LSPFileStructureViewElement
```

As a control, running the same LSP4IJ archive and verifier against IntelliJ IDEA Ultimate 2026.1.5 (`IU-261.27258.48`) reports:

```text
Compatible.
```

## Relevant source

The current `main` branch still references both unavailable classes:

- [`PsiTreeElementBase`](https://github.com/redhat-developer/lsp4ij/blob/d1a91b422cf8f72a73ddbdda0ea3a6ae32caead5/src/main/java/com/redhat/devtools/lsp4ij/features/documentSymbol/LSPDocumentSymbolStructureViewModel.java#L19)
- [`StructureViewWrapperImpl`](https://github.com/redhat-developer/lsp4ij/blob/d1a91b422cf8f72a73ddbdda0ea3a6ae32caead5/src/main/java/com/redhat/devtools/lsp4ij/internal/editor/StructureViewEditorFeature.java#L13)
- [`StructureViewWrapperImpl.STRUCTURE_CHANGED` usage](https://github.com/redhat-developer/lsp4ij/blob/d1a91b422cf8f72a73ddbdda0ea3a6ae32caead5/src/main/java/com/redhat/devtools/lsp4ij/internal/editor/StructureViewEditorFeature.java#L68-L70)

This appears separate from #1503, which addressed the `InlayHintsPassFactory` incompatibility in a 2026.2 EAP.

## Expected behavior

LSP4IJ should install and run on IntelliJ IDEA 2026.2.1 without unresolved IntelliJ Platform classes.

Until build 262 support is available, the published LSP4IJ plugin should be restricted to `until-build="261.*"`.

sxlijin:
Fixes #1633.

## What changed

- Declare an optional dependency on the bundled `intellij.structureView.plugin` and add its required optional configuration descriptor.
- Remove the Plugin Verifier upper build bound so stable releases after `261.*`, including 2026.2, are covered.
- Omit the coarse `MISSING_DEPENDENCIES` failure level because Plugin Verifier also applies it to unavailable optional dependencies ([MP-2974](https://youtrack.jetbrains.com/issue/MP-2974)); invalid plugins and compatibility problems remain fatal.

## Root cause

`PsiTreeElementBase` and `StructureViewWrapperImpl` still exist in IntelliJ IDEA 2026.2.1. Their class files moved from the IDE core classpath in 2026.1 to `plugins/platform-structureView-plugin/lib/modules/intellij.platform.structureView.jar` in 2026.2. Because LSP4IJ did not declare the owning bundled plugin, those classes were outside LSP4IJ's dependency-visible classpath. The dependency is optional so the same plugin archive remains compatible with 2026.1, where that bundled plugin ID is absent and the classes are still supplied by the core classpath.

The 262.9437 source and product descriptors identify [`intellij.structureView.plugin`](https://github.com/JetBrains/intellij-community/blob/eaefd49405a1a53d7ec2f0c7eed81b2d8628e219/platform/structure-view-impl/plugin/resources/META-INF/plugin.xml#L4-L12) as the owning plugin and expose its [`intellij.platform.structureView` module publicly](https://github.com/JetBrains/intellij-community/blob/eaefd49405a1a53d7ec2f0c7eed81b2d8628e219/platform/structure-view-impl/resources/intellij.platform.structureView.xml#L1-L27).

## Before and after

I built upstream `main` at `d1a91b422cf8f72a73ddbdda0ea3a6ae32caead5` with `./gradlew clean buildPlugin` and checked the resulting `lsp4ij-0.20.2-SNAPSHOT.zip` using JetBrains Plugin Verifier 1.408 against the exact IntelliJ IDEA Ultimate 2026.2.1 build `IU-262.9437.185`.

Before:

```text
IU-262.9437.185 against com.redhat.devtools.lsp4ij:0.20.2-SNAPSHOT: 2 compatibility problems.
Package 'com.intellij.ide.structureView.impl.common' is not found along with its class com.intellij.ide.structureView.impl.common.PsiTreeElementBase.
Method StructureViewEditorFeature.lambda$collectUiRunnable$0() references an unresolved class com.intellij.ide.impl.StructureViewWrapperImpl.
```

After changing only the plugin dependency metadata and verifier range:

```text
IU-262.9437.185 against com.redhat.devtools.lsp4ij:0.20.2-SNAPSHOT: Compatible.
```

The 2026.1.5 control remains compatible:

```text
IU-261.27258.48 against com.redhat.devtools.lsp4ij:0.20.2-SNAPSHOT: Compatible.
intellij.structureView.plugin (optional): Dependency is not resolved.
```

## Validation

- `./gradlew clean buildPlugin` — passed.
- `./gradlew verifyPluginStructure verifyPluginProjectConfiguration` — passed; the project configuration task retains its existing Java 17 versus platform 2024.2 warning.
- Plugin Verifier 1.408 against `IU-262.9437.185` — compatible.
- Plugin Verifier 1.408 against `IU-261.27258.48` — compatible, with the expected missing optional dependency note.
- `./gradlew test` — 441 tests executed, with 7 unrelated failures where the IntelliJ test harness rejected macOS temporary files as outside its allowed VFS roots; 3 tests were skipped.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -182,7 +182,6 @@ intellijPlatform {
                 types = listOf(IntelliJPlatformType.IntellijIdea)
                 channels = listOf(ProductRelease.Channel.RELEASE) // Only stable releases
                 sinceBuild = "242" // From your minimum supported version
-                untilBuild = "262.*" // Up to current major version
             }
         }
     }
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -175,7 +175,7 @@ intellijPlatform {
     }
 
     pluginVerification {
-        // MISSING_DEPENDENCIES also fails for unavailable optional dependencies; see https://youtrack.jetbrains.com/issue/MP-2974.
+        // MISSING_DEPENDENCIES also fails for unavailable optional="true" dependencies; see https://youtrack.jetbrains.com/issue/MP-2974.
         failureLevel = listOf(INVALID_PLUGIN, COMPATIBILITY_PROBLEMS)
         verificationReportsFormats = listOf(MARKDOWN, PLAIN)
         ides {
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -175,7 +175,8 @@ intellijPlatform {
     }
 
     pluginVerification {
-        failureLevel = listOf(INVALID_PLUGIN, COMPATIBILITY_PROBLEMS, MISSING_DEPENDENCIES)
+        // MISSING_DEPENDENCIES also fails for unavailable optional dependencies; see https://youtrack.jetbrains.com/issue/MP-2974.
+        failureLevel = listOf(INVALID_PLUGIN, COMPATIBILITY_PROBLEMS)
         verificationReportsFormats = listOf(MARKDOWN, PLAIN)
         ides {
             select {
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -182,7 +182,7 @@ intellijPlatform {
                 types = listOf(IntelliJPlatformType.IntellijIdea)
                 channels = listOf(ProductRelease.Channel.RELEASE) // Only stable releases
                 sinceBuild = "242" // From your minimum supported version
-                untilBuild = "261.*" // Up to current major version
+                untilBuild = "262.*" // Up to current major version
             }
         }
     }
--- a/src/main/resources/META-INF/plugin-structure-view.xml
+++ b/src/main/resources/META-INF/plugin-structure-view.xml
@@ -1,0 +1,3 @@
+<?xml version="1.0" encoding="utf-8"?>
+<idea-plugin>
+</idea-plugin>
--- a/src/main/resources/META-INF/plugin.xml
+++ b/src/main/resources/META-INF/plugin.xml
@@ -347,6 +347,7 @@
     </description>
 
     <depends>com.intellij.modules.platform</depends>
+    <depends optional="true" config-file="plugin-structure-view.xml">intellij.structureView.plugin</depends>
     <depends optional="true" config-file="plugin-textmate.xml">org.jetbrains.plugins.textmate</depends>
     <depends optional="true" config-file="plugin-telemetry.xml">com.redhat.devtools.intellij.telemetry</depends>
     <depends optional="true" config-file="plugin-json.xml">com.intellij.modules.json</depends>


angelozerr:
Thanks so much @sxlijin for working on this issue!

Please tell me when you will think it will be ready.

sxlijin:
Whoops, forgot to move it out of draft.

build.gradle.kts
```diff
@@ -175,14 +175,14 @@ intellijPlatform {
     }
 
     pluginVerification {
-        failureLevel = listOf(INVALID_PLUGIN, COMPATIBILITY_PROBLEMS, MISSING_DEPENDENCIES)
+        // MISSING_DEPENDENCIES also fails for unavailable optional dependencies; see https://youtrack.jetbrains.com/issue/MP-2974.
```
angelozerr: It is a bug from JetBrains right? Is it required to do that?

angelozerr: @sxlijin please tell me if we can keep MISSING_DEPENDENCIES

sxlijin: `intellij.structureView.plugin` is introduced in 262, it looks like, so it doesn't exist in 261.* or earlier, and that raises a MISSING_DEPENDENCIES error on 261

therefore the options were either:

- retain MISSING_DEPENDENCIES and ship a different version of lsp4ij starting in 262, or
- drop MISSING_DEPENDENCIES

angelozerr: even with optional="true" ?

angelozerr: We had a similare issue with spellchecker, we did that:

```
if (ideaVersionInt >= 252) {
      // Since 2025.2, JetBrains moved spellchecker in a new "intellij.spellchecker" module
      bundledModule("intellij.spellchecker")
}
```

Could you please try to follow this idea.

sxlijin: I tested this against the exact two controls with Plugin Verifier 1.409. `bundledModule("intellij.platform.structureView")` only adds the module to the Gradle compile classpath; it does not emit a runtime dependency into `plugin.xml`. With the runtime `<depends>` removed, IU-261.27258.48 was compatible, but IU-262.9437.185 reproduced the same two unresolved Structure View classes (`PsiTreeElementBase` and `StructureViewWrapperImpl`).

Keeping `<depends optional="true">intellij.structureView.plugin</depends>` makes both IU-261.27258.48 and IU-262.9437.185 compatible. However, restoring `MISSING_DEPENDENCIES` still makes the Gradle task fail on 261 solely for `intellij.structureView.plugin (optional): Dependency is not resolved`. The exposed `intellij.platform.structureView` module does not avoid this either.

So the conditional spellchecker pattern is not sufficient for this runtime classloader move. The safest single-artifact configuration remains the optional plugin dependency plus omitting the coarse `MISSING_DEPENDENCIES` failure level until MP-2974 is fixed; `INVALID_PLUGIN` and `COMPATIBILITY_PROBLEMS` remain enforced. I clarified the build comment in 4d33a632.

angelozerr:
Thanks so much @sxlijin !

sxlijin:
FYI:

> • I implemented and tested Angelo’s exact pattern locally with MISSING_DEPENDENCIES restored:
> 
>   if (ideaVersionInt >= 262) {
>       bundledModule("intellij.platform.structureView")
>   }
> 
>   It cannot safely be pushed:
> 
>   - Without the runtime <depends>, IU-261 passes, but IU-262 fails with the original two unresolved classes: PsiTreeElementBase and StructureViewWrapperImpl.
>   - With the optional runtime <depends>, both IDEs are compatible, but IU-261 fails MISSING_DEPENDENCIES because intellij.structureView.plugin does not exist there.
>   - JetBrains requires runtime dependencies in plugin.xml; bundledModule only configures the build dependency. Plugin dependency documentation (https://plugins.jetbrains.com/docs/intellij/plugin-dependencies.html)
>   - The optional-dependency false positive remains an open verifier bug: MP-2974 (https://youtrack.jetbrains.com/projects/MP/issues/MP-2974/Plugin-Verifier-unresolved-optional-dependencies-must-not-be-highlighted-as-errors).
> 
>   I restored the known-good configuration and left the worktree clean. Nothing new was committed or pushed. Restoring MISSING_DEPENDENCIES would require dropping 261 support, publishing separate 261/262 artifacts, or refactoring away the moved implementation classes.

PR #1638, opened 2026-08-18
fix: don't let a slow write stall unrelated LSP requests

jorgsowa:
All outgoing messages (notifications and requests) were serialized through one single-threaded, unbounded messageWriter executor, so a single slow write could pile up every other feature request (hover, completion, ...) against that language server, manifesting as an IDE freeze.

Notifications (`didOpen/didChange/didClose`) still need strict ordering, so they stay on messageWriter. Requests/responses now go through a separate pooled executor (requestWriter) since they're independent of each other. Requests still wait for the most recently dispatched notification's write to finish first, so a request can't reach the server ahead of a didChange it depends on.

### DIFF
--- a/src/main/java/com/redhat/devtools/lsp4ij/LanguageServerWrapper.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/LanguageServerWrapper.java
@@ -55,6 +55,7 @@ import org.eclipse.lsp4j.jsonrpc.Launcher;
 import org.eclipse.lsp4j.jsonrpc.MessageConsumer;
 import org.eclipse.lsp4j.jsonrpc.messages.Either;
 import org.eclipse.lsp4j.jsonrpc.messages.Message;
+import org.eclipse.lsp4j.jsonrpc.messages.NotificationMessage;
 import org.eclipse.lsp4j.services.LanguageServer;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
@@ -109,6 +110,9 @@ public class LanguageServerWrapper implements Disposable {
     private final ExecutorService dispatcher;
     private final ExecutorService listener;
     private final ExecutorService messageWriter;
+    private final ExecutorService requestWriter;
+    // Tracks the most recently enqueued notification write so requests can wait for it, preserving relative order.
+    private volatile CompletableFuture<Void> lastNotificationWrite = CompletableFuture.completedFuture(null);
     private final SimpleModificationTracker modificationTracker = new SimpleModificationTracker();
     /**
      * Map containing unregistration handlers for dynamic capability registrations.
@@ -180,9 +184,21 @@ public class LanguageServerWrapper implements Disposable {
         // Executor service passed through to the LSP4j layer when we attempt to start the LS. It will be used
         // to create a listener that sits on the input stream and processes inbound messages (responses, or server-initiated
         // requests).
+        // Notifications (didOpen/didChange/didClose, ...) must be written in order, hence the single thread.
         String messageWriterThreadNameFormat = "LS-" + serverDefinition.getId() + projectName + "#messageWriter"; //$NON-NLS-1$ //$NON-NLS-2$
         this.messageWriter = Executors.newSingleThreadExecutor(new ThreadFactoryBuilder()
                 .setNameFormat(messageWriterThreadNameFormat)
+                .setThreadFactory(r -> {
+                    var t = new Thread(r);
+                    t.setContextClassLoader(workerCtxClassLoader);
+                    return t;
+                })
+                .build());
+
+        // Requests/responses have no ordering constraint, so they don't share messageWriter: a slow write no longer stalls every other feature request.
+        String requestWriterThreadNameFormat = "LS-" + serverDefinition.getId() + projectName + "#requestWriter-%d"; //$NON-NLS-1$ //$NON-NLS-2$
+        this.requestWriter = Executors.newCachedThreadPool(new ThreadFactoryBuilder()
+                .setNameFormat(requestWriterThreadNameFormat)
                 .setThreadFactory(r -> {
                     var t = new Thread(r);
                     t.setContextClassLoader(workerCtxClassLoader);
@@ -237,6 +253,7 @@ public class LanguageServerWrapper implements Disposable {
     void stopDispatcher() {
         this.dispatcher.shutdownNow();
         this.messageWriter.shutdownNow();
+        this.requestWriter.shutdownNow();
 
         // Only really needed for testing - the listener (an instance of ConcurrentMessageProcessor) should exit
         // as soon as the input stream from the LS is closed, and a cached thread pool will recycle idle
@@ -253,6 +270,10 @@ public class LanguageServerWrapper implements Disposable {
             }
             if (!this.messageWriter.awaitTermination(2, TimeUnit.SECONDS)) {
                 LOGGER.warn("Message writer executor for language server '{}' did not terminate within 2s; classloader unload may be blocked",
+                        serverDefinition.getId());
+            }
+            if (!this.requestWriter.awaitTermination(2, TimeUnit.SECONDS)) {
+                LOGGER.warn("Request writer executor for language server '{}' did not terminate within 2s; classloader unload may be blocked",
                         serverDefinition.getId());
             }
             if (!this.listener.awaitTermination(2, TimeUnit.SECONDS)) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/redhat/devtools/lsp4ij/LanguageServerWrapper.java
```diff
@@ -109,6 +110,9 @@ public class LanguageServerWrapper implements Disposable {
     private final ExecutorService dispatcher;
     private final ExecutorService listener;
     private final ExecutorService messageWriter;
+    private final ExecutorService requestWriter;
+    // Tracks the most recently enqueued notification write so requests can wait for it, preserving relative order.
+    private volatile CompletableFuture<Void> lastNotificationWrite = CompletableFuture.completedFuture(null);
```
angelozerr, line 115: The volatile keyword guarantees visibility but not atomicity of the read-then-chain sequence. Scenario:
  
   1. Thread A (request) reads lastNotificationWrite → gets future F1
   2. Thread B (notification) updates lastNotificationWrite → F2
   3. Thread A chains .thenRunAsync(...) on F1 (the old one), not F2
  
   In practice this is acceptable (the request was submitted before notification F2), but an
   AtomicReference<CompletableFuture<Void>> would make the intent more explicit and remove any ambiguity.

src/main/java/com/redhat/devtools/lsp4ij/LanguageServerWrapper.java
```diff
@@ -475,8 +496,17 @@ public synchronized void start() throws LanguageServerException {
...
-                                        .exceptionally(e -> {
+                                // Requests use requestWriter (so a slow one can't stall others) but still wait for prior notifications to be written, to not overtake a didChange they depend on.
+                                CompletableFuture<Void> write;
+                                if (message instanceof NotificationMessage) {
+                                    write = CompletableFuture.runAsync(() -> consumer.consume(message), messageWriter);
+                                    lastNotificationWrite = write;
+                                } else {
+                                    write = lastNotificationWrite
+                                            .exceptionally(e -> null)
+                                            .thenRunAsync(() -> consumer.consume(message), requestWriter);
```
angelozerr, line 507: requestWriter is a cached thread pool, so multiple requests may call consumer.consume(message) concurrently. This only works because LSP4J's StreamMessageConsumer.consume() uses a synchronized (outputLock) internally.
 
   It would be worth adding a brief comment explaining this dependency, e.g.:
   // Safe to call consume() from multiple threads concurrently:
   // StreamMessageConsumer.consume() synchronizes on an internal outputLock.
  
 Without this comment, a future contributor might not understand why a multi-threaded pool is safe here.

src/main/java/com/redhat/devtools/lsp4ij/LanguageServerWrapper.java
```diff
@@ -180,6 +184,7 @@ public LanguageServerWrapper(@NotNull Project project,
         // Executor service passed through to the LSP4j layer when we attempt to start the LS. It will be used
         // to create a listener that sits on the input stream and processes inbound messages (responses, or server-initiated
         // requests).
+        // Notifications (didOpen/didChange/didClose, ...) must be written in order, hence the single thread.
```
angelozerr: // Executor service passed through to the LSP4j layer when we attempt to start the LS. It will be used
  // to create a listener that sits on the input stream and processes inbound messages (responses, or server-initiated
  // requests).
  // Notifications (didOpen/didChange/didClose, ...) must be written in order, hence the single thread.

  The original comment (lines 184-186) doesn't accurately describe messageWriter's role — it actually describes
   listener. Now that a new comment was added at line 187, the old one is even more confusing. Consider replacing lines 184-186 entirely:
 
 // Single-threaded executor for writing outgoing notifications (didOpen, didChange, didClose)
 // to the language server. Notifications must be written in order, hence the single thread.
 // Requests and responses use the separate requestWriter pool.

src/main/java/com/redhat/devtools/lsp4ij/LanguageServerWrapper.java
```diff
@@ -475,8 +496,17 @@ public synchronized void start() throws LanguageServerException {
                                 // To avoid having some lock problem when message is written in the stream output
                                 // (when there are a lot of messages to write it)
                                 // we consume the message in async mode
-                                CompletableFuture.runAsync(() -> consumer.consume(message), messageWriter)
-                                        .exceptionally(e -> {
+                                // Requests use requestWriter (so a slow one can't stall others) but still wait for prior notifications to be written, to not overtake a didChange they depend on.
+                                CompletableFuture<Void> write;
+                                if (message instanceof NotificationMessage) {
+                                    write = CompletableFuture.runAsync(() -> consumer.consume(message), messageWriter);
```
angelozerr, line 502: If consumer.consume(message) throws inside the runAsync, the write future completes exceptionally. The next request at line 505 handles this via .exceptionally(e -> null) — good.
 
  However, lastNotificationWrite is updated unconditionally at line 503, even before the write completes. This is
  correct behavior (it's a future, not a result), but worth noting: if the messageWriter executor is shut down,
  runAsync throws RejectedExecutionException synchronously, and lastNotificationWrite would point to a failed future.
  The .exceptionally(e -> null) on line 505 handles this gracefully, so no bug — just worth being aware of.

src/main/java/com/redhat/devtools/lsp4ij/LanguageServerWrapper.java
```diff
@@ -475,8 +496,17 @@ public synchronized void start() throws LanguageServerException {
                                 // To avoid having some lock problem when message is written in the stream output
                                 // (when there are a lot of messages to write it)
                                 // we consume the message in async mode
-                                CompletableFuture.runAsync(() -> consumer.consume(message), messageWriter)
-                                        .exceptionally(e -> {
+                                // Requests use requestWriter (so a slow one can't stall others) but still wait for prior notifications to be written, to not overtake a didChange they depend on.
+                                CompletableFuture<Void> write;
+                                if (message instanceof NotificationMessage) {
```
angelozerr, line 501: There is a subtle gap between line 501 (the runAsync is submitted to messageWriter) and line 503
  (lastNotificationWrite = write). During this gap, a concurrent request thread could read the old
  lastNotificationWrite and miss this notification entirely.
  
  Consider reversing the order — assign the future first, then submit. One way:
  if (message instanceof NotificationMessage) {
       CompletableFuture<Void> notifWrite = CompletableFuture.runAsync(() -> consumer.consume(message), messageWriter);
       lastNotificationWrite = notifWrite;
       write = notifWrite;
   } else {
   Though in practice the gap is negligible since this whole block runs on a single thread (the LSP4J dispatch thread),
   it's worth confirming that assumption holds.

src/main/java/com/redhat/devtools/lsp4ij/LanguageServerWrapper.java
```diff
@@ -475,8 +496,17 @@ public synchronized void start() throws LanguageServerException {
                                 // To avoid having some lock problem when message is written in the stream output
                                 // (when there are a lot of messages to write it)
                                 // we consume the message in async mode
-                                CompletableFuture.runAsync(() -> consumer.consume(message), messageWriter)
-                                        .exceptionally(e -> {
+                                // Requests use requestWriter (so a slow one can't stall others) but still wait for prior notifications to be written, to not overtake a didChange they depend on.
+                                CompletableFuture<Void> write;
```
angelozerr: For the request branch (lines 504-507), the .exceptionally(e -> null) at line 505 swallows the notification error
 before chaining the request write — that's correct.
 
  But if the thenRunAsync at line 507 itself fails (e.g., requestWriter is shut down, or consumer.consume() throws),
  the error will propagate to write.exceptionally(...) at line 508 — which is good, it's handled.
 
  However, note that write for the request branch is a 2-stage chain. If the error happens at the .exceptionally(e ->
  null) stage (line 505), it's swallowed silently without logging. This is fine since that error belongs to the
  notification, which already has its own error handler from a prior iteration. Just confirming this is intentional.

src/main/java/com/redhat/devtools/lsp4ij/LanguageServerWrapper.java
```diff
@@ -475,8 +496,17 @@ public synchronized void start() throws LanguageServerException {
                                 // To avoid having some lock problem when message is written in the stream output
                                 // (when there are a lot of messages to write it)
                                 // we consume the message in async mode
-                                CompletableFuture.runAsync(() -> consumer.consume(message), messageWriter)
-                                        .exceptionally(e -> {
+                                // Requests use requestWriter (so a slow one can't stall others) but still wait for prior notifications to be written, to not overtake a didChange they depend on.
```
angelozerr: The whole if/else block (lines 500-507) that reads and writes lastNotificationWrite assumes it runs on a single
  thread (the LSP4J message dispatch thread). If this wrapper could be called concurrently for different messages, the volatile field alone is insufficient — you'd need synchronization.
  
  Could you confirm that LSP4J guarantees this wrapper is invoked sequentially? If so, a comment like // Called
  sequentially by LSP4J's message dispatch thread would help future readers.

src/main/java/com/redhat/devtools/lsp4ij/LanguageServerWrapper.java
angelozerr, line 115: Add  @NotNull

src/main/java/com/redhat/devtools/lsp4ij/LanguageServerWrapper.java
angelozerr, line 187: // Executor service passed through to the LSP4j layer when we attempt to start the LS. It will be used
  // to create a listener that sits on the input stream and processes inbound messages (responses, or server-initiated
  // requests).
  // Notifications (didOpen/didChange/didClose, ...) must be written in order, hence the single thread.

  ▎ The original comment (lines 184-186) doesn't accurately describe messageWriter's role — it describes listener. The
  ▎ new comment at line 187 is correct but contradicts the old one. Consider replacing lines 184-186 entirely:
  ▎ // Single-threaded executor for writing outgoing notifications (didOpen, didChange, didClose)
  ▎ // to the language server. Notifications must be written in order, hence the single thread.
  ▎ // Requests and responses use the separate requestWriter pool.

angelozerr:
@jorgsowa please rebase your PR from main branch, since I did some changes for didChange just to check that your PR works correctly with my changes.

Issue #1300, opened 2025-09-26
Inconsistent identifier highlighting with 2025.2


SCWells72:
This seems to be a recent regression, but I noticed that when the caret is placed on an identifier in an LSP4IJ-managed document, identifier highlighting is applied in an inconsistent/unreliable manner. Here's an example:

![Image](https://github.com/user-attachments/assets/ef0595cf-6c43-4058-a7ac-014d7fb11ae2)

Hopefully that conveys the issue clearly where at first the identifier under the caret is highlighted properly, but as I click around, it stops working, though not completely. However, I no longer see `textDocument/documentHighlight` entries in the trace log as I click around, so that's probably a good lead on whatever is happening.

The example above is in TypeScript, but I see the same thing for other language servers as well, e.g., CSS.

Oh, and this is all with the latest published version, 0.16.1.

Let me know if I can provide any other details.

angelozerr:
Have you the problem with IJ community?

Have you the same problem with old IJ?

Thats strange because I dont remember having doing some changes about highlight (but perhaps I am wrong:))

SCWells72:
> Have you the problem with IJ community?

Yes, this is all in IntelliJ IDEA Community Edition since that doesn't have native JavaScript/TypeScript/CSS support and LSP4IJ is being used to provide support for those languages there.

> Have you the same problem with old IJ?

I'm not sure. I don't have older versions of Community Edition installed, only Ultimate Edition for verification and testing of my own plugin against older versions. I noticed this a little back so my guess is that it's been happening for at least a few versions.

Hopefully it should be pretty simple to reproduce based on the description and video I provided above, but let me know if not.

SCWells72:
The one in the demo has some additional files behind it, but I reproduced it in a standalone project with no other dependencies. That project contains the following two files:

**`Account.ts`**
```typescript
export interface Account {
    AccountNumber: string;
}
```

**`MyComponent.ts`**
```typescript
import {Account} from "./Account";

export default class MyComponent {
    testAccountAccess() {
        let account: Account = { AccountNumber: "12345" };
        let accountNumber = account.AccountNumber;
    }
}
```

LSP4IJ 0.16.1 is the only installed third-party plugin:

```
IntelliJ IDEA 2025.2.2 (Community Edition)
Build #IC-252.26199.169, built on September 18, 2025
Source revision: 7fd6c69de64ef
[...]

PR #1659, opened 2026-08-31
fix: Prevent whole-file PSI caching in highlight handler

ethanhann:
Relates to #1300
Plugin version: 0.21.0
I verified this fix on RustRover 2026.1.4.

When a language server returns an empty highlight list for a file whose only PSI element spans the whole file, such as plain text, the empty result is cached for the whole file. It stops sending `documentHighlight` requests until the file is reopened which breaks highlighting. This PR returns an empty handler instead of null in that case. The empty result is cached for the caret offset only which allows highlighting to keep working.

### DIFF
--- a/src/main/java/com/redhat/devtools/lsp4ij/features/highlight/LSPHighlightUsagesHandlerFactory.java
+++ b/src/main/java/com/redhat/devtools/lsp4ij/features/highlight/LSPHighlightUsagesHandlerFactory.java
@@ -20,6 +20,7 @@ import com.intellij.openapi.progress.util.ProgressIndicatorUtils;
 import com.intellij.openapi.progress.ProgressManager;
 import com.intellij.openapi.util.TextRange;
 import com.intellij.openapi.vfs.VirtualFile;
+import com.intellij.psi.PsiElement;
 import com.intellij.psi.PsiFile;
 import com.redhat.devtools.lsp4ij.LSPFileSupport;
 import com.redhat.devtools.lsp4ij.LSPIJUtils;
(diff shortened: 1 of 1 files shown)


src/main/java/com/redhat/devtools/lsp4ij/features/highlight/LSPHighlightUsagesHandlerFactory.java
```diff
@@ -58,7 +59,19 @@ public class LSPHighlightUsagesHandlerFactory implements HighlightUsagesHandlerF
             return null;
         }
         List<LSPHighlightPsiElement> targets = getTargets(editor, file);
-        return targets.isEmpty() ? null : new LSPHighlightUsagesHandler(editor, file, targets);
+        if (targets.isEmpty()) {
+            // Returning null here would make the IDE cache the empty result for the whole plain-text file,
+            // and it would stop sending 'textDocument/documentHighlight' until the document changes.
+            // See https://github.com/redhat-developer/lsp4ij/issues/1300
+            return isWholeFileElementAtCaret(editor, file) ? new LSPHighlightUsagesHandler(editor, file, targets) : null;
```
angelozerr: It means that you will call findElementAt every time when targets are empty.

I think this issue is only for `TEXT` and `textmate` language. Generally PsiFile findElementAt doesnt return the full text range file.

So my suggestion is to replace isWholeFileElementAtCaret call with SimpleLanguageUtils.isSUpported(file.getLanguage())

ethanhann: Thanks for the feedback. I wasn't aware of the `SimpleLanguageUtils.isSupported()` method. I retested with this change and it works perfectly.

angelozerr:
Thanks @ethanhann !

PR #1662, opened 2026-09-04
fix: Type/Call Hierarchy actions disabled when the IDE bundles JetBrains' LSP client

### LINKED ISSUE #1661: Type Hierarchy / Call Hierarchy actions are disabled for user-defined servers when the IDE bundles JetBrains' LSP client (its `language=""` provider shadows LSP4IJ's `TEXT`/`textmate` one)
## Environment

- LSP4IJ 0.21.0
- IDE: `IntelliJ IDEA 2026.2.1 Build #IU-262.9437.185, built on August 9, 2026`
  (any build that ships the `intellij.platform.lsp.impl` module reproduces it: Ultimate, and every
  edition from 2026.1.4 on, where the LSP client API was opened up)
- A user-defined language server mapped by **file name patterns** (`*.xdsl`, plain-text file type),
  whose `initialize` response advertises `typeHierarchyProvider: true` and answers
  `textDocument/prepareTypeHierarchy` / `typeHierarchy/supertypes` / `typeHierarchy/subtypes`
  (verified over stdio outside the IDE)

## Steps to reproduce

1. Register a user-defined language server that supports type hierarchy and map it to a file name
   pattern for an extension IntelliJ has no file type for (the file opens as plain text).
2. Open a mapped file, put the caret **on a word** (any identifier), press `Ctrl+H` or open
   **Navigate → Type Hierarchy**.

## Expected

The Hierarchy tool window opens and `textDocument/prepareTypeHierarchy` is sent (the documented
behaviour: `docs/LSPSupport.md`, *Type Hierarchy* — "After setting the cursor position in a file, you
can view the Type Hierarchy using the `Navigate / Type Hierarchy` menu (or `Ctrl+H`)").

## Actual

Nothing happens. **Navigate → Type Hierarchy** is greyed out. No request reaches the server
(nothing in the LSP console trace). With debug logging enabled for
`#com.intellij.ide.hierarchy.actions.BrowseHierarchyActionBase`, `idea.log` shows:

```
FINE - #c.i.i.h.a.BrowseHierarchyActionBase - Using provider com.intellij.platform.lsp.impl.features.hierarchy.type.LspTypeHierarchyProvider@5b2c3bd1
FINE - #c.i.i.h.a.BrowseHierarchyActionBase - Target: null
```

The provider IntelliJ picked is the platform's built-in LSP client's, not LSP4IJ's
`LSPTypeHierarchyProvider`, and it has no server for the file, so its target is null and the action
is disabled.

## Analysis

Four pieces, each straight from source:

1. **LSP4IJ reports the word under the caret as a `Language.ANY` element.** `LSPTargetElementEvaluator`
   (registered `<targetElementEvaluator language="">`) returns `new LSPPsiElement(file, wordRange)` for a
   caret on a word. `LSPPsiElement extends FakePsiElement`, and `FakePsiElement.getLanguage()` returns
   `Language.ANY` (id `""`). That element is what the editor exposes as `CommonDataKeys.PSI_ELEMENT`.

2. **The platform LSP client registers its hierarchy providers for exactly that key.**
   `platform/lsp-impl/resources/intellij.platform.lsp.impl.xml`:

   ```xml
   <typeHierarchyProvider language="" implementationClass="com.intellij.platform.lsp.impl.features.hierarchy.type.LspTypeHierarchyProvider" order="last"/>
   <callHierarchyProvider language="" implementationClass="com.intellij.platform.lsp.impl.features.hierarchy.call.LspCallHierarchyProvider" order="last"/>
   ```

   LSP4IJ registers `LSPTypeHierarchyProvider` for `TEXT` and `textmate` only, so for a
   `Language.ANY` element the provider list contains only the platform's.

3. **`BrowseHierarchyActionBase` never falls back to the file's language once a provider exists for
   the element's.** `findProvider` calls `findBestHierarchyProvider(extension, psiElement, …)` first;
   that method returns the first provider whose `getTarget` is non-null, **else the first provider
   in the list**. The `psiFile` lookup — where LSP4IJ's `TEXT` registration would be found — only
   runs when the element lookup returns `null`, i.e. when the list is empty. `isEnabled` then requires
   `provider.getTarget(dataContext) != null`, which the platform provider fails for a file no
   JetBrains-LSP server handles.

4. Consequently the action is disabled for every LSP4IJ user-defined server on these IDE builds
   whenever the caret is on a word — which is always, for a hierarchy request. Before the platform
   provider existed, the `ANY` list was empty and the `psiFile` fallback reached LSP4IJ's provider.

Call Hierarchy has the identical shape (`callHierarchyProvider language=""` on the platform side,
`TEXT`/`textmate` on LSP4IJ's), so it is expected to be broken the same way.

## Proposed fix

Also register LSP4IJ's providers for the `Language.ANY` key, e.g. in `plugin.xml`:

```xml
<typeHierarchyProvider language="" implementationClass="com.redhat.devtools.lsp4ij.features.typeHierarchy.LSPTypeHierarchyProvider"/>
<callHierarchyProvider language="" implementationClass="com.redhat.devtools.lsp4ij.features.callHierarchy.LSPCallHierarchyProvider"/>
```

The platform's own entries are `order="last"`, so LSP4IJ's would be consulted first for an
`LSPPsiElement`; `LSPHierarchyProviderBase.getTarget` already returns `null` for files LSP4IJ does not
support, so the two clients keep yielding to each other correctly on files the other one owns.

(Arguably `BrowseHierarchyActionBase` should also fall back to the file's language when no provider
yields a target — that is a platform issue and does not preclude the one-line fix here.)

## Workaround

Put the caret on **whitespace** inside the declaration (no word under it): `LSPTargetElementEvaluator`
returns no element, `PSI_ELEMENT` is null, the `psiFile` fallback finds LSP4IJ's `TEXT` provider, and
`textDocument/prepareTypeHierarchy` is sent for that position (our server resolves a whitespace
cursor to the enclosing declaration, so the tree opens). Confirmed working.

## Related: the tree's root node is the caret text, not the prepared item

`LSPHierarchyTreeStructureBase` makes the caret's `LSPPsiElement` the ROOT descriptor
(`setBaseElement(createHierarchyNodeDescriptor(project, null, psiElement, null))`) and attaches the
`prepareTypeHierarchy` result items as its CHILDREN (`buildRoot` when `isBase()`). `LSPPsiElement.getName()`
is the document text under the element's range, so the root reads as the word under the caret — a
duplicate of the prepared item's name when the caret is on the name (`Server › Server › Named`), and an
EMPTY label when the caret is on whitespace (the `getTarget` fallback range is `TextRange(offset,
offset + 1)`, one blank character), which is exactly the position the workaround above needs.

The LSP contract already answers "what is the root": `prepareTypeHierarchy` returns the item(s) the
hierarchy is about (VS Code roots the view on them). Suggest rooting the tree on the prepared
`TypeHierarchyItem`s (one root per item; the caret element never shown), which removes both the
duplicate and the blank node. Same for `prepareCallHierarchy`.

bamboo:
The JetBrains LSP client registers its type/call hierarchy providers with language="", which is the language of the LSPPsiElement returned by LSPTargetElementEvaluator for a caret on a word (FakePsiElement.getLanguage() is Language.ANY). BrowseHierarchyActionBase only falls back to the file language when no provider matches the element language, so LSP4IJ's TEXT / textmate providers were never consulted and the actions stayed disabled.

Register the LSP4IJ providers for language="" as well. The platform ones are order="last", so LSP4IJ's are consulted first, and LSPHierarchyProviderBase.getTarget already returns null for files LSP4IJ does not support.

Fixes #1661

Verified with:
- IntelliJ IDEA 2024.2 (Community Edition) Build #IC-242.20224.300, built on August 6, 2024
- IntelliJ IDEA 2026.2.1 Build #IU-262.9437.185, built on August 9, 2026

### DIFF
--- a/src/main/resources/META-INF/plugin.xml
+++ b/src/main/resources/META-INF/plugin.xml
@@ -743,6 +743,11 @@
         <typeHierarchyProvider
                 language="textmate"
                 implementationClass="com.redhat.devtools.lsp4ij.features.typeHierarchy.LSPTypeHierarchyProvider"/>
+        <!-- Required for LSPPsiElement (Language.ANY), otherwise the JetBrains LSP client provider
+             (registered with language="") shadows the TEXT / textmate ones. -->
+        <typeHierarchyProvider
+                language=""
+                implementationClass="com.redhat.devtools.lsp4ij.features.typeHierarchy.LSPTypeHierarchyProvider"/>
 
         <!-- LSP textDocument/callHierarchy request support -->
         <callHierarchyProvider
@@ -750,6 +755,11 @@
                 implementationClass="com.redhat.devtools.lsp4ij.features.callHierarchy.LSPCallHierarchyProvider"/>
         <callHierarchyProvider
                 language="textmate"
+                implementationClass="com.redhat.devtools.lsp4ij.features.callHierarchy.LSPCallHierarchyProvider"/>
+        <!-- Required for LSPPsiElement (Language.ANY), otherwise the JetBrains LSP client provider
+             (registered with language="") shadows the TEXT / textmate ones. -->
+        <callHierarchyProvider
+                language=""
                 implementationClass="com.redhat.devtools.lsp4ij.features.callHierarchy.LSPCallHierarchyProvider"/>
 
         <!-- LSP Find Usages (LSP) support shows:


angelozerr:
Thanks @bamboo !

PR #1667, opened 2026-09-14
fix: completion without textEdit removes the whitespace before the caret

### LINKED ISSUE #993: Simplest CompletionItem removes all leading whitespace
Reproduced in 0.12.0 and 0.13.0-20250424-115448.

1. Use a language server that returns the simplest `CompletionItem` with just `label` defined.
2. Put some whitespace in front before invoking code completion.
3. Accept the item.
4. All leading whitespace is deleted.
```JSON
{
  "isIncomplete": false,
  "items": [
    {
      "label": "driver"
    }
  ]
}
```

https://github.com/user-attachments/assets/1dc1f95b-923e-4ba6-885d-499d198ce653

sergeysavchuk:
A completion item with only `insertText` (no `textEdit`) removes the whitespace
before the caret when it is applied on a blank line.

The prefix start comes from `getWordRangeAt`, which returns the PSI element at
`offset - 1` — on a blank line, the whitespace leaf. The existing guard only
rejects a *one character* element, so `"\n\n"` or `"\n   "` gets through and the
completion replaces it.

The fix returns `null` for whitespace, which also restores the documented
`foo  |  bar -> null` contract.

Fixes #993


angelozerr:
LGTM, thanks @sergeysavchuk !

PR #1671, opened 2026-09-18
fix: off-by-one when navigating to diagnostic related information links

### LINKED ISSUE #1670: Off-by-one error when navigating to diagnostic related information links
Hovering an error with `relatedInformation` and clicking a related-info file link puts the caret one line above the reported location.

Repro: server reports (0-based per spec):

```json
{"location": {"uri": "...", "range": {"start": {"line": 3, "character": 9}, "end": {"line": 3, "character": 10}}}, "message": "..."}
```

Clicking the link navigates to displayed line 3 instead of 4 (column is correct, only the line is off).

Cause: `LSPNavigationLinkHandler.toNavigationUrl` writes the raw 0-based line into the `#L<line>:<character>`: https://github.com/redhat-developer/lsp4ij/blob/3bd9a7f13e4f3ae5f105b4f52ccf0561f22d8773/src/main/java/com/redhat/devtools/lsp4ij/hint/LSPNavigationLinkHandler.java#L74-L84

but `LSPIJUtils.toPosition` reads it back with `line - 1 // Line numbers should be 1-based`: https://github.com/redhat-developer/lsp4ij/blob/3bd9a7f13e4f3ae5f105b4f52ccf0561f22d8773/src/main/java/com/redhat/devtools/lsp4ij/LSPIJUtils.java#L373

Writer and reader disagree: either side could be adjusted to match the other.

Environment: LSP4IJ 0.21.0, CLion 2026.2.2, macOS 14.7 aarch64.

emillaine:
Clicking a related-information link in a diagnostic tooltip landed the caret one line above the reported location.

`LSPNavigationLinkHandler.toNavigationUrl` wrote the raw 0-based LSP line into the `#L<line>:<character>` fragment, but `LSPIJUtils.openInEditor` parses it back as 1-based, landing the caret one line above the reported location. Emit 1-based to match the reader.

Fixes #1670


angelozerr:
LGTM.

Thanks @emillaine !

PR #1674, opened 2026-09-20
docs: repoint dead links in the developer docs and plugin description

NotAFlightRisk:
The two extension point links in the plugin description - the ones you click from the Marketplace listing and the IDE's Plugins dialog - point at `DeveloperGuide.md` anchors that aren't headings in it, so both just drop you at the top of the page instead of the section. `LSPSupport.md` has one of those too: the "implementation details" link next to `textDocument/codeAction` goes to `#codeAction`, where the heading renders as `#code-action`. Every other in-page anchor in that file is fine.

The rest are 404s. `LSPApi.md` links `LSPAbstractFeature.java` twice while the link text beside it already says `AbstractLSPDocumentFeature`, which is the class thats actually there. `LSPSupport.md` has `LSPCodeBlockProvider` under `features/foldingRange/` and it moved to `features/codeBlockProvider/`. And the Rust Analyzer setup page links `pkg.Rust.dev/Rustlang.org/x/tools/Rustpls`, a host with no DNS record at all - its the gopls page with go swapped for Rust a bit too enthusiastically. I've sent that one to rust-analyzer.github.io, which is a guess at what you'd want there.


angelozerr:
Thanks @NotAFlightRisk !

